PR #8, opened 2017-05-26
Fix warnings and update to new version of compiler

Pratyush:
This fixes a couple of warnings related to unused macros, and makes the library compile with newer compilers.

### DIFF
--- a/libenclave/src/lib.rs
+++ b/libenclave/src/lib.rs
@@ -8,20 +8,20 @@
  * Free Software Foundation, either version 3 of the License, or (at your
  * option) any later version.
  */
+#![deny(unused_extern_crates)]
+#![deny(unused_imports)]
+#![deny(unused_import_braces, unused_qualifications)]
 
-#![feature(linkage,lang_items,unwind_attributes,asm,const_fn,collections,unicode,alloc,oom,heap_api)]
+#![feature(linkage,lang_items,unwind_attributes,asm,const_fn,collections,alloc,oom,heap_api)]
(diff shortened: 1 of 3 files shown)


libenclave/src/lib.rs
```diff
@@ -12,16 +12,16 @@
...
+extern crate std_unicode;
 extern crate alloc as rustc_alloc;
 extern crate sgx_isa;
 #[cfg(not(test))] pub extern crate core_io as io;
 
 extern crate spin;
 extern crate rlibc;
 extern crate alloc_buddy_simple;
-#[macro_use] extern crate bitflags;
+extern crate bitflags;
```
jethrogb: This import can probably just go then.

Pratyush: Done, I also removed `rlibc` and `std_unicode`. They didn't seem to be used anywhere else.

jethrogb: I don't think those can be removed as they will result in linker errors when using the crate. Definitely not rlibc and possibly std_unicode although I don't remember its function.

Pratyush:
OK, re-added `rlibc`. It seems leaving out `std_unicode` does not cause an issue (my enclave compiled). Do you want me to re-add it anyway?

Pratyush:
Bump, anything blocking this?

jethrogb:
I need to check where the std_unicode requirement is/was coming from before I'm ok with removing it.

jethrogb:
Closing, outdated.

PR #33, opened 2018-11-22
propagate error to calling code

akash-fortanix:


jethrogb:
I think this unwrap can never fail. Then it's ok to keep.

parthsane:
PR unnecessary as unwrap can never fail

PR #38, opened 2018-12-11
Add experimental support for libraries to enclave-runner

jethrogb:


enclave-runner/src/usercalls/mod.rs
```diff
@@ -229,59 +229,99 @@ struct RunningTcs {
     event_queue: Receiver<u8>,
 }
 
-pub(crate) struct EnclaveState {
+trait EnclaveKind: Send + Sync {
```
jseyfried: How about `enum EnclaveKind { Command(Command), Library(Library) }` for clarity and to avoid the `Box::new` allocation?

enclave-runner/src/usercalls/mod.rs
```diff
@@ -229,59 +229,99 @@ struct RunningTcs {
...
+
+impl EnclaveKind for Command {
+    fn as_command(&self) -> Option<&Command> {
+        Some(self)
+    }
+}
+
+struct Library {
+    threads: Mutex<Receiver<StoppedTcs>>,
+    thread_sender: Mutex<Sender<StoppedTcs>>,
```
jseyfried: Since this is `mpsc`, can't we have multiple `Sender`s, so no need for this `Mutex`?

It looks senders and receivers are the same (since a thread `recv`s a `Tcs` and then `send`s it back), so I think we'd want `mpmc` with no `Mutex` or just `Mutex<VecDeque<StoppedTcs>>` (which I believe would be no more contention than now).

jethrogb: Yes mpmc would be great but this is not very contended, so I didn't want to add a dependency on an mpmc implementation. I'm using spsc instead of mpsc here because again it's not very contended and I don't want to clone the sender to each thread. We can't do s(p+c) as you suggest because that would lead to deadlocks.

enclave-runner/src/usercalls/mod.rs
jethrogb: Sure I can do that

enclave-runner/src/usercalls/mod.rs
jseyfried: > this is not very contended

Makes sense.

> We can't do s(p+c) as you suggest because that would lead to deadlocks.

Why? I was envisioning `recv` is lock+pop_front+unlock and `send` is lock+push_back+unlock, which can always make progress.

jethrogb: Receiver needs to block if there is nothing to pop.

jseyfried: I see.

PR #39, opened 2018-12-11
Add libsgx_enclave_common enclave loader

jethrogb:
Depends on #37  (now merged)

### DIFF
--- a/sgxs-loaders/src/lib.rs
+++ b/sgxs-loaders/src/lib.rs
@@ -14,9 +14,12 @@ extern crate sgx_isa as abi;
 extern crate sgxs as sgxs_crate;
 #[macro_use]
 extern crate failure_derive;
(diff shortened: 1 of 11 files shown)


sgxs-loaders/src/lib.rs
```diff
@@ -14,5 +14,44 @@ extern crate sgx_isa as abi;
...
+
+use std::fmt::Debug;
+use std::os::raw::c_void;
+use std::sync::Arc;
+
+use sgxs_crate::loader;
+
+#[derive(Debug)]
+pub struct Tcs {
+    _mapping: Arc<Debug + Sync + Send>,
```
jseyfried: Nice. I suppose this would just be `Arc<Drop + Sync + Send>` if we didn't want `Debug`.

PR #57, opened 2019-01-08
Add .travis.yml to enable travis CI

parthsane:
Added Travis CI support. 

Currently this just checks if build succeeds. No tests are run.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,8 @@
+language: rust
+before_install:
+  - sudo apt-get update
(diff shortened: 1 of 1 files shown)


jethrogb:
You should be able to run `cargo test` here I think?

parthsane:
[travislog.txt](https://github.com/fortanix/rust-sgx/files/2740049/travislog.txt)

`cargo test --all --exclude aesm-client` seems to do the trick

PR #69, opened 2019-01-22
fix manifest parsing & ftxsgx-runner-cargo should never use coresident signature

akash-fortanix:


fortanix-sgx-tools/src/bin/ftxsgx-runner-cargo.rs
```diff
@@ -29,12 +29,12 @@ struct Target {
...
 #[serde(rename_all = "kebab-case")]
 struct Metadata {
-    fortanix_sgx: Target
+    fortanix_sgx: Option<Target>
 }
 
 #[derive(Deserialize, Debug)]
 struct Package {
-    metadata: Metadata
+    metadata: Option<Metadata>
```
jethrogb: Instead of using `Option`, add `#[derive(Default)]` on `Metadata` and `Target` and `#[serde(default)]` on the fields in `Package` and `Metadata`. This simplifies the code below.

fortanix-sgx-tools/src/bin/ftxsgx-runner-cargo.rs
```diff
@@ -97,7 +99,12 @@ fn run() -> Result<(), Error> {
     run_command(sgxs_append_command)?;
 
     let mut ftxsgx_runner_command = Command::new("ftxsgx-runner");
-    ftxsgx_runner_command.arg(&bin_with_ext);
+    ftxsgx_runner_command.arg(&bin_with_ext)
+        .arg("--signature")
+        .arg("dummy");
+
+    println!("{:?}", ftxsgx_runner_command);
```
jethrogb: Remove

fortanix-sgx-tools/src/bin/ftxsgx-runner.rs
```diff
@@ -7,38 +7,61 @@
...
-            ); // extra newline
-            bail!("Missing <ENCLAVE> parameter on command line");
-        }
-    };
+    let default_signature_value = format!("{:?}", &Signature::coresident);
+
+    let args = App::new("ftxsgx-runner")
+        .arg(
+            Arg::with_name("file")
+                .index(1)
```
jethrogb: I think it might not be necessary to call `index`

fortanix-sgx-tools/src/bin/ftxsgx-runner.rs
```diff
@@ -7,38 +7,61 @@
...
-                .and_then(|p| Path::new(p).file_name())
-                .map(|p| Path::new(p).display().to_string());
-            eprintln!(
-                "Usage: {} <ENCLAVE>\n",
-                cmdname.unwrap_or("ftxsgx-runner".into())
-            ); // extra newline
-            bail!("Missing <ENCLAVE> parameter on command line");
-        }
-    };
+    let default_signature_value = format!("{:?}", &Signature::coresident);
```
jethrogb: According to the docs, `arg_enum!` will implement `Display`, so you should just be able to do `Signature::coresident.to_string()`

fortanix-sgx-tools/src/bin/ftxsgx-runner.rs
```diff
@@ -7,38 +7,61 @@
...
+
+    let args = App::new("ftxsgx-runner")
```
jethrogb: I prefer the macro syntax but it's up to you

fortanix-sgx-tools/src/bin/ftxsgx-runner.rs
```diff
@@ -7,38 +7,61 @@
...
         .context("While opening SGX device")?
         .einittoken_provider(AesmClient::new())
         .build();
-    let enclave = Command::new(&file, &mut device).context("While loading SGX enclave")?;
+
+    let mut enclave_builder = EnclaveBuilder::new(file.as_ref());
+
+    match signature {
+        "dummy"=> { enclave_builder.dummy_signature(); }
+        _ => ()
```
jethrogb: `coresident` is not the same as not calling any XXX_signature function on the builder

fortanix-sgx-tools/src/bin/ftxsgx-runner-cargo.rs
akash-fortanix: Done

fortanix-sgx-tools/src/bin/ftxsgx-runner-cargo.rs
akash-fortanix: Done

fortanix-sgx-tools/src/bin/ftxsgx-runner.rs
akash-fortanix: True, removed.

fortanix-sgx-tools/src/bin/ftxsgx-runner.rs
akash-fortanix: Done

fortanix-sgx-tools/src/bin/ftxsgx-runner.rs
akash-fortanix: Unchanged

fortanix-sgx-tools/src/bin/ftxsgx-runner.rs
```diff
@@ -7,38 +7,61 @@
...
         .context("While opening SGX device")?
         .einittoken_provider(AesmClient::new())
         .build();
-    let enclave = Command::new(&file, &mut device).context("While loading SGX enclave")?;
+
+    let mut enclave_builder = EnclaveBuilder::new(file.as_ref());
+
+    match signature {
+        "dummy"=> { enclave_builder.dummy_signature(); },
+        "coresident" => { enclave_builder.coresident_signature().context("While loading coresident signature")?; },
```
jethrogb: Ok but now I can't get the default behavior. Maybe make the argument optional (instead of having a default)? You can so something like this:
```rust
match args.value_of("signature").map(|v| v.parse().unwrap()) {
    Some(Signature::coresident) => ...,
    Some(Signature::dummy) => ...,
    None => ...,
}
```

akash-fortanix: Done

PR #78, opened 2019-01-28
Enhancing gdb.py script to show backtrace for an abort.

VardhanThigle:
## Overview:
With this chage, To get a back trace in case of an abort, the user has to source gdb.py and,
call `sgxstate auto [<executable path>]`
### auto subcommand
1. It needs tcs address to be present in rbx register.
2. It calls `sgxstate tcs $rbx`
3. We can optionally provide path to the executable, in which case it will load the symbols.
### Runner:
In case of a panic, the enclave-runner would raise sigtrap with tcs address in Rbx.
In case there is no gdb attached, the runner's signal handler will continue execution.

### DIFF
--- a/enclave-runner/src/usercalls/mod.rs
+++ b/enclave-runner/src/usercalls/mod.rs
@@ -4,6 +4,9 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
+extern crate libc;
+extern crate nix;
+
 use std::alloc::{GlobalAlloc, Layout, System};
 use std::cell::RefCell;
 use std::collections::VecDeque;
@@ -24,12 +27,17 @@ use fnv::FnvHashMap;
 use fortanix_sgx_abi::*;
 
 use sgxs::loader::Tcs as SgxsTcs;
+lazy_static! {
+    static ref DEBUGGER_TOGGLE_SYNC: Mutex<()> = Mutex::new(());
+}
 
 pub mod abi;
 mod interface;
 
 use self::abi::dispatch;
 use self::interface::{Handler, OutputBuffer};
+use self::libc::*;
+use self::nix::sys::signal;
 use loader::{EnclavePanic, ErasedTcs};
 use tcs;
 
@@ -231,7 +239,7 @@ struct RunningTcs {
 
 enum EnclaveKind {
     Command(Command),
-    Library(Library)
+    Library(Library),
 }
 
 struct Command {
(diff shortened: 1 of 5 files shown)


enclave-runner/src/usercalls/mod.rs
```diff
@@ -411,6 +419,61 @@ enum EnclaveEntry {
...
+}
+
+fn debugger_toggle(tcs: usize) {
+    let _g = DEBUGGER_TOGGLE_SYNC.lock().unwrap();
+    let hdl = self::signal::SigHandler::SigAction(handle_trap);
+    let sig_action = signal::SigAction::new(hdl, signal::SaFlags::empty(), signal::SigSet::empty());
+    // Synchronized
+    unsafe {
+        let old = signal::sigaction(signal::SIGTRAP, &sig_action).unwrap();
+        println!("TCS Addr is {:x}", tcs);
```
jethrogb: Remove stray print

enclave-runner/src/usercalls/mod.rs
```diff
@@ -411,6 +419,61 @@ enum EnclaveEntry {
...
+    RSP,
+    RIP,
+    EFL,
+    CSGSFS, /* Actually short cs, gs, fs, __pad0. */
+    ERR,
+    TRAPNO,
+    OLDMASK,
+    CR2,
+}
+extern "C" fn handle_trap(_signo: c_int, _info: *mut siginfo_t, context: *mut c_void) {
```
jethrogb: Add comment explaining what this does

enclave-runner/src/usercalls/mod.rs
```diff
@@ -411,6 +419,61 @@ enum EnclaveEntry {
...
+        let rip = &mut context.uc_mcontext.gregs[Greg::RIP as usize];
+        let inst: *const u8 = *rip as _;
+        if *inst == 0xcc {
+            *rip += 1;
+        }
+    }
+    return;
+}
+
+fn debugger_toggle(tcs: usize) {
```
jethrogb: Rename: `trap_attached_debugger`

jethrogb: Add comment how this works

scripts/gdb.py
```diff
@@ -43,6 +44,20 @@ def find_vma_base(addr):
         return start-offset
   return None
 
+def get_text_offset(file_name):
+  #command=['readelf', '-SW', file_name]
```
jethrogb: Can you not make it work with an array? It would be better to avoid string parsing by the shell

scripts/gdb.py
```diff
@@ -43,6 +44,20 @@ def find_vma_base(addr):
...
+  command="readelf -SW {}".format(file_name)
+  proc = Popen(command, stdout=PIPE, stderr=PIPE, shell=True)
+  sections, err = proc.communicate()
+  if proc.returncode != 0:
+      raise Exception("ELF Read Error")
+  for line in sections.decode('utf-8').rstrip().split('\n'):
+      if ".text" in line:
+          text_info = line.split()
+          break
+  offset=text_info[5]
```
jethrogb: Add comment:
```
0 1    2    3    4       5
  [Nr] Name Type Address Off
```

scripts/gdb.py
```diff
@@ -128,6 +143,17 @@ def invoke (self, arg, from_tty):
...
+    elif args[0] == 'auto':
+      """
+      This should be called after the runner signals SIGTRAP with tcs address in RBX.
+      The user could optionally provide path to the executable, in which case the symbol mapping will be loaded.
+      """
+      tcs=int(gdb.parse_and_eval("$rbx"));
+      gdb.execute("sgxstate tcs {}".format(tcs))
+      if len(args) >= 2:
+          offset=get_text_offset(args[1])
+          address=find_vma_base(tcs) + int(offset, 16)
```
jethrogb: Do `int(offset ,16)` inside `get_text_offset`

scripts/gdb.py
```diff
@@ -43,6 +44,20 @@ def find_vma_base(addr):
...
+def get_text_offset(file_name):
+  #command=['readelf', '-SW', file_name]
+  command="readelf -SW {}".format(file_name)
+  proc = Popen(command, stdout=PIPE, stderr=PIPE, shell=True)
+  sections, err = proc.communicate()
+  if proc.returncode != 0:
+      raise Exception("ELF Read Error")
+  for line in sections.decode('utf-8').rstrip().split('\n'):
+      if ".text" in line:
+          text_info = line.split()
```
jethrogb: Split first, then check if `text_info[2] == ".text"`

enclave-runner/src/usercalls/mod.rs
```diff
@@ -24,12 +27,17 @@ use fnv::FnvHashMap;
 use fortanix_sgx_abi::*;
 
 use sgxs::loader::Tcs as SgxsTcs;
+lazy_static! {
+    static ref DEBUGGER_TOGGLE_SYNC: Mutex<u8> = Mutex::new(0);
```
jethrogb: `Mutex<()>`

scripts/gdb.py
VardhanThigle: Thanks. That works too. Was not sure which one would be more preferable.
Is the non shell mode better to avoid shell interpreting stuff like "*" ?

jethrogb: Exactly

scripts/gdb.py
VardhanThigle: Actually there is one potential problem (number of digits could change column position):
```
0  1  2                 3                         4           5
[  1] .text             PROGBITS         0000000000000240  00000240
```
I guess I should strip the part before section name to make it simpler? Let me know your thoughts.

jethrogb: I see. Then, use a regular expression

scripts/gdb.py
VardhanThigle: might have changed with regular expressions, please let me know after I push the code.

scripts/gdb.py
```diff
@@ -43,6 +44,29 @@ def find_vma_base(addr):
...
+          # 0     1    2    3       4
+          # [Nr]  Name Type Address Off
+          text_info=re.split('^\s+\[\s*(\d+)\]\s+(\.text)\s+(\S+)\s+([0-9a-fA-F]+)\s+([0-9a-fA-F]+)\s', line)
+          break
+
+  # text_info layout
+  # 0  1    2    3    4       5
+  # '' Nr   Name Type Address Off
+  offset = int(text_info[5], 16)
+  return offset
```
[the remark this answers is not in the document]
VardhanThigle: Works. Thanks a lot.

VardhanThigle:
r=jerhrogb

VardhanThigle:
r=jethrogb

VardhanThigle:
bors r=jethrogb

PR #85, opened 2019-01-29
Resolve deadlock in enclave-runner when joining on panicked threads in the enclave

jethrogb:


jethrogb:
Self-reviewing because we need this to land before the release

bors r+

PR #89, opened 2019-02-01
Win ftx runner

parthsane:
This still requires a cargo update before it can build the runner

### DIFF
--- a/aesm-client/Cargo.toml
+++ b/aesm-client/Cargo.toml
@@ -34,12 +34,18 @@ lazy_static = "1"          # MIT/Apache-2.0
 protobuf = "1.2.2"         # MIT/Apache-2.0
 failure = "0.1.1"          # MIT/Apache-2.0
 failure_derive = "0.1.1"   # MIT/Apache-2.0
+
+[target.'cfg(unix)'.dependencies]
 # We require a version of unix-socket with the following change:
 # https://github.com/rust-lang-nursery/unix-socket/pull/30 . Alternatively, in
 # the future, https://github.com/rust-lang/rust/issues/42048 might provide std
 # support. In addition, we need UnixStream::connect_timeout, which may be
 # provided by https://github.com/rust-lang/rust/issues/53615.
 unix_socket2 = "0.5.4"     # MIT/Apache-2.0
+
+[target.'cfg(windows)'.dependencies]
+winapi = { version = "0.3.7", features = ["combaseapi", "enclaveapi", "memoryapi", "objbase"] }
+sgx-isa = { version = "0.3.0", path = "../sgx-isa" }
 
 [build-dependencies]
 protoc-rust = "1.4" # MIT/Apache-2.0
--- a/aesm-client/src/imp/windows.rs
+++ b/aesm-client/src/imp/windows.rs
@@ -1,0 +1,393 @@
+use std::io::{Error as IoError, ErrorKind, Result as IoResult};
+use std::ptr;
+
+use winapi::_core::ffi::c_void;
+use winapi::shared::basetsd::UINT32;
+use winapi::shared::guiddef::{CLSID, IID};
+use winapi::shared::minwindef::ULONG;
+use winapi::shared::ntdef::HRESULT;
+use winapi::shared::winerror::SUCCEEDED;
+use winapi::um::combaseapi::{CoCreateInstance, CoInitializeEx, CoUninitialize, CLSCTX_ALL};
+use winapi::um::objbase::{COINIT_DISABLE_OLE1DDE, COINIT_MULTITHREADED};
+
+use error::{Error, Result};
+use {QuoteInfo, QuoteResult, QuoteType};
+
+type AesmError = UINT32;
+
+#[repr(C)]
+#[derive(Debug, Copy, Clone)]
+struct InnerAesmInterface {
+    vtbl: *mut AesmInterfaceVtbl,
+}
+
+#[repr(C)]
+#[derive(Copy, Clone)]
+struct AesmInterfaceVtbl {
+    query_interface: Option<
+        unsafe extern "system" fn(
+            this: *mut InnerAesmInterface,
+            riid: *const IID,
+            object: *mut *mut ::std::os::raw::c_void,
+        ) -> HRESULT,
+    >,
+    add_ref: Option<unsafe extern "system" fn(this: *mut InnerAesmInterface) -> ULONG>,
+    release: Option<unsafe extern "system" fn(this: *mut InnerAesmInterface) -> ULONG>,
+    get_license_token: Option<
+        unsafe extern "system" fn(
+            this: *mut InnerAesmInterface,
+            mrenclave: *const u8,
+            mrenclave_size: u32,
+            public_key: *const u8,
+            public_key_size: u32,
+            se_attributes: *const u8,
+            se_attributes_size: u32,
+            lictoken: *mut u8,
+            lictoken_size: u32,
+            result: *mut AesmError,
+        ) -> HRESULT,
+    >,
+    init_quote: Option<
+        unsafe extern "system" fn(
+            this: *mut InnerAesmInterface,
+            target_info: *mut u8,
+            target_info_size: u32,
+            gid: *mut u8,
+            gid_size: u32,
+            result: *mut AesmError,
+        ) -> HRESULT,
+    >,
+    get_quote: Option<
+        unsafe extern "system" fn(
+            this: *mut InnerAesmInterface,
+            report: *const u8,
+            report_size: u32,
+            type_: u32,
+            spid: *const u8,
+            spid_size: u32,
+            nonce: *const u8,
+            nonce_size: u32,
+            sig_rl: *const u8,
+            sig_rl_size: u32,
+            qe_report: *mut u8,
+            qe_report_size: u32,
+            quote: *mut u8,
+            buf_size: u32,
+            result: *mut AesmError,
+        ) -> HRESULT,
+    >,
+    create_session: Option<
+        unsafe extern "system" fn(
+            this: *mut InnerAesmInterface,
+            session_id: *mut u32,
+            se_dh_msg1: *mut u8,
+            se_dh_msg1_size: u32,
+            result: *mut AesmError,
+        ) -> HRESULT,
+    >,
+    exchange_report: Option<
+        unsafe extern "system" fn(
+            this: *mut InnerAesmInterface,
+            session_id: u32,
+            se_dh_msg2: *mut u8,
+            se_dh_msg2_size: u32,
+            se_dh_msg3: *mut u8,
+            se_dh_msg3_size: u32,
+            result: *mut AesmError,
+        ) -> HRESULT,
+    >,
+    close_session: Option<
+        unsafe extern "system" fn(
+            this: *mut InnerAesmInterface,
+            session_id: u32,
+            result: *mut AesmError,
+        ) -> HRESULT,
+    >,
+    invoke_service: Option<
+        unsafe extern "system" fn(
+            this: *mut InnerAesmInterface,
+            pse_message_req: *mut u8,
+            pse_message_req_size: u32,
+            pse_message_resp: *mut u8,
+            pse_message_resp_size: u32,
+            result: *mut AesmError,
+        ) -> HRESULT,
+    >,
+    report_attestation_status: Option<
+        unsafe extern "system" fn(
+            this: *mut InnerAesmInterface,
+            platform_info: *mut u8,
+            platform_info_size: u32,
+            attestation_status: u32,
+            update_info: *mut u8,
+            update_info_size: u32,
+            result: *mut AesmError,
+        ) -> HRESULT,
+    >,
+    get_ps_cap: Option<
+        unsafe extern "system" fn(
+            this: *mut InnerAesmInterface,
+            ps_cap: *mut u64,
+            result: *mut AesmError,
+        ) -> HRESULT,
+    >,
+    sgx_register: Option<
+        unsafe extern "system" fn(
+            this: *mut InnerAesmInterface,
+            white_list_cert: *mut u8,
+            white_list_cert_size: u32,
+            registration_data_type: u32,
+            result: *mut AesmError,
+        ) -> HRESULT,
+    >,
+    proxy_setting_assist: Option<
+        unsafe extern "system" fn(
+            this: *mut InnerAesmInterface,
+            proxy_info: *mut u8,
+            proxy_size: u32,
+            result: *mut AesmError,
+        ) -> HRESULT,
+    >,
+    query_sgx_status: Option<
+        unsafe extern "system" fn(
+            this: *mut InnerAesmInterface,
+            sgx_status: *mut u32,
+            result: *mut AesmError,
+        ) -> HRESULT,
+    >,
+    get_whitelist_size: Option<
+        unsafe extern "system" fn(
+            this: *mut InnerAesmInterface,
+            white_list_size: *mut u32,
+            result: *mut AesmError,
+        ) -> HRESULT,
+    >,
+    get_white_list: Option<
+        unsafe extern "system" fn(
+            this: *mut InnerAesmInterface,
+            white_list: *mut u8,
+            buf_size: u32,
+            result: *mut AesmError,
+        ) -> HRESULT,
+    >,
+    get_sec_domain_id: Option<
+        unsafe extern "system" fn(
+            this: *mut InnerAesmInterface,
+            sec_domain_id: *mut u32,
+            result: *mut AesmError,
+        ) -> HRESULT,
+    >,
+    switch_sec_domain: Option<
+        unsafe extern "system" fn(
+            this: *mut InnerAesmInterface,
+            sec_domain_id: u32,
+            result: *mut AesmError,
+        ) -> HRESULT,
+    >,
+    get_epid_provision_status: Option<
+        unsafe extern "system" fn(
+            this: *mut InnerAesmInterface,
+            epid_pr_status: *mut u32,
+            result: *mut AesmError,
+        ) -> HRESULT,
+    >,
+    get_platform_service_status: Option<
+        unsafe extern "system" fn(
+            this: *mut InnerAesmInterface,
+            pse_status: *mut u32,
+            result: *mut AesmError,
+        ) -> HRESULT,
+    >,
+}
+
+const CLSID_AESMINTERFACE: CLSID = CLSID {
+    Data1: 0x82367CAB,
+    Data2: 0xF2B9,
+    Data3: 0x461A,
+    Data4: [0xB6, 0xC6, 0x88, 0x9D, 0x13, 0xEF, 0xC6, 0xCA],
+};
+const IID_IAESMINTERFACE: IID = IID {
+    Data1: 0x50AFD900,
+    Data2: 0xF309,
+    Data3: 0x4557,
+    Data4: [0x8F, 0xCB, 0x10, 0xCF, 0xAB, 0x80, 0x2C, 0xDD],
+};
+
+trait HresultExt: Sized {
+    fn into_io_error(self) -> IoResult<Self>;
+}
+
+impl HresultExt for HRESULT {
+    fn into_io_error(self) -> IoResult<HRESULT> {
+        if SUCCEEDED(self) {
+            Ok(self)
+        } else {
+            Err(IoError::from_raw_os_error(self))
+        }
+    }
+}
+
+#[derive(Clone, Debug, Default)]
+pub struct AesmClient {
+    _private: (),
+}
+
+struct AesmInterface {
+    inner: *mut InnerAesmInterface,
+}
+
+impl Drop for AesmInterface {
+    fn drop(&mut self) {
+        unsafe {
+            if let Some(release) = (*(*self.inner).vtbl).release {
+                release(self.inner);
+            }
+            CoUninitialize();
+        }
+    }
+}
+
+impl AesmClient {
+    pub fn new() -> Self {
+        Default::default()
+    }
+
+    pub fn try_connect(&self) -> Result<()> {
+        self.create_instance().map(|_| ())
+    }
+
+    fn create_instance(&self) -> Result<AesmInterface> {
+        let mut interface: *mut InnerAesmInterface = std::ptr::null_mut();
+        unsafe {
+            CoInitializeEx(
+                ptr::null_mut(),
+                COINIT_MULTITHREADED | COINIT_DISABLE_OLE1DDE,
+            )
+            .into_io_error()
+            .map_err(|e| {
+                IoError::new(ErrorKind::Other, format!("Failed to initialize COM: {}", e))
+            })?;
+
+            CoCreateInstance(
+                &CLSID_AESMINTERFACE,
+                ptr::null_mut(),
+                CLSCTX_ALL,
+                &IID_IAESMINTERFACE,
+                &mut interface as *mut _ as *mut *mut c_void,
+            )
+            .into_io_error()
+            .map_err(|e| {
+                if let Some(release) = (*(*interface).vtbl).release {
+                    release(interface);
+                }
+                CoUninitialize();
+                IoError::new(
+                    ErrorKind::Other,
+                    format!("Fail to create Aesm Interface {}", e),
+                )
+            })?;
+        }
+        Ok(AesmInterface { inner: interface })
+    }
+
+    pub fn init_quote(&self) -> Result<QuoteInfo> {
+        let mut target_info: Vec<u8> = vec![0; sgx_isa::Targetinfo::UNPADDED_SIZE];
+        let mut gid: Vec<u8> = vec![0; 4];
+        let mut error: AesmError = 0;
+        unsafe {
+            let interface = self.create_instance()?;
+            if let Some(init_quote) = (*(*interface.inner).vtbl).init_quote {
+                init_quote(
+                    interface.inner,
+                    target_info.as_mut_ptr(),
+                    target_info.len() as _,
+                    gid.as_mut_ptr(),
+                    gid.len() as _,
+                    &mut error as _,
+                )
+                .into_io_error()?;
+                if error != 0 {
+                    return Err(Error::AesmCode(error.into()));
+                }
+            }
+        }
+        let quote_info: QuoteInfo = QuoteInfo { target_info, gid };
+
+        return Ok(quote_info);
+    }
+
+    pub fn get_quote(
+        &self,
+        session: &QuoteInfo,
+        report: Vec<u8>,
+        spid: Vec<u8>,
+        sig_rl: Vec<u8>,
+    ) -> Result<QuoteResult> {
+        use sgx_isa::Report;
+        let nonce = [0u8; 64];
+        let quote_buffer_size = session.quote_buffer_size(&sig_rl);
+        let mut qe_report: Vec<u8> = vec![0; Report::UNPADDED_SIZE];
+        let mut quote: Vec<u8> = vec![0; quote_buffer_size as usize];
+        let mut error: AesmError = 0;
+
+        unsafe {
+            let interface = self.create_instance()?;
+            if let Some(get_quote) = (*(*interface.inner).vtbl).get_quote {
+                get_quote(
+                    interface.inner,
+                    report.as_ptr(),
+                    report.len() as _,
+                    QuoteType::Linkable.into(),
+                    spid.as_ptr(),
+                    spid.len() as _,
+                    nonce.as_ptr(),
+                    nonce.len() as _,
+                    sig_rl.as_ptr(),
+                    sig_rl.len() as _,
+                    qe_report.as_mut_ptr(),
+                    qe_report.len() as _,
+                    quote.as_mut_ptr(),
+                    quote_buffer_size,
+                    &mut error as _,
+                )
+                .into_io_error()?;
+                if error != 0 {
+                    return Err(Error::AesmCode(error.into()));
+                }
+            }
+        }
+        return Ok(QuoteResult::new(quote, qe_report));
+    }
+
+    pub fn get_launch_token(
+        &self,
+        mr_enclave: Vec<u8>,
+        signer_modulus: Vec<u8>,
+        attributes: Vec<u8>,
+    ) -> Result<Vec<u8>> {
+        let mut licence_token = vec![0; sgx_isa::Einittoken::UNPADDED_SIZE];
+        let mut error: AesmError = 0;
+        unsafe {
+            let interface = self.create_instance()?;
+            if let Some(get_license_token) = (*(*interface.inner).vtbl).get_license_token {
+                get_license_token(
+                    interface.inner,
+                    mr_enclave.as_ptr(),
+                    mr_enclave.len() as _,
+                    signer_modulus.as_ptr(),
+                    signer_modulus.len() as _,
+                    attributes.as_ptr(),
+                    attributes.len() as _,
+                    licence_token.as_mut_ptr(),
+                    licence_token.len() as _,
+                    &mut error as _,
+                )
+                .into_io_error()?;
+                if error != 0 {
+                    return Err(Error::AesmCode(error.into()));
+                }
+            }
+        }
+        return Ok(licence_token);
+    }
+}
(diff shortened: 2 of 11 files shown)


sgxs-loaders/src/enclaveapi/mod.rs
```diff
@@ -0,0 +1,268 @@
+use std::{mem, ptr};
```
jethrogb: Run `rustfmt` on this file

sgxs-loaders/src/enclaveapi/mod.rs
```diff
@@ -0,0 +1,268 @@
...
+
+        if eadd
+            .secinfo
+            .flags
+            .intersects(SecinfoFlags::PENDING | SecinfoFlags::MODIFIED | SecinfoFlags::PR)
+        {
+            return Err(Error::Add(ErrorKind::InvalidInput.into()));
+        }
+        let mut flags : winapi::shared::minwindef::DWORD = 0;
+        match (eadd.secinfo.flags & (SecinfoFlags::R | SecinfoFlags::W | SecinfoFlags::X)).bits() as u8 {
```
jethrogb: `as u8` is not necessary

sgxs-loaders/src/enclaveapi/mod.rs
```diff
@@ -0,0 +1,268 @@
...
+            0b000 => {},
+            0b001 => flags = PAGE_READONLY,
+            0b010 => {return Err(Error::Add(ErrorKind::InvalidInput.into()))},
+            0b011 => flags = PAGE_READWRITE,
+            0b100 => flags = PAGE_EXECUTE,
+            0b101 => flags = PAGE_EXECUTE_READ,
+            0b110 => {return Err(Error::Add(ErrorKind::InvalidInput.into()))},
+            0b111 => flags = PAGE_EXECUTE_READWRITE,
+            0b1000...255 => unreachable!(),
```
jethrogb: You can write `0b1000...std::u64::MAX`

sgxs-loaders/src/enclaveapi/mod.rs
jethrogb: Write `let mut flags = match ...`

sgxs-loaders/src/enclaveapi/mod.rs
```diff
@@ -0,0 +1,268 @@
...
+                if eadd.secinfo.flags.contains(SecinfoFlags::R | SecinfoFlags::W | SecinfoFlags::X) {
+                    return Err(Error::Add(ErrorKind::InvalidInput.into()))
+                }
+                // NOTE: For some reason the windows API needs the Read flag set but then removes it
+                flags = PAGE_ENCLAVE_THREAD_CONTROL|PAGE_READWRITE;
+            },
+            _ => return Err(Error::Add(ErrorKind::InvalidInput.into())),
+        }
+        match chunks.0 {
+            0 => flags = flags | PAGE_ENCLAVE_UNVALIDATED,
```
jethrogb: `flags |= ...`

sgxs-loaders/src/enclaveapi/mod.rs
```diff
@@ -0,0 +1,268 @@
...
+            0xffff => {}
+            _ => return Err(Error::Add(ErrorKind::InvalidInput.into())),
```
jethrogb: `IoError::new(ErrorKind::Other, "Partially-measured pages not supported on Windows")`

sgxs-loaders/src/enclaveapi/mod.rs
```diff
@@ -0,0 +1,268 @@
...
+            0b1000...255 => unreachable!(),
+        }
+
+        match PageType::from_repr(eadd.secinfo.flags.page_type()) {
+            Some(PageType::Reg) => {}
+            Some(PageType::Tcs) => {
+                if eadd.secinfo.flags.contains(SecinfoFlags::R | SecinfoFlags::W | SecinfoFlags::X) {
+                    return Err(Error::Add(ErrorKind::InvalidInput.into()))
+                }
+                // NOTE: For some reason the windows API needs the Read flag set but then removes it
```
jethrogb: I don't remember the exact details, but I believe it's like this, is that right? You *must* call `LoadEnclaveData` with `PAGE_ENCLAVE_THREAD_CONTROL|PAGE_READWRITE` but Windows will call EADD with only `SecinfoFlags::R`? If that is the case, the error checking here is not sufficient.

sgxs-loaders/src/enclaveapi/mod.rs
```diff
@@ -0,0 +1,268 @@
...
+            0xffff => {}
+            _ => return Err(Error::Add(ErrorKind::InvalidInput.into())),
+        }
+        unsafe {
+            let mut data_loaded : usize = 0;
+            let curhandle: HANDLE = GetCurrentProcess();
+            let ret = LoadEnclaveData(
+                curhandle,
+                (mapping.base + eadd.offset) as _,
+
```
jethrogb: Stray newline

sgxs-loaders/src/enclaveapi/mod.rs
```diff
@@ -0,0 +1,268 @@
...
+    ) -> Result<(), Self::Error> {
+        let mut init_info : ENCLAVE_INIT_INFO_SGX = ENCLAVE_INIT_INFO_SGX {
+            SigStruct : [0 ; 1808],
+            Reserved1 : [0 ; 240],
+            EInitToken: [0 ; 304],
+            Reserved2 : [0 ; 1744],
+        };
+        init_info.SigStruct.clone_from_slice(&sigstruct.as_ref());
+        if let Some(e) = einittoken {
+            init_info.EInitToken.clone_from_slice(e.as_ref());
```
jethrogb: Nit: you can use `copy_from_slice` here (twice)

sgxs-loaders/src/enclaveapi/mod.rs
```diff
@@ -0,0 +1,268 @@
...
+    ) -> Result<Mapping<Self>, Self::Error> {
+        let secs = Secs {
+            size: ecreate.size,
+            ssaframesize: ecreate.ssaframesize,
+            miscselect,
+            attributes,
+            ..Default::default()
+        };
+
+        let curhandle: HANDLE = unsafe {GetCurrentProcess()};
```
jethrogb: Just do the call to `GetCurrentProcess()` directly as a function argument of `CreateEnclave`. You can do this everywhere.

sgxs-loaders/src/enclaveapi/mod.rs
```diff
@@ -0,0 +1,268 @@
...
+pub struct Sgx {
+    inner: generic::Device<WinInnerLibrary>,
+}
+
+pub struct DeviceBuilder {
+    inner: generic::DeviceBuilder<WinInnerLibrary>,
+}
+
+impl Sgx {
+    pub fn open() -> IoResult<DeviceBuilder> {
```
jethrogb: Let's call this `new`

sgxs-loaders/src/sgx_enclave_common/mod.rs
```diff
@@ -244,13 +246,26 @@ impl EnclaveLoad for InnerLibrary {
...
             );
 
+            #[cfg(windows)]
+                {
+                    let mut prev_protection : winapi::shared::minwindef::DWORD = 0;
+                    if winapi::um::memoryapi::VirtualProtect(
+                        mapping.base as _,
+                        mapping.size as _,
+                        winapi::um::winnt::PAGE_EXECUTE_READWRITE,
+                        &mut prev_protection
```
jethrogb: Can you pass `ptr::null_mut` instead?

sgxs-loaders/src/sgx_enclave_common/mod.rs
```diff
@@ -244,13 +246,26 @@ impl EnclaveLoad for InnerLibrary {
...
             libc::mprotect(
                 mapping.base as _,
                 mapping.size as _,
                 libc::PROT_READ | libc::PROT_WRITE | libc::PROT_EXEC,
             );
 
+            #[cfg(windows)]
+                {
+                    let mut prev_protection : winapi::shared::minwindef::DWORD = 0;
+                    if winapi::um::memoryapi::VirtualProtect(
```
jethrogb: Did you try without this and it failed?

sgxs-loaders/src/sgx_enclave_common/mod.rs
```diff
@@ -244,13 +246,26 @@ impl EnclaveLoad for InnerLibrary {
             ) {
                 return Err(Error::Init(error.into()));
             }
-
+            #[cfg(unix)]
             libc::mprotect(
                 mapping.base as _,
                 mapping.size as _,
                 libc::PROT_READ | libc::PROT_WRITE | libc::PROT_EXEC,
             );
```
jethrogb: Not sure why this wasn't here yet already, please add error handling similar to the Windows case.

sgxs-tools/Cargo.toml
```diff
@@ -14,7 +14,7 @@ autobins = true
 
 [[bin]]
 name = "isgx-pe2sgx"
-required-features = ["pe2sgxs"]
+required-features = ["pe2sgx"]
```
jethrogb: Why are you renaming this feature? The name was correct.

aesm-client/src/win_aesm_client.rs
```diff
@@ -0,0 +1,349 @@
+use std::ptr;
+use winapi::_core::ffi::c_void;
+use winapi::shared::basetsd::UINT32;
+use winapi::shared::guiddef::{CLSID, IID};
+use winapi::shared::ntdef::HRESULT;
+use winapi::shared::minwindef::ULONG;
+use winapi::shared::winerror::{S_OK, S_FALSE};
+use winapi::um::combaseapi::{CoInitializeEx, CoCreateInstance, CoUninitialize, CLSCTX_ALL};
+use winapi::um::objbase::{COINIT_MULTITHREADED, COINIT_DISABLE_OLE1DDE};
+pub type AesmErrorT = UINT32;
```
jethrogb: Throughout this file: remove `pub` when not needed.

aesm-client/src/win_aesm_client.rs
```diff
@@ -0,0 +1,349 @@
...
+pub type AesmInterfaceT = _aesm_interface;
```
jethrogb: Instead of a type alias, just name the struct this. But get rid of the `T`.

aesm-client/src/win_aesm_client.rs
```diff
@@ -0,0 +1,349 @@
...
+use {QuoteResult, QuoteType, QuoteInfo, AesmClient};
+#[repr(C)]
+#[derive(Debug, Copy, Clone)]
+pub struct _aesm_interface {
+    pub vtbl: *mut aesm_interface_vtbl,
+}
+
+#[repr(C)]
+#[derive(Copy, Clone)]
+pub struct aesm_interface_vtbl {
```
jethrogb: `AesmInterfaceVtbl`

aesm-client/src/win_aesm_client.rs
```diff
@@ -0,0 +1,349 @@
...
+    pub query_interface: ::std::option::Option<
```
jethrogb: Throughout this `struct`: just write `Option`

aesm-client/src/win_aesm_client.rs
jethrogb: Rename to `AesmError`

aesm-client/src/win_aesm_client.rs
```diff
@@ -0,0 +1,349 @@
...
+    pub get_platform_service_status: ::std::option::Option<
+        unsafe extern "system" fn(
+            this: *mut AesmInterfaceT,
+            pse_status: *mut u32,
+            result: *mut AesmErrorT,
+        ) -> HRESULT,
+    >,
+}
+
+#[cfg(windows)] impl Drop for AesmClient {
```
jethrogb: Get rid of all `cfg` in this file, this file already has a top-level `cfg`

aesm-client/src/win_aesm_client.rs
```diff
@@ -0,0 +1,349 @@
+use std::ptr;
```
jethrogb: Run `rustfmt` on this file

aesm-client/src/win_aesm_client.rs
```diff
@@ -0,0 +1,349 @@
...
+            Data2: 0xF2B9,
+            Data3: 0x461A,
+            Data4: [0xB6, 0xC6, 0x88, 0x9D, 0x13, 0xEF, 0xC6, 0xCA]
+        };
+        let iid_iaesminterface : IID = IID {
+            Data1: 0x50AFD900,
+            Data2: 0xF309,
+            Data3: 0x4557,
+            Data4: [0x8F, 0xCB, 0x10, 0xCF, 0xAB, 0x80, 0x2C, 0xDD]
+        };
```
jethrogb: Make these two definitions `const` and put them somewhere near the vtbl definition.

aesm-client/src/win_aesm_client.rs
```diff
@@ -0,0 +1,349 @@
...
+        unsafe {
+            let res = CoInitializeEx(
+                ptr::null_mut(),
+                COINIT_MULTITHREADED | COINIT_DISABLE_OLE1DDE
+            );
+            if res != S_OK && res != S_FALSE {
+                return Err(Error::AesmBadResponse(("Fail to initialize Com interface").to_string()));
```
jethrogb: `Error::AesmCommunication` (with `io::ErrorKind::Other` and this string), also below

aesm-client/src/win_aesm_client.rs
```diff
@@ -0,0 +1,349 @@
...
+            }
+        }
+        Ok(AesmClient {
+            interface: instance
+        })
+    }
+
+    pub fn init_quote(&self) -> Result<QuoteInfo> {
+        let mut target_info : Vec<u8> = vec![0; sgx_isa::Targetinfo::UNPADDED_SIZE];
+        let mut gid: Vec<u8> = vec![0; 4usize];
```
jethrogb: Don't need to put `usize` suffix here

aesm-client/src/win_aesm_client.rs
```diff
@@ -0,0 +1,349 @@
...
+        &self,
+        session: &QuoteInfo,
+        report: Vec<u8>,
+        spid: Vec<u8>,
+        sig_rl: Vec<u8>,
+    ) -> Result<QuoteResult> {
+        use sgx_isa::Report;
+        let nonce : Vec<u8> = vec![0; 64];
+        let quote_buffer_size = session.quote_buffer_size(&sig_rl);
+        let mut qe_report:Vec<u8> = vec![0; Report::UNPADDED_SIZE as usize];
```
jethrogb: Don't need `as usize`

aesm-client/src/win_aesm_client.rs
```diff
@@ -0,0 +1,349 @@
...
+
+    pub fn get_quote(
+        &self,
+        session: &QuoteInfo,
+        report: Vec<u8>,
+        spid: Vec<u8>,
+        sig_rl: Vec<u8>,
+    ) -> Result<QuoteResult> {
+        use sgx_isa::Report;
+        let nonce : Vec<u8> = vec![0; 64];
```
jethrogb: Don't need to allocate, can just use `let nonce = [0u8; 64]`;

aesm-client/src/lib.rs
```diff
@@ -38,23 +33,19 @@ use sgxs::sigstruct::{Attributes, Sigstruct};
...
 pub use error::{AesmError, Error, Result};
-
+#[cfg(windows)]
+extern crate winapi;
+#[cfg(windows)]
+extern crate sgx_isa;
+#[cfg(windows)]
+mod win_aesm_client;
+#[cfg(unix)]
+mod unix_aesm_client;
```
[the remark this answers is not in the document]
parthsane: I have one question about this, the windows implementation doesn't have a `with_path` function. Should there be a dummy implementation for that? Or should that function be omitted by using  `#[cfg(unix)]`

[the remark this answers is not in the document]
jethrogb: Good point. I think we want to use the same mechanism that `std` uses with the `Ext` traits. This forces people to explicitly opt-in to non-cross-platform behavior. See e.g. https://doc.rust-lang.org/std/os/unix/ffi/trait.OsStringExt.html

sgxs-loaders/src/sgx_enclave_common/mod.rs
parthsane: I don't think I had tried it before. I remember it was mostly done to ensure the same steps happen for both OS

I tried it now.
It works correctly without the VirtualProtect function.

parthsane: Should I remove it or keep it as it is?

jethrogb: Remove. I don't remember why the mprotect was necessary on Linux, but it is.

sgxs-loaders/src/enclaveapi/mod.rs
parthsane: I think you are right about the flag requirements for calling LoadEnclaveData. 
However, I am unsure of the following.
1) What are the flags that the windows loader should get for a `PageType::Tcs`?
what does the linux loader get? 
2) Where are these flags set? From what I see, they are defined in the `.sig` file. Where is the .sig file generated?
3) If the same `.sig` file should work on both linux and windows, then shouldn't the windows loader accept the same flags as the linux loader?

jethrogb: When you say `.sig`, do you mean `.sgxs`?

Yes Windows should be accepting whatever flag we want, but it doesn't, and there's nothing we can do about it. So our only recourse is throwing an error if we know it won't work...

parthsane: Ok.
 I will change the error checking to check if `eadd.secinfo.flags` contains both the R & W flags, and if it doesn't then throw an error. 

However (I guess) this would mean that we would have to build sgxs files differently for windows.

jethrogb: > However (I guess) this would mean that we would have to build sgxs files differently for windows.

Why? We already know it works right now.

aesm-client/Cargo.toml
```diff
@@ -28,15 +28,22 @@ lazy_static = "1"          # MIT/Apache-2.0
...
 
 [build-dependencies]
 protoc-rust = "1.4" # MIT/Apache-2.0
 
 [dev-dependencies]
 sgx-isa = { version = "0.2.0-rc1", path = "../sgx-isa" }
+
+[target.'cfg(windows)'.dependencies]
+winapi = { version = "0.3", features = ["combaseapi", "enclaveapi", "memoryapi", "objbase"] }
+sgx-isa = { version = "0.2.0-rc1", path = "../sgx-isa" }
```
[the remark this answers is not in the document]
parthsane: I'll have the correct version number for the sgx-isa after a rebase.

sgxs-loaders/src/enclaveapi/mod.rs
parthsane: The way it was working was that if the PageType is Tcs , we were forcing the flags to be the following. 
```
line number 127: flags = PAGE_ENCLAVE_THREAD_CONTROL|PAGE_READWRITE;
```

In the sgxs file which was built on linux, the value of `eadd.secinfo.flags` for a `PageType::Tcs` was `PT_B0` when i tried to find it's value while loading it on windows.

jethrogb: Yes, that's exactly the behavior I described above?

aesm-client/src/lib.rs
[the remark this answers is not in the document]
parthsane: I've made changes which I believe satisfies the requirement. But I have a feelign you'll have some review comments on it.

aesm-client/src/win_aesm_client.rs
```diff
@@ -0,0 +1,359 @@
...
+    Data4: [0xB6, 0xC6, 0x88, 0x9D, 0x13, 0xEF, 0xC6, 0xCA],
+};
+const IID_IAESMINTERFACE: IID = IID {
+    Data1: 0x50AFD900,
+    Data2: 0xF309,
+    Data3: 0x4557,
+    Data4: [0x8F, 0xCB, 0x10, 0xCF, 0xAB, 0x80, 0x2C, 0xDD],
+};
+
+impl Drop for AesmClient {
```
jethrogb: It's customary to put the `struct` definition above any `impl` blocks.

aesm-client/src/win_aesm_client.rs
```diff
@@ -0,0 +1,359 @@
...
+    Data3: 0x461A,
+    Data4: [0xB6, 0xC6, 0x88, 0x9D, 0x13, 0xEF, 0xC6, 0xCA],
+};
+const IID_IAESMINTERFACE: IID = IID {
+    Data1: 0x50AFD900,
+    Data2: 0xF309,
+    Data3: 0x4557,
+    Data4: [0x8F, 0xCB, 0x10, 0xCF, 0xAB, 0x80, 0x2C, 0xDD],
+};
+
```
jethrogb: Add function:

```rust
use winapi::shared::winerror::SUCCEEDED;

trait HresultExt {
    fn into_io_error(self) -> Result<Self, IoError>;
}

impl HresultExt for HRESULT {
    fn into_io_error(self) -> Result<HRESULT, IoError> {
        if SUCCEEDED(self) {
            Ok(self)
        } else {
            Err(IoError::from_raw_os_error(self))
        }
    }
}
```

Use this everywhere you have a COM function that returns an HRESULT.

aesm-client/src/win_aesm_client.rs
```diff
@@ -0,0 +1,359 @@
...
+        unsafe {
+            let res = CoInitializeEx(
+                ptr::null_mut(),
+                COINIT_MULTITHREADED | COINIT_DISABLE_OLE1DDE,
+            );
+            if res != S_OK && res != S_FALSE {
+                return Err(Error::AesmCommunication(
+                    IoError::new(ErrorKind::Other,"Fail to initialize Com interface"))
+                );
+            }
```
jethrogb:
```rust
CoInitializeEx(
    ptr::null_mut(),
    COINIT_MULTITHREADED | COINIT_DISABLE_OLE1DDE,
)
.into_io_error()
.map_err(|e| IoError::new(ErrorKind::Other, format!("Failed to initialize COM: {}", e)) )?;
```

aesm-client/src/win_aesm_client.rs
```diff
@@ -0,0 +1,359 @@
...
+                ptr::null_mut(),
+                CLSCTX_ALL,
+                &IID_IAESMINTERFACE,
+                &mut instance as *mut _ as *mut *mut c_void,
+            );
+            if res < 0 {
+                return Err(Error::AesmCommunication(
+                    IoError::new(ErrorKind::Other, "Fail to create Aesm Interface"))
+                );
+            }
```
jethrogb:
```rust
CoCreateInstance(
    &CLSID_AESMINTERFACE,
    ptr::null_mut(),
    CLSCTX_ALL,
    &IID_IAESMINTERFACE,
    &mut instance as *mut _ as *mut *mut c_void,
);
.into_io_error()
.map_err(|e| IoError::new(ErrorKind::Other, format!("Failed to initialize AESM interface: {}", e)) )?;
```

aesm-client/src/win_aesm_client.rs
```diff
@@ -0,0 +1,359 @@
...
+        unsafe {
+            if let Some(init_quote) = (*(*self.interface).vtbl).init_quote {
+                let ret = init_quote(
+                    self.interface,
+                    target_info.as_mut_ptr(),
+                    target_info.len() as _,
+                    gid.as_mut_ptr(),
+                    gid.len() as _,
+                    &mut error as _,
+                );
```
jethrogb: For all calls like these: check `ret` first (and convert to IO error), then check `error`.

sgxs-loaders/src/sgx_enclave_common/mod.rs
```diff
@@ -69,6 +69,8 @@ pub enum LibraryError {
     NotInitialized,
     #[fail(display = "Unknown error ({}) in SGX device interface", _0)]
     Other(u32),
+    #[fail(display = "OS error ({}) in SGX device interface", _0)]
+    OS(IoError),
```
jethrogb: Let's make this a little more specific:
```rust
#[fail(display = "Failed to adjust the page table permissions: {}", _0)]
PageTableFailure(IoError), 
```

parthsane:
I think a version change might be needed because there is a change in the interface of AesmClient

aesm-client/src/lib.rs
```diff
@@ -152,157 +147,59 @@ impl QuoteResult {
...
-
-    pub fn with_path<P: AsRef<Path>>(path: P) -> Self {
-        AesmClient {
-            path: Some(path.as_ref().to_owned()),
+    pub fn new() -> Result<Self> {
+        let aesm_client = imp::AesmClient::new();
+        match aesm_client {
+            Ok(inner) => Ok(AesmClient {inner}),
+            Err(e) => Err(e),
         }
```
jethrogb:
```rust
imp::AesmClient::new().map(|inner| AesmClient { inner })
```

aesm-client/src/lib.rs
```diff
@@ -21,19 +21,14 @@ extern crate lazy_static;
...
 extern crate sgxs;
+#[cfg(unix)]
 extern crate unix_socket;
-
-use std::ffi::OsStr;
-use std::io::{Read, Write};
-use std::os::unix::ffi::OsStrExt;
-use std::path::{Path, PathBuf};
+#[cfg(unix)]
+use std::path::Path;
```
jethrogb: This import is probably unused?

aesm-client/src/lib.rs
```diff
@@ -42,23 +37,23 @@ use sgxs::sigstruct::{Attributes, Sigstruct};
...
 include!(concat!(env!("OUT_DIR"), "/mod_aesm_proto.rs"));
 mod error;
-
 use self::aesm_proto::*;
 pub use error::{AesmError, Error, Result};
-
+#[cfg(windows)]
+extern crate winapi;
+#[cfg(windows)]
+extern crate sgx_isa;
```
jethrogb: See coding style guide for the order in which to put things at the top of the module

aesm-client/src/unix/mod.rs
```diff
@@ -0,0 +1,13 @@
+use std::path::Path;
+use imp::AesmClient;
+pub trait AesmClientExt {
+    fn with_path<P: AsRef<Path>>(path: P) -> Self;
+}
+
+impl AesmClientExt for crate::AesmClient {
```
jethrogb: If you put this in `unix_aesm_client.rs`, you don't need to add a separate `AesmClient::with_path` function. Also, then this file has become so small, you can inline it in `lib.rs`

aesm-client/src/unix/unix_aesm_client.rs
```diff
@@ -0,0 +1,170 @@
...
+/// without network interaction.
+const LOCAL_AESM_TIMEOUT_US: u32 = 1_000_000;
+/// This timeout is an argument in AESM request protobufs.
+///
+/// This value should be used for requests that might need interaction with
+/// remote servers, such as provisioning EPID.
+const REMOTE_AESM_TIMEOUT_US: u32 = 30_000_000;
+
+#[derive(Default, Debug, Clone)]
+pub struct AesmClient {
```
jethrogb: `derive(Default)` is no longer useful

aesm-client/src/windows/win_aesm_client.rs
```diff
@@ -0,0 +1,393 @@
...
+                ptr::null_mut(),
+                COINIT_MULTITHREADED | COINIT_DISABLE_OLE1DDE,
+            )
+                .into_io_error()
+                .map_err(|e| Error::AesmCommunication(IoError::new(ErrorKind::Other, format!("Failed to initialize COM: {}", e))) )
+        }
+    }
+
+    pub fn try_connect(&self) -> Result<()> {
+        self.initialize_com().map(|_| ())
```
jethrogb: This doesn't make sense. You will already have called `create_instance` when you get here, which already did this. But see our Slack discussion for possible change in architecture.

aesm-client/src/windows/win_aesm_client.rs
```diff
@@ -0,0 +1,393 @@
...
+                    target_info.len() as _,
+                    gid.as_mut_ptr(),
+                    gid.len() as _,
+                    &mut error as _,
+                ).into_io_error().
+                    map_err(|e| IoError::new(ErrorKind::Other,
+                                             format!("Fail to init Quote. init_quote returned {}", e)) )?;
+                if error != 0 {
+                    return Err(Error::AesmCommunication(IoError::new(ErrorKind::Other,
+                                             format!("Fail to init Quote with error {}", error))));
```
jethrogb: This used to be correct, but now it's not. This error is an AesmCode error. Same thing for other functions below.

aesm-client/src/windows/win_aesm_client.rs
```diff
@@ -0,0 +1,393 @@
...
+            if let Some(init_quote) = (*(*self.interface).vtbl).init_quote {
+                init_quote(
+                    self.interface,
+                    target_info.as_mut_ptr(),
+                    target_info.len() as _,
+                    gid.as_mut_ptr(),
+                    gid.len() as _,
+                    &mut error as _,
+                ).into_io_error().
+                    map_err(|e| IoError::new(ErrorKind::Other,
```
jethrogb: No need to add an error message if the function is only doing a single operation. Same thing for other functions below.

sgxs-loaders/src/enclaveapi/mod.rs
```diff
@@ -0,0 +1,286 @@
...
+            0b111 => PAGE_EXECUTE_READWRITE,
+            0b1000...std::u64::MAX => unreachable!(),
+        };
+        match PageType::from_repr(eadd.secinfo.flags.page_type()) {
+            Some(PageType::Reg) => {}
+            Some(PageType::Tcs) => {
+                if !eadd
+                    .secinfo
+                    .flags
+                    .contains(SecinfoFlags::PT_B0)
```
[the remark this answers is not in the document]
jethrogb: This is not resolved?

parthsane:
@jethrogb, please review this PR

aesm-client/src/windows/win_aesm_client.rs
```diff
@@ -0,0 +1,400 @@
...
+            CoCreateInstance(
+                &CLSID_AESMINTERFACE,
+                ptr::null_mut(),
+                CLSCTX_ALL,
+                &IID_IAESMINTERFACE,
+                &mut interface as *mut _ as *mut *mut c_void,
+            )
+                .into_io_error()
+                .map_err(|e| {
+                    if interface != ptr::null_mut() {
```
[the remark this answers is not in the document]
parthsane: I checked the COM documentation and saw that if there is an error, then the interface pointer is always null.

aesm-client/src/imp/windows.rs
```diff
@@ -0,0 +1,393 @@
...
+            Ok(self)
+        } else {
+            Err(IoError::from_raw_os_error(self))
+        }
+    }
+}
+
+#[derive(Clone, Debug, Default)]
+pub struct AesmClient {
+    dummy_private_var: u32,
```
jethrogb: Use: `_private: ()`

1. I don't know why there is no warning about this field being unused, but it could be because of the `derive`s. Prefixing a field with `_` silences that warning.
2. `()` doesn't take up any space in memory.

aesm-client/src/lib.rs
```diff
@@ -152,72 +150,14 @@ impl QuoteResult {
     }
 }
 
-#[derive(Default, Debug, Clone)]
+#[derive(Debug, Clone)]
```
jethrogb: Bring back `Default`

aesm-client/src/lib.rs
```diff
@@ -21,45 +21,43 @@ extern crate lazy_static;
 extern crate protobuf;
```
jethrogb: `extern crate lazy_static` (above) should be `#[cfg(unix)]`

jethrogb: After `#![doc(...)]` (above), add: `#![deny(warnings)]`

jethrogb:
bors retry

PR #93, opened 2019-02-04
Initial Changes for service API request and test example.

VardhanThigle:
TODO (Description of the changes)


examples/api-extension/runner/src/main.rs
```diff
@@ -0,0 +1,101 @@
...
+arg_enum!{
+    #[derive(PartialEq, Debug)]
+    #[allow(non_camel_case_types)]
+    pub enum Signature {
+        coresident,
+        dummy
+    }
+}
+
+struct FileStream {
```
jethrogb: A file is a poor example because it doesn't really resemble a bi-directional stream.

enclave-runner/src/command.rs
```diff
@@ -12,13 +12,26 @@ use sgxs::loader::{Load, MappingInfo};
...
-#[derive(Debug)]
 pub struct Command {
     main: ErasedTcs,
     threads: Vec<ErasedTcs>,
     address: *mut c_void,
     size: usize,
+    api_ext_impl : Option<Box<ApiExtension>>,
+}
+
+impl fmt::Debug for Command {
```
jethrogb: Just add `ApiExtension: Debug` and put the derive back

examples/api-extension/runner/src/main.rs
```diff
@@ -0,0 +1,101 @@
...
+         let file = OpenOptions::new().write(true).read(true).truncate(true).create(true)
+                             .open(&name)?;
+        let stream = FileStream{f : file};
+        Ok(Box::new(stream))
+    }
+}
+
+
+
+fn main() -> Result<(), Error> {
```
VardhanThigle: This is the copy of the default runner, with addition of `ApiExtension`

enclave-runner/src/usercalls/mod.rs
```diff
@@ -545,6 +553,34 @@ fn trap_attached_debugger(tcs: usize) {
     }
 }
 
+pub trait ApiExtension : 'static + Send + Sync {
+    fn connect_stream(
+        &self,
+        addr: &[u8],
+        local_addr: Option<&mut OutputBuffer>,
+        peer_addr: Option<&mut OutputBuffer>,
```
jethrogb: Can you make these `Option<&mut String>` instead?

examples/api-extension/app/src/main.rs
```diff
@@ -0,0 +1,15 @@
+#![feature(sgx_platform)]
+//use std::os::fortanix_sgx::usercalls::raw::connect_stream;
+//use std::os::fortanix_sgx::io::FromRawFd::*;
```
jethrogb: Remove above 3 lines

examples/api-extension/runner/src/main.rs
jethrogb: Unnecessary wrapper type, `File` is already `SyncStream`

examples/api-extension/runner/src/main.rs
```diff
@@ -0,0 +1,101 @@
...
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+extern crate aesm_client;
+extern crate enclave_runner;
+extern crate sgxs_loaders;
+extern crate failure;
+#[macro_use]
+extern crate clap;
```
jethrogb: Don't use clap, trim the example code to have the minimal code for what you want to demonstrate

enclave-runner/src/loader.rs
```diff
@@ -200,6 +205,10 @@ impl<'a> EnclaveBuilder<'a> {
         self
     }
 
+    pub fn api_ext_impl(&mut self, e: Box<ApiExtension>) {
```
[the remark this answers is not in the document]
VardhanThigle: I will use `UsercallExtension`

examples/api-extension/runner/src/main.rs
VardhanThigle: Removed dependency on Clap and Failure.

enclave-runner/src/lib.rs
```diff
@@ -30,3 +30,4 @@ mod usercalls;
 pub use command::Command;
 pub use library::Library;
 pub use loader::{EnclaveBuilder, EnclavePanic};
+pub use usercalls::{UsercallExtension, SyncStream};
```
jethrogb: I think this makes more sense in its own module. Can you make `mod usercalls` `pub`?

enclave-runner/src/loader.rs
```diff
@@ -200,6 +205,10 @@ impl<'a> EnclaveBuilder<'a> {
         self
     }
 
+    pub fn user_call_ext_impl(&mut self, e: Box<UsercallExtension>) {
```
jethrogb: Call this `usercall_extension`. Change the parameter name to not be a single letter

jethrogb: Please add
```
impl<T: UsercallExtension> Into<Box<UsercallExtension>> for T { ... }
```
And change this function so it takes `T: Into<Box<UsercallExtension>>`

enclave-runner/src/usercalls/mod.rs
```diff
@@ -545,6 +552,43 @@ fn trap_attached_debugger(tcs: usize) {
     }
 }
 
+fn connect_stream_dflt(
```
jethrogb: Move this implementation into `fn connect_stream` below. This way you can handle the `OutputBuffer`s the same way for both the default and custom cases.

enclave-runner/src/usercalls/mod.rs
```diff
@@ -545,6 +552,43 @@ fn trap_attached_debugger(tcs: usize) {
...
+        }
+    Ok(Box::new(stream))
+}
+
+pub trait UsercallExtension : 'static + Send + Sync + std::fmt::Debug {
+    fn connect_stream(
+        &self,
+        _addr: &[u8],
+        _local_addr: Option<&mut String>,
+        _peer_addr: Option<&mut String>,
```
jethrogb: Don't prefix with `_` as this will show in the docs. Either use `let _ =` statements in the function body, or silence the unused lint for this function.

enclave-runner/src/usercalls/mod.rs
```diff
@@ -32,7 +32,7 @@ lazy_static! {
 }
 
 pub mod abi;
-mod interface;
+pub mod interface;
```
jethrogb: Why?

enclave-runner/src/usercalls/mod.rs
```diff
@@ -144,12 +144,11 @@ where
...
 
-trait SyncStream: 'static + Send + Sync {
-    fn read_alloc(&self, out: &mut OutputBuffer) -> IoResult<()> {
+pub trait SyncStream: 'static + Send + Sync {
+    fn read_alloc(&self) -> IoResult<Vec<u8>> {
         let mut buf = [0u8; 8192];
         let len = self.read(&mut buf)?;
-        out.set(&buf[..len]);
-        Ok(())
+        Ok(buf[..len].into())
```
jethrogb: Just start with a vec to begin with (instead of an array).

enclave-runner/src/usercalls/mod.rs
```diff
@@ -310,26 +310,32 @@ impl EnclaveState {
...
         event_queues: FnvHashMap<TcsAddress, Mutex<Sender<u8>>>,
-    ) -> Arc<Self> {
+        user_call_ext_impl: Option<Box<UsercallExtension>>) -> Arc<Self> {
         let mut fds = FnvHashMap::default();
         fds.insert(FD_STDIN, Arc::new(FileDesc::stream(Shared(io::stdin()))));
         fds.insert(FD_STDOUT, Arc::new(FileDesc::stream(Shared(io::stdout()))));
         fds.insert(FD_STDERR, Arc::new(FileDesc::stream(Shared(io::stderr()))));
         let last_fd = AtomicUsize::new(fds.keys().cloned().max().unwrap() as _);
 
+        let user_call_ext_impl = user_call_ext_impl.unwrap_or_else( || {
```
jethrogb: Don't need `unwrap_or_else` here, Box doesn't allocate for zero-sized types.
```rust
user_call_ext_impl.unwrap_or(Box::new(ConnectStreamDefault))
```
or even (after you make the impl suggested elsewehere)
```rust
user_call_ext_impl.unwrap_or(ConnectStreamDefault.into())
```

enclave-runner/src/usercalls/mod.rs
```diff
@@ -545,6 +552,43 @@ fn trap_attached_debugger(tcs: usize) {
...
+        _addr: &[u8],
+        _local_addr: Option<&mut String>,
+        _peer_addr: Option<&mut String>,
+    ) -> IoResult<Option<Box<SyncStream>>> {
+        Ok(None)
+    }
+}
+
+#[derive(Debug)]
+struct ConnectStreamDefault;
```
jethrogb: Rename to `UsercallExtensionDefault`

examples/usercall-extension/runner/Cargo.toml
```diff
@@ -0,0 +1,12 @@
+[package]
+name = "runner"
+version = "0.1.0"
+authors = ["Fortanix, Inc."]
+license = "MPL-2.0"
+
+[dependencies]
+# Project dependencies
+#TODO Update Path Depedency to cates.io depedency, once changes are published.
```
jethrogb: Do this now. `enclave-runner` version will be 0.1.1. Also update enclave-runner Cargo.toml

examples/usercall-extension/runner/src/main.rs
```diff
@@ -0,0 +1,112 @@
...
+            .stdin(Stdio::piped())
+            .spawn()
+            .expect("Failed to start cat process")))
+        }
+   }
+}
+impl SyncStream for CatService {
+
+    fn read(&self, buf: &mut [u8]) -> IoResult<usize> {
+        self.c.lock().unwrap().stdout.as_mut().expect("failed to get stdout").read(buf)
```
jethrogb: (3×) just use `unwrap` here

examples/usercall-extension/runner/src/main.rs
```diff
@@ -0,0 +1,112 @@
...
+#[derive(Debug)]
+struct ExternalService;
+impl UsercallExtension for ExternalService {
+    fn connect_stream(
+        &self,
+        addr: &[u8],
+        local_addr: Option<&mut String>,
+        peer_addr: Option<&mut String>,
+    ) -> IoResult<Option<Box<SyncStream>>> {
+        let name = String::from_utf8(addr.to_vec()).map_err(|_| IoErrorKind::ConnectionRefused)?;
```
jethrogb: Don't error out if the address is not UTF-8

examples/usercall-extension/app/Cargo.toml
```diff
@@ -0,0 +1,9 @@
+[package]
+name = "app"
+version = "0.1.0"
+authors = ["Vardhan Thigle <vardhan.thigle@fortanix.com>"]
+
+#[profile.dev]
+panic = "abort"
```
jethrogb: What are you trying to do here?

examples/usercall-extension/app/src/main.rs
```diff
@@ -0,0 +1,17 @@
...
+
+    
+    s.write_all(&message.as_bytes())?;
+    let mut b = BufReader::new(s);
+    b.read_line(&mut echo).unwrap();
+    println!("{}", echo);
+
+    println!("Done!!");
+    Ok(())
+}
```
jethrogb: Cleaned this up a bit:
```rust
use std::io::{BufRead, BufReader, Read, Write};
use std::net::TcpStream;

fn main() -> std::io::Result<()> {
    let mut stream  = BufReader::new(TcpStream::connect("cat")?);

    stream.get_mut().write_all(b"Hello, world!\n")?;
    
    let mut echo = String::new();
    b.read_line(&mut echo)?;
    println!("{}", echo);

    Ok(())
}
```

examples/usercall-extension/runner/Cargo.toml
```diff
@@ -0,0 +1,12 @@
+[package]
+name = "runner"
+version = "0.1.0"
+authors = ["Fortanix, Inc."]
+license = "MPL-2.0"
+
+[dependencies]
+# Project dependencies
```
jethrogb: “Project dependencies” header doesn't make sense for example

examples/usercall-extension/runner/src/main.rs
```diff
@@ -0,0 +1,112 @@
+/* Copyright (c) Fortanix, Inc.
```
jethrogb: Add more comments explaining what this does

examples/usercall-extension/runner/src/main.rs
```diff
@@ -0,0 +1,112 @@
...
+
+use std::io::{ErrorKind as IoErrorKind, Result as IoResult, Write, Read};
+use std::process::{Command, Child, Stdio};
+use std::sync::{Arc, Mutex};
+
+struct CatService {
+   c : Arc<Mutex<Child>>
+}
+impl CatService {
+   fn new() -> CatService {
```
jethrogb: This should return a result

enclave-runner/src/loader.rs
VardhanThigle: Thanks. Should I do something similar for Box<SyncStream> too?

examples/usercall-extension/app/Cargo.toml
VardhanThigle: This was, since I was trying to use gdb to debug the panic. Panic does not happen on latest nightly.

examples/usercall-extension/app/src/main.rs
VardhanThigle: Thanks a lot.

enclave-runner/src/usercalls/mod.rs
```diff
@@ -32,7 +32,7 @@ lazy_static! {
 }
 
 pub mod abi;
```
jethrogb: Make this `pub(crate)`

enclave-runner/src/usercalls/mod.rs
```diff
@@ -545,6 +559,28 @@ fn trap_attached_debugger(tcs: usize) {
     }
 }
 
+#[allow(unused)]
+pub trait UsercallExtension : 'static + Send + Sync + std::fmt::Debug {
+    fn connect_stream(
```
jethrogb: Add documentation

enclave-runner/src/usercalls/mod.rs
```diff
@@ -681,21 +734,52 @@ impl RunningTcs {
...
+        let o = 
+        {
+            let la = local_addr.as_ref().map(|_| &mut l);
+            let pa = peer_addr.as_ref().map(|_| &mut p);
+
+            self.enclave.usercall_ext.connect_stream(addr, la, pa)?
+        };
+
+        let stream = 
+        {
```
jseyfried: nit: no need for this block, just start `let stream = match o {`

enclave-runner/src/usercalls/mod.rs
```diff
@@ -681,21 +734,52 @@ impl RunningTcs {
...
+            match o {
+                Some(s) => {
+                    s
+                }
```
jseyfried: nit: `Some(stream) => stream,`

enclave-runner/src/usercalls/mod.rs
```diff
@@ -681,21 +734,52 @@ impl RunningTcs {
...
-        let addr = str::from_utf8(addr).map_err(|_| IoErrorKind::ConnectionRefused)?;
-        let stream = TcpStream::connect(addr)?;
-        if let Some(local_addr) = local_addr {
-            match stream.local_addr() {
-                Ok(local) => local_addr.set(local.to_string().into_bytes()),
-                Err(_) => local_addr.set(&b"error"[..]),
+        let mut l = String::new();
+        let mut p = String::new();
+        let o = 
+        {
```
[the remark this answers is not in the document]
jseyfried: Also avoid one-letter variable names like `o`.

examples/usercall-extension/runner/src/main.rs
jseyfried: In particular, return `Ok(None)` instead.

enclave-runner/src/usercalls/mod.rs
```diff
@@ -545,6 +565,37 @@ fn trap_attached_debugger(tcs: usize) {
...
+/// UsercallExtension provides a mechanism for the enclave code interface with an external service via a modified runner.
+/// If the implemenentation of the functions of UsercallExtension reutrn None, then [default functions](../../fortanix_sgx_abi/struct.Usercalls.html#networking)
+/// would be called which inteface the enclave with TCP stream.
+/// An implementation of UsercallExtension, can be registered while [building](../struct.EnclaveBuilder.html#method.usercall_ext) the enclave.
+
+#[allow(unused)]
+pub trait UsercallExtension : 'static + Send + Sync + std::fmt::Debug {
+    /// Allows the enclave to connect to an external service specified by the addr bytes.
+    /// The Runner should determine the service that the enclave is trying to connect to by parsing the addr bytes.
+    /// If `connect_stream` returns None, the default implementation of [connect_stream](../../fortanix_sgx_abi/struct.Usercalls.html#method.connect_stream) would be called.
```
jseyfried: nit: would be called -> is used

enclave-runner/src/usercalls/mod.rs
```diff
@@ -545,6 +565,37 @@ fn trap_attached_debugger(tcs: usize) {
...
+    /// Apart from connecting to the external service, the behavior of this function should be same as the default implementation.
```
jseyfried: In practice, what should a service push to `local_addr` and `peer_addr`? Why are you pushing `"None"` in the example? (This might be answered in other docs, but I think it would be useful to put here).

examples/usercall-extension/runner/src/main.rs
```diff
@@ -0,0 +1,119 @@
...
+fn parse_args() -> Result<String, ()>
+{
+    let args: Vec<String> = std::env::args().collect();
+    match args.len() {
+        2 => Ok(args[1].parse().map_err(|_| ())?),
+        _ => {
+            usage(args[0].parse().map_err(|_| ())?);
+            Err(())
+        }
+    }
```
jseyfried: Why are we `parse()`ing a string into a string?
I believe you can just do `args[1].to_owned()` instead.

examples/usercall-extension/runner/src/main.rs
```diff
@@ -0,0 +1,119 @@
...
+    let mut device = IsgxDevice::new()
+        .unwrap()
+        .einittoken_provider(AesmClient::new())
+        .build();
+
+    let mut enclave_builder = EnclaveBuilder::new(file.as_ref());
+    enclave_builder.dummy_signature(); 
+    enclave_builder.usercall_ext(ExternalService.into());
+
+
```
jseyfried: nit: avoid extra spaces here, and avoid leading space at top of block.

examples/usercall-extension/runner/src/main.rs
```diff
@@ -0,0 +1,119 @@
...
+    let enclave = enclave_builder.build(&mut device).unwrap();
+
+    enclave.run().map_err(|e| {
+        println!("Error while executing SGX enclave.\n{}", e);
+        std::process::exit(-1)
```
jseyfried: `-1` doesn't make sense as an exit code, since often only the least significant 8 bits are seen (i.e. in practice it must be between 0 and 255). Use `1` instead.

enclave-runner/src/usercalls/mod.rs
```diff
@@ -310,26 +323,32 @@ impl EnclaveState {
...
         let mut fds = FnvHashMap::default();
         fds.insert(FD_STDIN, Arc::new(FileDesc::stream(Shared(io::stdin()))));
         fds.insert(FD_STDOUT, Arc::new(FileDesc::stream(Shared(io::stdout()))));
         fds.insert(FD_STDERR, Arc::new(FileDesc::stream(Shared(io::stderr()))));
         let last_fd = AtomicUsize::new(fds.keys().cloned().max().unwrap() as _);
 
+        let usercall_ext = usercall_ext.unwrap_or_else( || {
+                                      let d = ConnectStreamDefault;
+                                      Box::new(d)
+                                  });
```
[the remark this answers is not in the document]
VardhanThigle: Thanks. Will make it simpler.

enclave-runner/src/usercalls/mod.rs
```diff
@@ -310,26 +323,32 @@ impl EnclaveState {
...
+
         Arc::new(EnclaveState {
             kind,
             event_queues,
             fds: Mutex::new(fds),
             last_fd,
-            exiting: AtomicBool::new(false)
+            exiting: AtomicBool::new(false),
+            usercall_ext : usercall_ext,
```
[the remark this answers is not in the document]
VardhanThigle: Thanks, will edit.

examples/usercall-extension/runner/src/main.rs
VardhanThigle: makes sense.

enclave-runner/src/usercalls/mod.rs
VardhanThigle: Thanks.

examples/usercall-extension/runner/src/main.rs
VardhanThigle: Will remove.

examples/usercall-extension/runner/src/main.rs
VardhanThigle: Thanks.

enclave-runner/src/usercalls/mod.rs
```diff
@@ -545,6 +565,37 @@ fn trap_attached_debugger(tcs: usize) {
     }
 }
 
+/// UsercallExtension provides a mechanism for the enclave code interface with an external service via a modified runner.
+/// If the implemenentation of the functions of UsercallExtension reutrn None, then [default functions](../../fortanix_sgx_abi/struct.Usercalls.html#networking)
+/// would be called which inteface the enclave with TCP stream.
+/// An implementation of UsercallExtension, can be registered while [building](../struct.EnclaveBuilder.html#method.usercall_ext) the enclave.
+
+#[allow(unused)]
```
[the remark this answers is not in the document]
VardhanThigle: This was to avoid "_" in docuemntation. [ref](https://github.com/fortanix/rust-sgx/pull/93#discussion_r256693940)

examples/usercall-extension/runner/src/main.rs
VardhanThigle: Thanks. Will use owned()

enclave-runner/src/usercalls/mod.rs
```diff
@@ -545,6 +565,37 @@ fn trap_attached_debugger(tcs: usize) {
...
+    ) -> IoResult<Option<Box<SyncStream>>> {
+        Ok(None)
+    }
+}
+
+impl <T: UsercallExtension> From<T> for Box<UsercallExtension> {
+    fn from(value : T) -> Box<UsercallExtension> {
+        Box::new(value)
+    }
+}
```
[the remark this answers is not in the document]
VardhanThigle: Earlier, I had kept this as Box::new(...), mostly since I was not much used to implementing From/Into on user defined structures. 
But after earlier [code review] (https://github.com/fortanix/rust-sgx/pull/93#discussion_r256693092), I added From conversion.
I am do not have a strong inclination for either, just that I feel into, might abstract the box wrapper for trait dispatch from the user who extends the usercalls. 
Please suggest which alternative to use.

enclave-runner/src/usercalls/mod.rs
```diff
@@ -144,19 +144,31 @@ where
...
+    /// Same as [flush](https://doc.rust-lang.org/nightly/std/io/trait.Write.html#tymethod.flush) , except that it takes an immutable reference to the source.
     fn flush(&self) -> IoResult<()>;
 }
 
+
+impl <T: SyncStream> From<T> for Box<SyncStream> {
+    fn from(value : T) -> Box<SyncStream> {
+        Box::new(value)
+    }
+}
```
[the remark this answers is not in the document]
VardhanThigle: Please suggest on the below [conversation](https://github.com/fortanix/rust-sgx/pull/93#discussion_r260258274)

enclave-runner/src/usercalls/mod.rs
VardhanThigle: It makes more sense for [connect for a socket](https://edp.fortanix.com/docs/api/fortanix_sgx_abi/struct.Usercalls.html#method.connect_stream).
I am not sure what would be exact use case for generic API extension.
I will add the below part :
```
The enclave may optionally request the local or peer socket addresses be returned in local_addr or peer_addr, respectively. On success, if local_addr and/or peer_addr is not NULL, userspace will allocate memory for the address and populate the respective ByteBuffer appropriately.
```
Please suggest ways to make it more informative.

enclave-runner/src/usercalls/mod.rs
```diff
@@ -545,6 +565,37 @@ fn trap_attached_debugger(tcs: usize) {
...
+    /// The Runner should determine the service that the enclave is trying to connect to by parsing the addr bytes.
+    /// If `connect_stream` returns None, the default implementation of [connect_stream](../../fortanix_sgx_abi/struct.Usercalls.html#method.connect_stream) would be called.
+    /// Apart from connecting to the external service, the behavior of this function should be same as the default implementation.
+    fn connect_stream(
+        &self,
+        addr: &[u8],
+        local_addr: Option<&mut String>,
+        peer_addr: Option<&mut String>,
+    ) -> IoResult<Option<Box<SyncStream>>> {
+        Ok(None)
```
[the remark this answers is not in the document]
VardhanThigle: I had earlier thought, when we add more functions here like bind and accept, user could implement a subset of them, hence I kept a default. 
I am open if we keep the policy that the user must implement a function for all members, which could, in some cases, just return None.
Please suggest which way to proceed.

enclave-runner/src/usercalls/mod.rs
VardhanThigle: Thanks. Will remove the unnecessary block.

enclave-runner/src/usercalls/mod.rs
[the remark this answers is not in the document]
VardhanThigle: Sure. I had added it while experimenting to overcome borrow checker issues. Will make it more readable.

enclave-runner/src/usercalls/mod.rs
VardhanThigle: Thanks.

enclave-runner/src/usercalls/mod.rs
VardhanThigle: Thanks. I had missed reverting this.

enclave-runner/src/usercalls/mod.rs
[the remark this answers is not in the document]
jseyfried: Ok, but you're public API is still accepting a `Box<UsercallExtension>` https://github.com/fortanix/rust-sgx/pull/93/files#diff-63c5a8420129862da1b63a4f31723ed8R208, no?

I personally would prefer the public API to just take a `Box<UsercallExtension>` rather than a `T` where `T: Into<Box<UsercallExtension>>`, but if Jethro already asked for the public API to be more generic that's fine.

enclave-runner/src/usercalls/mod.rs
jseyfried: Looking at Jethro's implementation and the Rust source code, the data we push to `local_addr` and `peer_addr` should be able to be parsed into a `SocketAddr`, see the following parsing code https://doc.rust-lang.org/src/std/net/parser.rs.html#262-297. So, I believe pushing the string `"None"` makes no sense.

If the custom usercall is not a connection to a socket (i.e. probably custom usercalls, since you can connect to a socket directly without a custom usercall), then you should ignore `local_addr` and `peer_addr` and not push anything, since the local address and peer address is irrelevant.

> Apart from connecting to the external service, the behavior of this function should be same as the default implementation.

enclave-runner/src/usercalls/mod.rs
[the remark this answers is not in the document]
jseyfried: Ok, if/when we add custom usercalls for `bind` and `accept`, I could see making this method optional. Until then, I think it should be required. Note that it is backward compatible to make a required method optional, but not to make an optional method required.

enclave-runner/src/usercalls/mod.rs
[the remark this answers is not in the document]
jseyfried: Are there any public APIs that take `Box<SyncStream>`? I don't think the other conversation is relevant here.

enclave-runner/src/usercalls/mod.rs
```diff
@@ -681,21 +742,49 @@ impl RunningTcs {
...
-                Err(_) => local_addr.set(&b"error"[..]),
+        let mut l = String::new();
+        let mut p = String::new();
+        let stream_ext = 
+        {
+            let la = local_addr.as_ref().map(|_| &mut l);
+            let pa = peer_addr.as_ref().map(|_| &mut p);
+
+            self.enclave.usercall_ext.connect_stream(addr, la, pa)?
+        };
```
jseyfried: Avoid very short names (i.e. 1 or 2 character names) `l`, `p`, `la`, `pa`.

If the stream returns `None`, we can/should ignore what `usercall_ext.connect_stream()` pushed to `local_addr` and `peer_addr`, since `None` indicates that the extension doesn't want to handle the usercall.

It might be a little clearer to write:
```
{
    let mut local_addr_str = local_addr.as_ref().map(|_| String::new());
    let mut peer_addr_str = peer_addr.as_ref().map(|_| String::new());
    if let Some(stream_ext) = self.enclave.usercall_ext
            .connect_stream(addr, local_addr_str.as_mut(), peer_addr_str.as_mut())? {
        if let Some(local_addr_str) = local_addr_str {
            local_addr.set(local_addr_str.into_bytes());
        }
        if let Some(peer_addr_str) = peer_addr_str {
            peer_addr.set(peer_addr_string.into_bytes());
        }
        return Ok(self.alloc_fd(FileDesc::Stream(stream)));
    }
}
// original function body before this PR would go here
```

examples/usercall-extension/runner/src/main.rs
```diff
@@ -0,0 +1,116 @@
...
+#[derive(Debug)]
+struct ExternalService;
+impl UsercallExtension for ExternalService {
+    fn connect_stream(
+        &self,
+        addr: &[u8],
+        local_addr: Option<&mut String>,
+        peer_addr: Option<&mut String>,
+    ) -> IoResult<Option<Box<SyncStream>>> {
+        let name = String::from_utf8(addr.to_vec()).unwrap_or_else(|_| "def".to_owned());
```
jseyfried: Could we change the `UsercallExtension` trait so that `connect_stream` take `addr: &str` rather than `addr: &[u8]`. This way, each implementer won't have to use `String::from_utf8` and handle the error. Also, the default implementation expects `addr` to be a string, so I think it makes sense to enforce this for extensions as well.

enclave-runner/src/usercalls/mod.rs
VardhanThigle: Thanks. I had lot of variables with short names to avoid borrow-checker issues. But I guess this is the cleanest way to achieve the same.

examples/usercall-extension/runner/src/main.rs
VardhanThigle: Thanks. Incorporated in current push.

enclave-runner/src/usercalls/mod.rs
```diff
@@ -545,6 +557,53 @@ fn trap_attached_debugger(tcs: usize) {
     }
 }
 
+/// UsercallExtension provides a mechanism for the enclave code interface with an external service via a modified runner.
+/// If the implemenentation of the functions of UsercallExtension reutrn None, then [default functions](../../fortanix_sgx_abi/struct.Usercalls.html#networking)
+/// would be called which inteface the enclave with TCP stream.
+/// An implementation of UsercallExtension, can be registered while [building](../struct.EnclaveBuilder.html#method.usercall_ext) the enclave.
+
+#[allow(unused)]
```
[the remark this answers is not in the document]
VardhanThigle: Thanks. Incorporated in recent push.

enclave-runner/src/usercalls/mod.rs
VardhanThigle: I have added this to current push. Changed a bit as needed for making addr as str and other minor compiler fixes.

enclave-runner/src/usercalls/mod.rs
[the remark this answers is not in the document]
VardhanThigle: Thanks. made the change to accept `T: Into<Box<UsercallExtension>>` as per 
https://github.com/fortanix/rust-sgx/pull/93#discussion_r256693092

enclave-runner/src/usercalls/mod.rs
```diff
@@ -545,6 +557,52 @@ fn trap_attached_debugger(tcs: usize) {
...
+
+pub trait UsercallExtension : 'static + Send + Sync + std::fmt::Debug {
+    /// Allows the enclave to connect to an external service specified by the addr bytes.
+    /// The Runner should determine the service that the enclave is trying to connect to by parsing the addr bytes.
+    /// If `connect_stream` returns None, the default implementation of [connect_stream](../../fortanix_sgx_abi/struct.Usercalls.html#method.connect_stream) is used.
+    /// The enclave may optionally request the local or peer addresses
+    /// be returned in `local_addr` or `peer_addr`, respectively. On success,
+    /// if `local_addr` and/or `peer_addr` is not NULL, userspace will allocate
+    /// memory for the address and populate the respective [`ByteBuffer`]
+    /// appropriately.
```
jseyfried: nit: I think this is too low-level an implementation detail to put in the public API of `connect_stream`. The user of this API shouldn't have to be aware that `ByteBuffer` exists at all.

enclave-runner/src/usercalls/mod.rs
```diff
@@ -545,6 +557,52 @@ fn trap_attached_debugger(tcs: usize) {
...
+    /// The enclave must handle and deallocate each buffer according to the
+    /// [`ByteBuffer`](../../fortanix_sgx_abi/struct.ByteBuffer.html) documentation.
```
jseyfried: Same here.

enclave-runner/src/usercalls/mod.rs
```diff
@@ -681,6 +741,21 @@ impl RunningTcs {
...
         peer_addr: Option<&mut OutputBuffer>,
     ) -> IoResult<Fd> {
+        let addr_str = String::from_utf8(addr.to_vec());
+        if addr_str.is_ok() {
+            let mut local_addr_str = local_addr.as_ref().map(|_| String::new());
+            let mut peer_addr_str = peer_addr.as_ref().map(|_| String::new());
+            if let Some(stream_ext) = self.enclave.usercall_ext
+                .connect_stream(&addr_str.unwrap(), local_addr_str.as_mut(), peer_addr_str.as_mut())? {
+                if let Some(local_addr) = local_addr {
+                   local_addr.set(local_addr_str.unwrap().into_bytes());
```
jseyfried: nit: indent (needs one extra space to get to 4 spaces)

examples/usercall-extension/runner/src/main.rs
```diff
@@ -0,0 +1,115 @@
...
+        // an IP address which is the default behavior.
+        match &*addr {
+           "cat" => { 
+               let stream = CatService::new()?;
+               if let Some(local_addr) = local_addr {
+                   local_addr.push_str("None");
+               }
+               if let Some(peer_addr) = peer_addr {
+                   peer_addr.push_str("None");
+               }
```
jseyfried: We shouldn't push the string `"None"` here, since it is not a valid address. Rust will error when it tries to parse it. Unless I'm missing something you already discussed with Jethro, just ignore `local_addr` and `peer_addr` unless you want to make use of them for some reason in the enclave.

We could even add another version of `connect_stream` later to this trait that doesn't take `local_addr` and `peer_addr` so that the user doesn't have to think about them at all when not relevant (which I believe will be most of the time).

enclave-runner/src/usercalls/mod.rs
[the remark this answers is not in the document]
jethrogb: The default impl is necessary when this trait is extended as @VardhanThigle explained. I want to review the implementation now instead of in a follow-up PR that otherwise doesn't touch the connect logic.

enclave-runner/src/usercalls/mod.rs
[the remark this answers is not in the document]
jethrogb: Need this again, see comment above

enclave-runner/src/usercalls/mod.rs
VardhanThigle: Thanks. I guess I can drop mention about byte buffer and freeing in context of UserCallExtension, since our ABI will take care of it.

examples/usercall-extension/runner/src/main.rs
VardhanThigle: Sure. I will remove the "None" and left a comment that we are ignoring local_addr and peer_addr since it is not relevant to the context of the example.

enclave-runner/src/usercalls/mod.rs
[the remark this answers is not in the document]
VardhanThigle: Sure. Reverted back the default implementation. Will start working on the accept and bind once this gets merged.

enclave-runner/src/usercalls/mod.rs
VardhanThigle: Thanks. Fixed in current push.

enclave-runner/src/loader.rs
```diff
@@ -200,6 +205,10 @@ impl<'a> EnclaveBuilder<'a> {
         self
     }
 
+    pub fn usercall_ext<T: Into<Box<UsercallExtension>>>(&mut self, extension: T) {
```
jethrogb: Expand abbreviation to `usercall_extension` (only here, and in the documentation link URL for UsercallExtension).

enclave-runner/src/usercalls/mod.rs
```diff
@@ -144,19 +144,26 @@ where
     }
 }
 
-trait SyncStream: 'static + Send + Sync {
-    fn read_alloc(&self, out: &mut OutputBuffer) -> IoResult<()> {
-        let mut buf = [0u8; 8192];
+/// This trait is mostly same as [Read](https://doc.rust-lang.org/nightly/std/io/trait.Read.html) + [Write](https://doc.rust-lang.org/nightly/std/io/trait.Write.html) except that it takes an immutable reference to the source.
```
jethrogb: It's hard to have this automatically link to the correct version of the std docs. I'd suggest not including the links, e.g.:
```
... same as `std::io::Read + std::io::Write` except ...
```

enclave-runner/src/usercalls/mod.rs
```diff
@@ -144,19 +144,26 @@ where
...
+pub trait SyncStream: 'static + Send + Sync {
+
```
jethrogb: Extraneous whitespace

enclave-runner/src/usercalls/mod.rs
```diff
@@ -144,19 +144,26 @@ where
...
+    fn read_alloc(&self) -> IoResult<Vec<u8>> {
+        let mut buf = vec![0; 8192];
         let len = self.read(&mut buf)?;
-        out.set(&buf[..len]);
-        Ok(())
+        buf.resize(len, 0);
+        Ok(buf)
     }
 
+    /// Same as [read](https://doc.rust-lang.org/nightly/std/io/trait.Read.html#tymethod.read), except that it takes an immutable reference to the source.
```
jethrogb: See above re. links and formatting.
```
`Read::read`
```

enclave-runner/src/usercalls/mod.rs
```diff
@@ -144,19 +144,26 @@ where
...
     fn read(&self, buf: &mut [u8]) -> IoResult<usize>;
+    /// Same as [write](https://doc.rust-lang.org/nightly/std/io/trait.Write.html#tymethod.write) , except that it takes an immutable reference to the source.
```
jethrogb: See above re. links and formatting.
```
`Write::write`
```

enclave-runner/src/usercalls/mod.rs
```diff
@@ -144,19 +144,26 @@ where
...
     fn write(&self, buf: &[u8]) -> IoResult<usize>;
+    /// Same as [flush](https://doc.rust-lang.org/nightly/std/io/trait.Write.html#tymethod.flush) , except that it takes an immutable reference to the source.
```
jethrogb: See above re. links and formatting.
```
`Write::flush`
```

enclave-runner/src/usercalls/mod.rs
```diff
@@ -144,19 +144,26 @@ where
...
     fn flush(&self) -> IoResult<()>;
 }
 
+
```
jethrogb: Extraneous whitespace

enclave-runner/src/usercalls/mod.rs
```diff
@@ -545,6 +557,52 @@ fn trap_attached_debugger(tcs: usize) {
     }
 }
 
+/// UsercallExtension provides a mechanism for the enclave code interface with an external service via a modified runner.
+/// If the implemenentation of the functions of UsercallExtension reutrn None, then [default functions](../../fortanix_sgx_abi/struct.Usercalls.html#networking)
```
jethrogb: Delete this sentence, this is already explained in the individual method docs

enclave-runner/src/usercalls/mod.rs
```diff
@@ -545,6 +557,52 @@ fn trap_attached_debugger(tcs: usize) {
     }
 }
 
+/// UsercallExtension provides a mechanism for the enclave code interface with an external service via a modified runner.
```
jethrogb:
```
`UsercallExtension`
```

Add empty line after this so it looks good in the module overview

enclave-runner/src/usercalls/mod.rs
```diff
@@ -545,6 +557,52 @@ fn trap_attached_debugger(tcs: usize) {
...
+/// If the implemenentation of the functions of UsercallExtension reutrn None, then [default functions](../../fortanix_sgx_abi/struct.Usercalls.html#networking)
+/// would be called which inteface the enclave with TCP stream.
+/// An implementation of UsercallExtension, can be registered while [building](../struct.EnclaveBuilder.html#method.usercall_ext) the enclave.
```
jethrogb:
```
`UsercallExtension`
```
(add backticks, remove comma)

enclave-runner/src/usercalls/mod.rs
```diff
@@ -545,6 +557,52 @@ fn trap_attached_debugger(tcs: usize) {
...
+
+pub trait UsercallExtension : 'static + Send + Sync + std::fmt::Debug {
+    /// Allows the enclave to connect to an external service specified by the addr bytes.
+    /// The Runner should determine the service that the enclave is trying to connect to by parsing the addr bytes.
```
jethrogb: Replace first two sentences: “Override the connection target for `connect` calls by the enclave. The runner should determine the service that the enclave is trying to connect to by looking at `addr`.”

enclave-runner/src/usercalls/mod.rs
```diff
@@ -545,6 +557,52 @@ fn trap_attached_debugger(tcs: usize) {
...
+    /// If `connect_stream` returns None, the default implementation of [connect_stream](../../fortanix_sgx_abi/struct.Usercalls.html#method.connect_stream) is used.
```
jethrogb: add backticks around function name

examples/usercall-extension/runner/src/main.rs
```diff
@@ -0,0 +1,109 @@
...
+use enclave_runner::usercalls::{SyncStream, UsercallExtension};
+use enclave_runner::EnclaveBuilder;
+use sgxs_loaders::isgx::Device as IsgxDevice;
+
+use std::io::{Read, Result as IoResult, Write};
+use std::process::{Child, Command, Stdio};
+use std::sync::{Arc, Mutex};
+
+/// This example demonstrates use of user-call extensions.
+/// User call extension allow the enclave code to "connect" to an external service via a customized enclave runner.
```
jethrogb: Style: usercall is one word, no dash

enclave-runner/src/usercalls/mod.rs
```diff
@@ -681,6 +741,21 @@ impl RunningTcs {
...
+                if let Some(local_addr) = local_addr {
+                    local_addr.set(local_addr_str.unwrap().into_bytes());
+                }
+                if let Some(peer_addr) = peer_addr {
+                    peer_addr.set(peer_addr_str.unwrap().into_bytes());
+                }
+                return Ok(self.alloc_fd(FileDesc::Stream(stream_ext)));
+            }
+        }
         let addr = str::from_utf8(addr).map_err(|_| IoErrorKind::ConnectionRefused)?;
```
jseyfried: We already converted `addr` to utf8 at the beginning of this function, so we shouldn't do it again here. You should be able to just move this line to the top of the function (replacing `let addr_str = ...`)

jethrogb:
> Looking at the docs, there's a module usercalls::abi. This should not be visible.

This has not been resolved

VardhanThigle:
Thanks a lot for the detailed review @jethrogb and @jseyfried. I have updaed the PR with review comments after incorporating review comments.

enclave-runner/src/usercalls/mod.rs
```diff
@@ -545,6 +556,50 @@ fn trap_attached_debugger(tcs: usize) {
     }
 }
 
+/// `UsercallExtension`
+/// Provides a mechanism for the enclave code interface with an external service via a modified runner.
```
jethrogb: This isn't quite what I meant with adding an empty line. But I now see this needs additional rewording. Do this:

```patch
-/// `UsercallExtension`
-/// Provides a mechanism for the enclave code interface with an external service via a modified runner.
+/// Provides a mechanism for the enclave code to interface with an external service via a modified runner.
+///
```

VardhanThigle: Thanks. Fixed in the recent push.

jethrogb: You didn't make the exact change I suggested.

VardhanThigle: Got it. I missed the empty comment instead I used empty newline. Pushed a fix.

jethrogb: You still didn't delete the line that just says
```
/// `UsercallExtension`
```

VardhanThigle: Fixed in recent push.

Issue #19, opened 2018-11-20
GDB: automatically load binary at right address


jethrogb:
Currently, when running `add-symbol-file`, you need to manually add the offset to the `.text` section. This could come from ELF instead.

jethrogb:
@VardhanThigle can you piece out the appropriate information from your enhancements to the script so they can be used independently of sgxstate auto?

VardhanThigle:
Sure.
We need 2 inputs to `add-symbol-file`  - file name and the offset at run-time. The run-time offset is calculated from offset of `.text` section in the ELF and the base address of TCS.

So should the new function be `add-sgx-symbol-file <tcs_address>`, or, 
just `add-sgx-symbol-file` assuming TCS address is present in rbx (for example case of an abort)?
I am leaning towards the former, since in case of later, `sgxstate auto` might be more preferred.

PR #106, opened 2019-03-04
Adding a new gdb command to load enclave symbols.

VardhanThigle:
sgx-add-symbol-file <ADDR_IN_ENCLAVE | auto> <ELF_PATH>
auto indicates address is to be read from rbx.

This fixes [#19](https://github.com/fortanix/rust-sgx/issues/19)

### DIFF
--- a/scripts/gdb.py
+++ b/scripts/gdb.py
@@ -65,6 +65,12 @@ def get_text_offset(file_name):
   offset = int(text_info[5], 16)
   return offset
 
+def sgx_load_sym_file(file_name, encl_addr):
+   offset=get_text_offset(file_name)
+   address=find_vma_base(encl_addr) + offset
+   gdb.execute("add-symbol-file {} {}".format(file_name, address))
+
+
 class SgxState (gdb.Command):
   """Set/restore register state from SGX memory"""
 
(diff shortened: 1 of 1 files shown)


scripts/gdb.py
```diff
@@ -174,6 +178,28 @@ def set_state(self, registers):
     for r, v in registers.items():
       gdb.execute("set ${} = {}".format(r,v))
 
+class Sgx_Add_Symbol_File (gdb.Command):
+  """
+  Given TCS address and ELF path, load enclave's symbols.
```
jethrogb: “Given any address in the enclave and ELF path...”

VardhanThigle: Thanks. Fixed in current push.

jethrogb: Still says TCS in a few more places

jethrogb: Also I think you can say "Specifying auto is equivalent to specifying `$rbx`"

VardhanThigle: Changed the description in current push. Also removed TCS from the command syntax.

scripts/gdb.py
```diff
@@ -174,6 +178,28 @@ def set_state(self, registers):
     for r, v in registers.items():
       gdb.execute("set ${} = {}".format(r,v))
 
+class Sgx_Add_Symbol_File (gdb.Command):
+  """
+  Given any address in the enclave and ELF path, load enclave's symbols.
+  sgx-add-symbol-file <ADDR_IN_ENCLAVE | auto> <ELF_PATH>
+  Specifing address as auto is equivalent to specifying it in $rbx.
```
jethrogb: "specifying it in $rbx" doesn't make sense. You can say "specifying $rbx"

jethrogb: s/Specifing/Specifying/

VardhanThigle: Corrected in recent push.

VardhanThigle: Corrected in recent push.

PR #115, opened 2019-03-14
Revert changes to report-test/enclave/report.sgxs in 886d08d441c89fa8…

jethrogb:
…de10801b6459ec2d75992688

@VardhanThigle please think about how you could've caught this in review.


jethrogb:
bors r=jseyfried

PR #117, opened 2019-03-19
Improve sgx-isa usability, add test for local attestation

jethrogb:


sgx-tests/tests/report_mac.rs
```diff
@@ -0,0 +1,21 @@
...
+    let request = Keyrequest {
+        keyname: Keyname::Report as _,
+        keyid: report.keyid.clone(),
+        ..Default::default()
+    };
+    let key = request.egetkey().unwrap();
+
+    let mut mac_out = [0u8; 16];
+    Cipher::new(CipherId::Aes, CipherMode::ECB, 128).unwrap()
+        .cmac(&key, report.mac_data(), &mut mac_out).unwrap();
```
jethrogb: this should use the verify function

Goirad:
Would it be worth it to introduce `cfg_if` to clean up that tests file a bit?

PR #121, opened 2019-04-02
Implementing ServiceApi changes for bind call.

VardhanThigle:
1. Implement  changes for bind extension.
2. Added an example for bind extension to de-encapsulate HA proxy streams.


VardhanThigle:
Rebased on top of master.

VardhanThigle:
Re-based on top of master,

enclave-runner/src/usercalls/mod.rs
```diff
@@ -579,6 +579,21 @@ pub trait UsercallExtension : 'static + Send + Sync + std::fmt::Debug {
     ) -> IoResult<Option<Box<SyncStream>>> {
         Ok(None)
     }
+
+    /// Override the target for bind calls by the enclave. The runner should determine the service that the enclave is trying to bind to by looking at addr.
+    /// If `bind_stream` returns None, the default implementation of [`bind_stream`](../../fortanix_sgx_abi/struct.Usercalls.html#method.bind_stream) is used.
+    /// The enclave may optionally request the local be returned in `local_addr`.
```
jethrogb: Typo: “may optionally request the local *address*”

enclave-runner/src/usercalls/mod.rs
```diff
@@ -579,6 +579,21 @@ pub trait UsercallExtension : 'static + Send + Sync + std::fmt::Debug {
...
+    /// On success, if `local_addr` is not None, user-space can fill in the string as appropriate.
```
jethrogb:
```
`None`
```

enclave-runner/src/usercalls/mod.rs
```diff
@@ -598,6 +613,13 @@ impl UsercallExtension for UsercallExtensionDefault{
     ) -> IoResult<Option<Box<SyncStream>>> {
         Ok(None)
     }
+
+    fn bind_stream(&self,
+                  _addr: &str,
+                  _local_addr: Option<&mut String>
+                  ) -> IoResult<Option<Box<SyncListener>>> {
+        Ok(None)
+    }
```
jethrogb: Don't need this `impl` since it's defaulted. Actually, you can also remove the `impl` for `fn connect_stream`.

examples/usercall-extension-bind/app/src/main.rs
```diff
@@ -0,0 +1,20 @@
+use std::io::{BufRead, BufReader};
+use std::net::{TcpListener};
+
+fn main() -> std::io::Result<()> {
+    for _ in 0..3 {
+        let listener = TcpListener::bind("localhost:6010")?;
+        let (stream, _) = listener.accept()?;
```
jethrogb: Showcase how HAproxy can override peer address

examples/usercall-extension-bind/runner/src/main.rs
```diff
@@ -0,0 +1,301 @@
...
+use std::net::Shutdown;
+use std::net::{TcpListener, TcpStream};
+use std::thread;
+
+/// This example demonstrates use of usercall extensions for bind call.
+/// User call extension allow the enclave code to "bind" to an external service via a customized enclave runner.
+/// Here we customize the runner to intercept calls to bind to an address and advance the stream before returning it to enclave
+/// This can be usefull to strip protocol encapsulations, say while servicing requests load balanced by HA Proxy.
+/// This example demonstrates de-encapsulation for various HA proxy configurations before handing over the stream to the enclave.
+/// To simulate HA proxy configurations, the runner sapws a thread that connects to the same address which enclave binds to and
```
jethrogb: Typo: sapws

examples/usercall-extension-bind/runner/src/main.rs
```diff
@@ -0,0 +1,301 @@
...
+                ProxyProtocol {listener, local_address}
+        })
+    }
+}
+
+impl SyncListener for ProxyProtocol {
+    fn accept(&self) -> IoResult<(Box<SyncStream>, Box<ToString>, Box<ToString>)> {
+        let (stream, peer_address) = self.listener.accept()?;
+        let local_address = stream.local_addr().map(|addr| addr.to_string())?;
+        let peer_address = peer_address.to_string();
```
jethrogb: You should already strip the HAproxy protocol here so you can return the address found in there.

VardhanThigle:
Thanks, @jethrogb  for the detailed review.
I have incorporated the changes. For showcasing that HAProxy can override the address, I am printing the peer address in runner after actual bind and the one returned to the enclave. I am not sure if that might suffice or should we use a completely different address for enclave to bind to than the one where the runner actually binds to the thread sending test data.

examples/usercall-extension-bind/app/src/main.rs
```diff
@@ -0,0 +1,22 @@
+use std::io::{BufRead, BufReader};
+use std::net::{TcpListener};
+
+//TODO Showcase override of peer address
```
jethrogb: This is done, right?

examples/usercall-extension-bind/runner/src/main.rs
```diff
@@ -0,0 +1,331 @@
...
+                    Ok(local_address) => local_address.to_string(),
+                    Err(_) => "error".to_string(),
+                };
+                ProxyProtocol {listener, local_address}
+        })
+    }
+}
+
+impl SyncListener for ProxyProtocol {
+    fn accept(&self) -> IoResult<(Box<SyncStream>, Box<ToString>, Box<ToString>)> {
```
jethrogb: Since we're basically guaranteed to have stripped off the proxy protocol by the time this function returns, I think the logic in this module can be significantly simplified: we don't need to do the fusing or separate state tracking in `ProxyState`. I think all that's needed is a `fn read_proxy_protocol_header(&mut TcpStream) -> Result<Option<(SocketAddr, SocketAddr)>>`. After the proxy protocol has been read, you can simply return the `Box<TcpStream>` from this function. Make sure to also get the local address from the proxy protocol.

VardhanThigle: Thanks @jethrogb,  I have simplified the code to remove `ProxyState`.
Reading stream's local address in app would give an error for local profile, where we were filling `local_addr` as `Some("Err".to_string())`.
As of now, I am filling in `"0.0.0.0:0"` would that be ok?, In that case should I make `Option<(SocketAddr, SocketAddr)>` as `(SocketAddr, SocketAddr)`.

examples/usercall-extension-bind/runner/src/main.rs
```diff
@@ -0,0 +1,355 @@
...
+                       let mut buf = vec![0; *len as _];
+                       let _ = rdr.read_exact(&mut buf[..])?;
+                       Ok((Some(([0, 0, 0, 0], 0).into()), Some(([0, 0, 0, 0], 0).into())))
+                      },
+        }
+    }
+}
+
+#[allow(unused)]
+struct ProxyHdrV2 {
```
VardhanThigle: Names of this struct's elements are same as:
https://www.haproxy.org/download/1.8/doc/proxy-protocol.txt

examples/usercall-extension-bind/runner/src/main.rs
jethrogb: For local, I think it's best to use the TCP stream's actual addresses.

VardhanThigle: Thanks @jethrogb 
Recent push returns TCP stream's actual addresses for local profile.

VardhanThigle:
@jethrogb ,
I have rebased this PR on top of recent master, and fixed the deprecated syntax warnings in this PR too.
Please review the updated PR.

VardhanThigle:
Rebased on master.

examples/usercall-extension-bind/runner/src/main.rs
```diff
@@ -0,0 +1,356 @@
...
+const SIZE_HEADER_SIG: usize = 12;
+static HEADER_SIG: [u8; SIZE_HEADER_SIG] = [0x0d, 0x0a, 0x0d, 0x0a, 0x00, 0x0d, 0x0a, 0x51, 0x55, 0x49, 0x54, 0x0a];
+const V2CMD_LOCAL: u8 = 0x2_0;
+const V2CMD_PROXY: u8 = 0x2_1;
+const FAMILY_UNSPEC: u8 = 0;
+const FAMILY_TCP4: u8 = 0x11;
+const FAMILY_TCP6: u8 = 0x21;
+
+
+struct ProxyIpV4Addr {
```
jethrogb: Rust convention is to use a lowercase `v` here https://doc.rust-lang.org/std/net/struct.Ipv4Addr.html (same for v6)

examples/usercall-extension-bind/runner/src/main.rs
```diff
@@ -0,0 +1,356 @@
...
+use std::thread;
+use std::mem::size_of;
+
+/// This example demonstrates use of usercall extensions for bind call.
+/// User call extension allow the enclave code to "bind" to an external service via a customized enclave runner.
+/// Here we customize the runner to intercept calls to bind to an address and advance the stream before returning it to enclave
+/// This can be useful to strip protocol encapsulations, say while servicing requests load balanced by HA Proxy.
+/// This example demonstrates de-encapsulation for various HA proxy configurations before handing over the stream to the enclave.
+/// To simulate HA proxy configurations, the runner spawns a thread that connects to the same address which enclave binds to and 
+/// writes encapsulated test data for various HA proxy configurations to the stream.
```
jethrogb: This is really neat!

examples/usercall-extension-bind/runner/src/main.rs
```diff
@@ -0,0 +1,356 @@
...
+        let peer_addr = SocketAddr::new(dst_ip_addr.into(), proxy_addr.dst_port);
+        (local_addr, peer_addr)
+    }
+}
+
+enum ProxyAddrReader {
+    V4(u16),
+    V6(u16),
+    UNSPEC(u16),
+}
```
jethrogb: I think this is better since you always have a length:
```rust
enum ProxyAddrType {
    V4,
    V6,
    Unspec,
}

struct ProxyAddrReader {
    ty: ProxyAddrType,
    len: u16,
}
```

examples/usercall-extension-bind/runner/src/main.rs
```diff
@@ -0,0 +1,356 @@
...
+                if *len as usize != size_of::<ProxyIpV6Addr>() {
+                    Err(Error::new(ErrorKind::InvalidData, "Unexpected address length received"))?;
+                }
+                let addr = ProxyIpV6Addr::from_reader(rdr)?;
+                let (local, peer) = addr.into();
+                Ok((Some(local), Some(peer)))
+            },
+            ProxyAddrReader::UNSPEC(len) => {
+                       let mut buf = vec![0; *len as _];
+                       let _ = rdr.read_exact(&mut buf[..])?;
```
jethrogb: Instead of allocating a useless buffer, you can send the data to a sink like so:

```
io::copy(rdr.take(len), io::sink())?;
```

examples/usercall-extension-bind/runner/src/main.rs
```diff
@@ -0,0 +1,356 @@
+/* Copyright (c) Fortanix, Inc.
```
jethrogb: rustfmt this file

examples/usercall-extension-bind/runner/src/main.rs
```diff
@@ -0,0 +1,356 @@
...
+                       let _ = rdr.read_exact(&mut buf[..])?;
+                       Ok((None, None))
+                      },
+        }
+    }
+}
+
+#[allow(unused)]
+struct ProxyHdrV2 {
+    sig     : [u8; 12], /* hex 0D 0A 0D 0A 00 0D 0A 51 55 49 54 0A */
```
jethrogb: Use `SIZE_HEADER_SIG` instead of a literal

examples/usercall-extension-bind/runner/src/main.rs
```diff
@@ -0,0 +1,356 @@
...
+    ver_cmd : u8,       /* protocol version and command*/
+    fam     : u8,       /* protocol family and address*/
+    len     : u16,      /* number of following bytes part of the header*/
+    addr_reader : ProxyAddrReader /* read len bytes into a tuple of SocketAddr */
+}
+
+impl ProxyHdrV2 {
+    fn from_reader(rdr: &mut impl Read) -> IoResult<Self> {
+        let mut sig : [u8; 12] = [0; 12];
```
jethrogb: Use `SIZE_HEADER_SIG` instead of literals

examples/usercall-extension-bind/runner/src/main.rs
```diff
@@ -0,0 +1,356 @@
...
+}
+
+enum ProxyAddrReader {
+    V4(u16),
+    V6(u16),
+    UNSPEC(u16),
+}
+
+impl ProxyAddrReader {
+    fn read(&self, rdr: &mut impl Read) -> IoResult<(Option<SocketAddr>, Option<SocketAddr>)> {
```
jethrogb: I think it makes more sense for this to return `IoResult<Option<(SocketAddr, SocketAddr)>>`

examples/usercall-extension-bind/runner/src/main.rs
```diff
@@ -0,0 +1,356 @@
...
+                                                                  Option<SocketAddr>)> {
+
+    let hdr = ProxyHdrV2::from_reader(stream)?;
+    hdr.addr_reader.read(stream)
+}
+
+
+struct ProxyProtocol {
+    listener      : TcpListener,
+    pub local_address : String,
```
jethrogb: Does this need to be stored? Can you just return this from `fn new`?

examples/usercall-extension-bind/runner/src/main.rs
```diff
@@ -0,0 +1,356 @@
...
+            None => peer_address_str.to_string()
+        };
+        Ok((Box::new(stream), Box::new(local_address), Box::new(peer_address)))
+    }
+}
+
+const HAPROXY_ADDRESS : &str ="localhost:6010";
+
+#[derive(Debug)]
+struct ExternalService;
```
jethrogb: Name this `HaproxyService` or so

examples/usercall-extension-bind/runner/src/main.rs
```diff
@@ -0,0 +1,356 @@
...
+// Ignoring local_addr and peer_addr, as they are not relevant in the current context.
```
jethrogb: What is this comment referring to?

examples/usercall-extension-bind/runner/src/main.rs
VardhanThigle: Thanks. Did not know about io::sink before.

examples/usercall-extension-bind/runner/src/main.rs
VardhanThigle: Incorporated in recent push.

examples/usercall-extension-bind/runner/src/main.rs
VardhanThigle: Incorporated in recent push.

examples/usercall-extension-bind/runner/src/main.rs
VardhanThigle: Incorporated in recent push.

examples/usercall-extension-bind/runner/src/main.rs
VardhanThigle: Incorporated in recent push.

examples/usercall-extension-bind/runner/src/main.rs
VardhanThigle: Thanks. Removed the comment in recent push.

examples/usercall-extension-bind/runner/src/main.rs
VardhanThigle: Incorporated in recent push.

examples/usercall-extension-bind/runner/src/main.rs
VardhanThigle: Incorporated in recent push.

examples/usercall-extension-bind/runner/src/main.rs
VardhanThigle: Thanks, Jethro.

examples/usercall-extension-bind/runner/src/main.rs
VardhanThigle: Incorporated in recent push.

enclave-runner/src/usercalls/mod.rs
```diff
@@ -165,18 +165,18 @@ pub trait SyncStream: 'static + Send + Sync {
 }
 
 
-trait SyncListener: 'static + Send + Sync {
-    fn accept(&self) -> IoResult<(FileDesc, Box<dyn ToString>, Box<dyn ToString>)>;
+pub trait SyncListener: 'static + Send + Sync {
+    fn accept(&self) -> IoResult<(Box<dyn SyncStream>, Box<dyn ToString>, Box<dyn ToString>)>;
```
jethrogb: Should this return the socket addresses? Or should this be in line with what we do in `connect_stream/bind_stream`: take an argument `local_addr: Option<&mut String>, peer_addr: Option<&mut String>,`?

VardhanThigle: I guess returning socket address might be ok, since in the rust std library we use `addr_to_sockaddr` for `peer_address` in other cases we do accept strings.

jethrogb: I don't think you got my question. `String`s are correct. I was asking about the way it's passed.

VardhanThigle: Sorry, mostly I sill did not get the question. I understand the part that passing string between runner and enclave is correct, but in case of accept passing a string that does not convert to a socket address leads to an error in the std library, some thing that does not happen say for connect_stream returning a non TCP address.

From that point of view, at the level of UserCallExtn/SyncListener, should we enforce that a SocketAddress is returned?

jethrogb: My question is about the function signature: return value versus mutable reference argument.

Regarding your point: it doesn't matter what `std` does right now, we need to follow the [ABI](https://edp.fortanix.com/docs/api/fortanix_sgx_abi/struct.Usercalls.html#networking), which says that addresses are UTF-8 strings.

VardhanThigle: Thanks @jethrogb .
I have made it consistent with connect_stream and pushed 2 new commits.
If the approach to accept an optional mutable string looks fine, I will fixup the commits as needed.

enclave-runner/src/usercalls/mod.rs
```diff
@@ -165,18 +165,23 @@ pub trait SyncStream: 'static + Send + Sync {
...
+    fn accept(&self, local_addr: Option<&mut String>, peer_addr: Option<&mut String>) -> IoResult<Box<dyn SyncStream>> {
         TcpListener::accept(self).map(|(s, peer)| {
-            let local = match s.local_addr() {
-                Ok(local) => Box::new(local) as _,
-                Err(_) => Box::new("error") as _,
-            };
-            (FileDesc::stream(s), local, Box::new(peer) as _)
+            if let Some(local_addr) = local_addr {
+                match self.local_addr() {
+                    Ok(local) => local_addr.push_str(&local.to_string()),
```
jethrogb: I think the intent is clearer when you write `*local_addr = local.to_string()`.

Do you think we should guarantee an empty string is passed in?

VardhanThigle: I have made it cleaner in the current push by using `*local_addr = local.to_string()`
I guess, since the reference to the string originates in this file itself, we need not add an extra check to ensure it is empty.

jethrogb: > since the reference to the string originates in this file itself, we need not add an extra check to ensure it is empty.

What do you mean? We call `accept` on user implementations of `SyncListener`. If we guarantee that an empty string is passed in, then we should put that in the function documentation.

VardhanThigle: Got it. My earlier understanding of 
`we should guarantee an empty string is passed in` was incorrect ( i thought this function should error out if empty string is not passed in).
I will update the documentation.

VardhanThigle: I have added the documentation changes in recent push.

enclave-runner/src/usercalls/mod.rs
```diff
@@ -165,18 +165,30 @@ pub trait SyncStream: 'static + Send + Sync {
...
 
-trait SyncListener: 'static + Send + Sync {
-    fn accept(&self) -> IoResult<(FileDesc, Box<dyn ToString>, Box<dyn ToString>)>;
+/// SyncListener lets an implementation implement a slightly modified form of `std::net::TcpListener::accept`.
+/// The enclave may optionally request the local or peer addresses
+/// be returned in `local_addr` or `peer_addr`, respectively.
+/// On success, if `local_addr` and/or `peer_addr` are not `None`, they will point to an empty `String`.
+/// On success, user-space can fill in the strings as appropriate.
+///
+/// The enclave must not make any security decisions based on the local address received.
```
jethrogb: Most of this documentation should be on the function

enclave-runner/src/usercalls/mod.rs
```diff
@@ -569,8 +581,8 @@ pub trait UsercallExtension : 'static + Send + Sync + std::fmt::Debug {
     /// If `connect_stream` returns None, the default implementation of [`connect_stream`](../../fortanix_sgx_abi/struct.Usercalls.html#method.connect_stream) is used.
     /// The enclave may optionally request the local or peer addresses
     /// be returned in `local_addr` or `peer_addr`, respectively. On success,
```
jethrogb: Extra “On success” here

jethrogb: And also in bind/accept

enclave-runner/src/usercalls/mod.rs
VardhanThigle: Fixed in current push.

enclave-runner/src/usercalls/mod.rs
VardhanThigle: Fixed in current push.

enclave-runner/src/usercalls/mod.rs
```diff
@@ -165,18 +165,30 @@ pub trait SyncStream: 'static + Send + Sync {
...
         TcpListener::accept(self).map(|(s, peer)| {
-            let local = match s.local_addr() {
-                Ok(local) => Box::new(local) as _,
-                Err(_) => Box::new("error") as _,
-            };
-            (FileDesc::stream(s), local, Box::new(peer) as _)
+            if let Some(local_addr) = local_addr {
+                match self.local_addr() {
+                    Ok(local) => {*local_addr = local.to_string()},
+                    Err(_) => {*local_addr = "error".to_string()},
```
jseyfried: Why do we set `"error"` here? It doesn't make sense to me -- it's not an IP address.

jseyfried: After further discussion, we were thinking it could make more sense to early return from this function if `self.local_addr()` returns error, or alternatively not set `local_addr` in that case.

jethrogb: The ABI [specifies](https://edp.fortanix.com/docs/api/fortanix_sgx_abi/struct.Usercalls.html#networking) that it doesn't have to be an IP address. The ABI also [specifies](https://edp.fortanix.com/docs/api/fortanix_sgx_abi/struct.Usercalls.html#method.accept_stream) that the addresses must be returned. Do you have a suggestion for a better return value? The empty string?

jseyfried: The empty string or early returning an error.
See http://man7.org/linux/man-pages/man2/getsockname.2.html

jseyfried: I don't think there is case where `self.local_addr()` returns an error but we still want to return successfully.

jseyfried: cc @mzohreva

jseyfried: Sorry to complain, this is unimportant.

PR #132, opened 2019-04-18
WIP arch change async io : synchronous impl

parthsane:
Cleaned up code + Did some Error checking
    
    Added in debug statements (using eprintln!)
    
    TODO:
    remove the previous main_entry function and related functions
    find best way to select numer of worker threads
    find a way to make current arch work in library mode
    need to fine tune some Error handling
    1) bail! wont work (because of crossbeam scope requiring a different return type??)
    2) need to  send an error handling message to the usercall thread if "The enclave returned from the main entrypoint in violation of the specification"
    3) need to do something about populationg `cmddata.other_reasons`
    4) need to be able to read 'cmddata.primary_panic_reason'

### DIFF
--- a/enclave-runner/src/tcs.rs
+++ b/enclave-runner/src/tcs.rs
@@ -6,41 +6,12 @@
 
 use std;
 use std::cell::RefCell;
+use std::os::raw::c_void;
 
 use sgx_isa::Enclu;
 use sgxs::loader::Tcs;
-use usercalls::abi::DispatchResult;
 
 pub(crate) type DebugBuffer = [u8; 1024];
-
-pub(crate) fn enter<T: Tcs, F>(
-    tcs: T,
-    mut on_usercall: F,
-    p1: u64,
-    p2: u64,
-    p3: u64,
-    p4: u64,
-    p5: u64,
-    debug_buf: Option<&RefCell<DebugBuffer>>,
-) -> (T, DispatchResult)
-where
-    F: FnMut(u64, u64, u64, u64, u64) -> DispatchResult,
-{
-    let mut result = coenter(tcs, p1, p2, p3, p4, p5, debug_buf);
-
-    while let CoResult::Yield(usercall) = result {
-        let (p1, p2, p3, p4, p5) = usercall.parameters();
-        result = match on_usercall(p1, p2, p3, p4, p5) {
-            Ok(ret) => usercall.coreturn(ret, debug_buf),
-            Err(err) => return (usercall.tcs, Err(err)),
-        }
-    }
-
-    match result {
-        CoResult::Return((tcs, v1, v2)) => (tcs, Ok((v1, v2))),
-        CoResult::Yield(_) => unreachable!(),
-    }
-}
 
 #[derive(Debug)]
 pub enum CoResult<Y, R> {
@@ -67,6 +38,10 @@ impl<T: Tcs> Usercall<T> {
         debug_buf: Option<&RefCell<DebugBuffer>>,
     ) -> ThreadResult<T> {
         coenter(self.tcs, 0, retval.0, retval.1, 0, 0, debug_buf)
+    }
+
+    pub fn tcs_address(&self) -> *mut c_void {
+        self.tcs.address()
     }
 }
 
--- a/enclave-runner/src/usercalls/mod.rs
+++ b/enclave-runner/src/usercalls/mod.rs
@@ -17,8 +17,9 @@ use std::net::{TcpListener, TcpStream};
 use std::result::Result as StdResult;
 use std::str;
 use std::sync::atomic::{AtomicBool, AtomicUsize, Ordering};
+
 use std::sync::mpsc::{self, channel, Receiver, Sender};
-use std::sync::{Arc, Condvar, Mutex};
+use std::sync::{Arc, Mutex};
 use std::thread;
 use std::time;
 
@@ -26,8 +27,8 @@ use failure;
 use fnv::FnvHashMap;
 
 use fortanix_sgx_abi::*;
-
 use sgxs::loader::Tcs as SgxsTcs;
+
 lazy_static! {
     static ref DEBUGGER_TOGGLE_SYNC: Mutex<()> = Mutex::new(());
 }
@@ -41,13 +42,16 @@ use self::libc::*;
 #[cfg(unix)]
 use self::nix::sys::signal;
 use loader::{EnclavePanic, ErasedTcs};
+use std::thread::JoinHandle;
 use tcs;
+use tcs::{CoResult, ThreadResult};
 
 const EV_ABORT: u64 = 0b0000_0000_0000_1000;
 
 struct ReadOnly<R>(R);
 struct WriteOnly<W>(W);
 
+type UsercallSendData = (ThreadResult<ErasedTcs>, RunningTcs, RefCell<[u8; 1024]>);
 macro_rules! forward {
     (fn $n:ident(&mut self $(, $p:ident : $t:ty)*) -> $ret:ty) => {
         fn $n(&mut self $(, $p: $t)*) -> $ret {
@@ -164,7 +168,6 @@ pub trait SyncStream: 'static + Send + Sync {
     fn flush(&self) -> IoResult<()>;
 }
 
-
 /// SyncListener lets an implementation implement a slightly modified form of `std::net::TcpListener::accept`.
 pub trait SyncListener: 'static + Send + Sync {
     /// The enclave may optionally request the local or peer addresses
@@ -173,16 +176,24 @@ pub trait SyncListener: 'static + Send + Sync {
     /// On success, user-space can fill in the strings as appropriate.
     ///
     /// The enclave must not make any security decisions based on the local address received.
-    fn accept(&self, local_addr: Option<&mut String>, peer_addr: Option<&mut String>) -> IoResult<Box<dyn SyncStream>>;
+    fn accept(
+        &self,
+        local_addr: Option<&mut String>,
+        peer_addr: Option<&mut String>,
+    ) -> IoResult<Box<dyn SyncStream>>;
 }
 
 impl SyncListener for TcpListener {
-    fn accept(&self, local_addr: Option<&mut String>, peer_addr: Option<&mut String>) -> IoResult<Box<dyn SyncStream>> {
+    fn accept(
+        &self,
+        local_addr: Option<&mut String>,
+        peer_addr: Option<&mut String>,
+    ) -> IoResult<Box<dyn SyncStream>> {
         TcpListener::accept(self).map(|(s, peer)| {
             if let Some(local_addr) = local_addr {
                 match self.local_addr() {
-                    Ok(local) => {*local_addr = local.to_string()},
-                    Err(_) => {*local_addr = "error".to_string()},
+                    Ok(local) => *local_addr = local.to_string(),
+                    Err(_) => *local_addr = "error".to_string(),
                 }
             }
             if let Some(peer_addr) = peer_addr {
@@ -226,7 +237,9 @@ impl FileDesc {
 
 #[derive(Debug)]
 pub(crate) enum EnclaveAbort<T> {
-    Exit { panic: T },
+    Exit {
+        panic: T,
+    },
     /// Secondary threads exiting due to an abort
     Secondary,
     IndefiniteWait,
@@ -254,11 +267,17 @@ struct StoppedTcs {
     event_queue: Receiver<u8>,
 }
 
+struct IOHandlerInput<'a> {
+    tcs: &'a mut RunningTcs,
+    enclave: &'a EnclaveState,
+    work_sender: &'a crossbeam::crossbeam_channel::Sender<Work>,
+}
+
 struct RunningTcs {
-    enclave: Arc<EnclaveState>,
     pending_event_set: u8,
     pending_events: VecDeque<u8>,
     event_queue: Receiver<u8>,
+    mode: EnclaveEntry,
 }
 
 enum EnclaveKind {
@@ -266,24 +285,16 @@ enum EnclaveKind {
     Library(Library),
 }
 
-struct CommandSync {
-    threads: Vec<StoppedTcs>,
+struct PanicReason {
     primary_panic_reason: Option<EnclaveAbort<EnclavePanic>>,
     other_reasons: Vec<EnclaveAbort<EnclavePanic>>,
-    running_secondary_threads: usize,
 }
 
 struct Command {
-    data: Mutex<CommandSync>,
-    // Any lockholder reducing data.running_secondary_threads to 0 must
-    // notify_all before releasing the lock
-    wait_secondary_threads: Condvar,
+    panic_reason: Mutex<PanicReason>,
 }
 
-struct Library {
-    threads: Mutex<Receiver<StoppedTcs>>,
-    thread_sender: Mutex<Sender<StoppedTcs>>,
-}
+struct Library {}
 
 impl EnclaveKind {
     fn as_command(&self) -> Option<&Command> {
@@ -293,7 +304,7 @@ impl EnclaveKind {
         }
     }
 
-    fn as_library(&self) -> Option<&Library> {
+    fn _as_library(&self) -> Option<&Library> {
         match self {
             EnclaveKind::Library(l) => Some(l),
             _ => None,
@@ -308,6 +319,35 @@ pub(crate) struct EnclaveState {
     last_fd: AtomicUsize,
     exiting: AtomicBool,
     usercall_ext: Box<dyn UsercallExtension>,
+    threads_queue: crossbeam::queue::SegQueue<StoppedTcs>,
+}
+
+struct Work {
+    tcs: RunningTcs,
+    entry: CoEntry,
+}
+
+enum CoEntry {
+    Initial(ErasedTcs, u64, u64, u64, u64, u64),
+    Resume(tcs::Usercall<ErasedTcs>, (u64, u64)),
+}
+
+impl Work {
+    fn do_work(self, io_send_queue: &mpsc::Sender<UsercallSendData>) {
+        let buf = RefCell::new([0u8; 1024]);
+        let usercall_send_data = match self.entry {
+            CoEntry::Initial(erased_tcs, p1, p2, p3, p4, p5) => {
+                let coresult = tcs::coenter(erased_tcs, p1, p2, p3, p4, p5, Some(&buf));
+                ((coresult, self.tcs, buf))
+            }
+            CoEntry::Resume(usercall, coresult) => {
+                let coresult = usercall.coreturn(coresult, Some(&buf));
+                ((coresult, self.tcs, buf))
+            }
+        };
+        // if there is an error do nothing, as it means that the main thread has exited
+        let _ = io_send_queue.send(usercall_send_data);
+    }
 }
 
 impl EnclaveState {
@@ -330,15 +370,23 @@ impl EnclaveState {
 
     fn new(
         kind: EnclaveKind,
-        event_queues: FnvHashMap<TcsAddress, Mutex<Sender<u8>>>,
-        usercall_ext: Option<Box<dyn UsercallExtension>>) -> Arc<Self> {
+        mut event_queues: FnvHashMap<TcsAddress, Mutex<Sender<u8>>>,
+        usercall_ext: Option<Box<dyn UsercallExtension>>,
+        threads_vector: Vec<ErasedTcs>,
+    ) -> Arc<Self> {
         let mut fds = FnvHashMap::default();
         fds.insert(FD_STDIN, Arc::new(FileDesc::stream(Shared(io::stdin()))));
         fds.insert(FD_STDOUT, Arc::new(FileDesc::stream(Shared(io::stdout()))));
         fds.insert(FD_STDERR, Arc::new(FileDesc::stream(Shared(io::stderr()))));
         let last_fd = AtomicUsize::new(fds.keys().cloned().max().unwrap() as _);
 
-        let usercall_ext = usercall_ext.unwrap_or_else( || Box::new(UsercallExtensionDefault));
+        let usercall_ext = usercall_ext.unwrap_or_else(|| Box::new(UsercallExtensionDefault));
+
+        let threads_queue = crossbeam::queue::SegQueue::new();
+
+        for thread in threads_vector {
+            threads_queue.push(Self::event_queue_add_tcs(&mut event_queues, thread));
+        }
 
         Arc::new(EnclaveState {
             kind,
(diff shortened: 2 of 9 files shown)


enclave-runner/src/usercalls/mod.rs
```diff
@@ -1199,15 +1216,15 @@ impl RunningTcs {
...
-                } else {
-                    match self.event_queue.try_recv() {
+//                let ev = if wait {
+//                    self.event_queue.recv()
+//                } else {
+                    let ev = match self.event_queue.try_recv() {
                         Ok(ev) => Ok(ev),
                         Err(mpsc::TryRecvError::Disconnected) => Err(mpsc::RecvError),
-                        Err(mpsc::TryRecvError::Empty) => break,
+                        Err(mpsc::TryRecvError::Empty) => return Ok(event_mask.into()),
```
jethrogb: Don't change the behavior for `WAIT_NO`

enclave-runner/src/usercalls/mod.rs
```diff
@@ -602,50 +904,48 @@ impl UsercallExtension for UsercallExtensionDefault{
...
 #[allow(unused_variables)]
 impl RunningTcs {
-    fn entry(
+
+    fn entry_async(
         enclave: Arc<EnclaveState>,
         tcs: StoppedTcs,
-        mode: EnclaveEntry,
-    ) -> StdResult<(StoppedTcs, (u64, u64)), EnclaveAbort<EnclavePanic>> {
+        io_send_queue: &mpsc::Sender<(ThreadResult<ErasedTcs>, RunningTcs, EnclaveEntry, RefCell<[u8;1024]>)>,
```
jethrogb: Create a type alias or `struct` for the items in this queue so you don't have to repeat the type everywhere.

enclave-runner/src/usercalls/mod.rs
```diff
@@ -343,52 +352,234 @@ impl EnclaveState {
...
+                                    tcs,
+                                    event_queue: state.event_queue,
+                                });
+                            }
+                        },
+                        CoResult::Yield(usercall) => {
+                            let result;
+                            {
+                                let mut on_usercall =
+                                    |p1, p2, p3, p4, p5| dispatch(&mut Handler(&mut state, Some(&work_sender)), p1, p2, p3, p4, p5);
```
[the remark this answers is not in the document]
parthsane: EnclaveState is shared between threads, the work sender doesn't need to be shared between threads, it's only used by the usercall thread. Also, after the main enclave thread exits, the work_sender needs to be dropped. If I keep it in the EnclaveState then I'll have to jump through hoops to make this happen (if it can happen) and I'm unsure if it is worth it.

parthsane:
@jethrogb, I have taken care of all but one of the review comments. for that one comment I have left a response. Please review this PR.

parthsane:
@jethrogb reminder to review this PR

enclave-runner/src/usercalls/mod.rs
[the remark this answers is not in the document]
jethrogb: The current setup kind of breaks the abstraction layer (`Handler`), let me think if there's a better way.

enclave-runner/src/usercalls/mod.rs
```diff
@@ -336,59 +343,292 @@ impl EnclaveState {
...
+    fn create_worker_threads(scope : &crossbeam::thread::Scope ,num_of_threads: usize,
+                             work_receiver: crossbeam::crossbeam_channel::Receiver<Work>,
+                             io_queue_send: mpsc::Sender<UsercallSendData>,
+                             enclave: Arc<EnclaveState>) {
+        for _ in 1..num_of_threads {
+            let work_receiver = work_receiver.clone();
+            let io_queue_send = io_queue_send.clone();
+            let enclave = enclave.clone();
+
+            scope.spawn(move |_| {
```
jethrogb: Why are you using a crossbeam scope for thread spawning? Using a `move` closure pretty much defeats the purpose of that.

enclave-runner/src/usercalls/mod.rs
```diff
@@ -336,59 +343,292 @@ impl EnclaveState {
...
+                    };
+                    let res = match work {
+                        Work::Stopped(tcs, mode) => {
+                            RunningTcs::entry_async(enclave.clone(), tcs, &io_queue_send, mode)
+                        },
+                        Work::Running(state, usercall, coresult, mode) => {
+                            RunningTcs::coentry_async(state, usercall, coresult, &io_queue_send, mode)
+                        },
+                    };
+                    if let Err(err) = res {
```
jethrogb: This code seems... weird. Why are you putting threads back on the library queue after they return an error? And AFAICT threads are never put back on any queue when succesful.

enclave-runner/src/usercalls/mod.rs
```diff
@@ -336,59 +343,292 @@ impl EnclaveState {
...
+                                    (CoResult::Yield(usercall), running_tcs, _, _) => StoppedTcs {tcs : usercall.tcs, event_queue: running_tcs.event_queue},
+                                    (CoResult::Return((t, _, _)), running_tcs, _, _) => StoppedTcs { tcs: t, event_queue: running_tcs.event_queue},
+                                };
+                                if thread_sender.send(tcs).is_err() {
+                                    panic!("Couldn't send tcs after tcs work is done");
+                                }
+                            },
+                            EnclaveKind::Command(cmd) => {
+                                let mut cmddata = cmd.data.lock().unwrap();
+                                cmddata.running_secondary_threads -= 1;
```
jethrogb: This code now gets run for the main thread as well. That doesn't seem right.

enclave-runner/src/usercalls/mod.rs
```diff
@@ -446,13 +674,32 @@ impl EnclaveState {
...
+        // As currently we are not supporting spawning threads let the number of threads be 2
+        // one for usercall handling the other for actually runnign
+        let num_of_threads = 2;
+
+        let result = crossbeam::scope(|scope| {
+            EnclaveState::create_worker_threads(scope, num_of_threads, work_receiver,io_queue_send, enclave.clone());
+            // ///////////////////////////////
+            // // main syscall polling loop //
+            // ///////////////////////////////
+            return EnclaveState::syscall_loop(io_queue_receive, work_sender);
```
jethrogb: I don't think it's correct that there's a separate event loop for every library thread. Is this one of the solutions we discussed?

parthsane: Yes. We discussed various things. This was one of the solutions.

enclave-runner/src/usercalls/mod.rs
```diff
@@ -333,36 +348,258 @@ impl EnclaveState {
...
+                            (_, _, EnclaveEntry::ExecutableNonMain, _) => {
+                                if let Some(cmd) = enclave.kind.as_command() {
+                                    let mut cmddata = cmd.data.lock().unwrap();
+                                    cmddata.running_secondary_threads -= 1;
+                                    if cmddata.running_secondary_threads == 0 {
+                                        cmd.wait_secondary_threads.notify_all();
+                                    }
+                                }
+                                else {
+                                    panic!("Could not extract Command from EnclaveState");
```
jethrogb: Just use `unwrap` you already checked this is a `EnclaveEntry::ExecutableNonMain`

enclave-runner/src/usercalls/mod.rs
```diff
@@ -601,55 +834,52 @@ impl UsercallExtension for UsercallExtensionDefault{
 }
 
 #[allow(unused_variables)]
-impl RunningTcs {
-    fn entry(
+impl<'a> IOHandlerInput<'a> {
+
+    fn entry_async(
         enclave: Arc<EnclaveState>,
```
jethrogb: This parameter is not used.

enclave-runner/src/usercalls/mod.rs
```diff
@@ -601,55 +834,52 @@ impl UsercallExtension for UsercallExtensionDefault{
 }
 
 #[allow(unused_variables)]
-impl RunningTcs {
-    fn entry(
+impl<'a> IOHandlerInput<'a> {
+
+    fn entry_async(
```
jethrogb: This and coentry_async don't really make sense as a `fn` on `impl IOHandlerInput`. Instead, create a `fn do(io_send_queue: &mpsc::Sender<UsercallSendData>)` on `impl Work`.

enclave-runner/src/usercalls/mod.rs
```diff
@@ -333,36 +348,258 @@ impl EnclaveState {
...
             usercall_ext : usercall_ext,
+            work_sender: Mutex::new(work_sender)
         })
     }
 
+    fn create_worker_threads(num_of_threads: usize,
+                             work_receiver: crossbeam::crossbeam_channel::Receiver<Work>,
+                             io_queue_send: mpsc::Sender<UsercallSendData>,
+                             enclave: Arc<EnclaveState>) {
+        for _ in 1..num_of_threads {
```
jethrogb: This will just result in confusing off-by-one errors. This function should just take in the number of worker threads. Then you can also get rid of the minimum `num_of_threads` in the other function.

parthsane:
@jethrogb, I've resolved the review comments

enclave-runner/src/usercalls/mod.rs
```diff
@@ -294,6 +301,44 @@ pub(crate) struct EnclaveState {
...
+
+                let coresult = {
+                    let (p1, p2, p3, p4, p5) = match mode {
+                        EnclaveEntry::Library { p1, p2, p3, p4, p5 } => (p1, p2, p3, p4, p5),
+                        _ => (0, 0, 0, 0, 0),
+                    };
+                    tcs::coenter(tcs.tcs, p1, p2, p3, p4, p5, Some(&buf))
+                };
+
+                io_send_queue.send((coresult, state, mode, buf))
```
jethrogb: Can move this outside the `match`

enclave-runner/src/usercalls/mod.rs
```diff
@@ -294,6 +301,44 @@ pub(crate) struct EnclaveState {
...
+    fn do_work(self, io_send_queue: &mpsc::Sender<UsercallSendData>
+    ) -> StdResult<(), mpsc::SendError<UsercallSendData>> {
+        let buf = RefCell::new([0u8; 1024]);
+        match self {
+            Work::Stopped(tcs, mode) => {
+                let state = RunningTcs {
+                    event_queue: tcs.event_queue,
+                    pending_event_set: 0,
+                    pending_events: Default::default(),
+                };
```
jethrogb: It doesn't really make sense to construct this here. Also, it's kind of weird to pass `mode` around all the time. Probably do something like this:

```
struct Work {
    tcs: RunningTcs,
    entry: CoEntry
}

enum CoEntry {
    Initial(u64, u64, u64, u64, u64),
    Resume(tcs::Usercall<ErasedTcs>, (u64,u64)),
}
```

Here `RunningTcs` would contain `mode`. I'm hoping you can remove the initial parameters from `EnclaveEntry::Library` as well.

enclave-runner/src/usercalls/mod.rs
```diff
@@ -371,13 +628,11 @@ impl EnclaveState {
...
         let cmd = enclave.kind.as_command().unwrap();
         let mut cmddata = cmd.data.lock().unwrap();
-
-        cmddata.threads.clear();
+        //clear the threads_queue
+        cmddata.threads_queue  = crossbeam::queue::SegQueue::new();
         enclave.abort_all_threads();
-
         while cmddata.running_secondary_threads > 0 {
             cmddata = cmd.wait_secondary_threads.wait(cmddata).unwrap();
```
jethrogb: This wait logic probably needs to be revised because of the event loop. I'm kind of expecting that waiting on other threads is supposed to happen in the event loop and that no more waiting needs to happen after the event loop exits.

enclave-runner/src/usercalls/mod.rs
```diff
@@ -240,8 +243,12 @@ struct StoppedTcs {
     event_queue: Receiver<u8>,
 }
 
+struct IOHandlerInput <'a> {
+    tcs: &'a mut RunningTcs,
+    enclave: Arc<EnclaveState>
```
jethrogb: Can this be `&'a EnclaveState`?

enclave-runner/src/usercalls/mod.rs
```diff
@@ -333,36 +380,246 @@ impl EnclaveState {
...
+                                if !enclave.exiting.load(Ordering::SeqCst) {
+                                    cmddata.threads_queue.push(StoppedTcs {
+                                        tcs,
+                                        event_queue: state.event_queue,
+                                    });
+                                }
+                            }
+                        }
+                    },
+                    CoResult::Yield(usercall) => {
```
jethrogb: Can you put `Yield` first in the `match`? That seems to be a more logical order to read things in. Also, can you maybe merge the Yield (error case) handling with the `CoResult::Return` handling?

enclave-runner/src/usercalls/mod.rs
```diff
@@ -294,6 +301,44 @@ pub(crate) struct EnclaveState {
...
+}
+
+enum Work {
+    Stopped(StoppedTcs, EnclaveEntry),
+    Running(RunningTcs, tcs::Usercall<ErasedTcs>, (u64,u64), EnclaveEntry),
+}
+
+impl Work {
+    fn do_work(self, io_send_queue: &mpsc::Sender<UsercallSendData>
+    ) -> StdResult<(), mpsc::SendError<UsercallSendData>> {
```
jethrogb: I don't think you need to return an error here, the only possible error is that the main event loop has already exited.

enclave-runner/src/usercalls/mod.rs
```diff
@@ -347,36 +394,246 @@ impl EnclaveState {
...
+                        Ok(work) => work,
+                    };
+                    let res = work.do_work(&io_queue_send);
+                    if let Err(err) = res {
+                        match err.0 {
+                            (_, _, EnclaveEntry::ExecutableNonMain, _) => {
+
+                                let cmd = enclave.kind.as_command().unwrap();
+                                let mut cmddata = cmd.data.lock().unwrap();
+                                cmddata.running_secondary_threads -= 1;
```
jethrogb: I don't think you need to do this accounting here, see my comment above about the kinds of errors you can get in `do_work`. Removing this then allows you to avoid passing `EnclaveState` to worker threads entirely.

parthsane:
@jethrogb, I've pushed a commit which resolves the above comments

enclave-runner/src/usercalls/mod.rs
```diff
@@ -308,6 +316,35 @@ pub(crate) struct EnclaveState {
...
+                let coresult = tcs::coenter(erased_tcs,p1, p2, p3, p4, p5, Some(&buf));
+                ((coresult, self.tcs, buf))
+            }
+            CoEntry::Resume(usercall, coresult) => {
+                let coresult = usercall.coreturn(coresult, Some(&buf));
+                ((coresult, self.tcs, buf))
+            }
+        };
+        // if there is an error do nothing, as it means that the main thread has exited
+        if io_send_queue.send(usercall_send_data).is_err() { /*do nothing*/ }
```
jethrogb: Write `let _ = io_send_queue.send(...);`

enclave-runner/src/usercalls/mod.rs
```diff
@@ -346,37 +385,228 @@ impl EnclaveState {
...
+    }
+
+    fn syscall_loop(io_queue_receive: mpsc::Receiver<UsercallSendData>, enclave: Arc<EnclaveState>) -> StdResult<(u64, u64), EnclaveAbort<EnclavePanic>> {
+        let main_result;
+        'outer: loop {
+            let maybe_block_recv = |block| if block {
+                io_queue_receive.recv().ok()
+            } else {
+                io_queue_receive.try_recv().ok()
+            };
```
jethrogb: This is never called with `false`

enclave-runner/src/usercalls/mod.rs
```diff
@@ -346,37 +385,228 @@ impl EnclaveState {
...
+                                    (0, 0),
+                                    "Expected enclave thread entrypoint to return zero"
+                                );
+                                erased_tcs = Some(tcs);
+                                my_result = Ok((0,0));
+                            }
+                        }
+                    }
+                };
+                if state.mode != EnclaveEntry::ExecutableNonMain && my_result.is_err() {
```
jethrogb: Combine this with the match below so it's easier to see what's handled and what's not, e.g.:
```rust
match (my_result, enclave.kind, state.mode) {
    ...
}
```

enclave-runner/src/usercalls/mod.rs
```diff
@@ -346,37 +385,228 @@ impl EnclaveState {
...
+                    Err(e @ EnclaveAbort::Exit {..}) |
+                    Err(e @ EnclaveAbort::InvalidUsercall(_)) => {
+                        let cmd = enclave.kind.as_command().unwrap();
+                        let mut cmddata = cmd.data.lock().unwrap();
+
+                        if cmddata.primary_panic_reason.is_none() {
+                            cmddata.primary_panic_reason = Some(e)
+                        } else {
+                            cmddata.other_reasons.push(e)
+                        }
```
jethrogb: Shouldn't you exit the loop when you get here?

enclave-runner/src/usercalls/mod.rs
```diff
@@ -346,37 +385,228 @@ impl EnclaveState {
...
+                    };
+                    work.do_work(&io_queue_send);
+                }
+            });
+        }
+    }
+
+    fn syscall_loop(io_queue_receive: mpsc::Receiver<UsercallSendData>, enclave: Arc<EnclaveState>) -> StdResult<(u64, u64), EnclaveAbort<EnclavePanic>> {
+        let main_result;
+        'outer: loop {
```
jethrogb: I don't think the outer loop has any purpose anymore

enclave-runner/src/usercalls/mod.rs
```diff
@@ -346,37 +385,228 @@ impl EnclaveState {
...
+            let maybe_block_recv = |block| if block {
+                io_queue_receive.recv().ok()
+            } else {
+                io_queue_receive.try_recv().ok()
+            };
+            'inner: while let Some((coresult, mut state, buf)) = maybe_block_recv(true) {
+                let mut erased_tcs = None;
+                let enclave = enclave.clone();
+                let my_result;
+                match coresult {
```
jethrogb: Write `let my_result = match coresult { ... };`, this avoids having to write `my_result = ` everywhere.

enclave-runner/src/usercalls/mod.rs
```diff
@@ -346,37 +385,228 @@ impl EnclaveState {
...
+                            let mut input = IOHandlerInput { enclave: &enclave, tcs: &mut state };
+                            let (p1, p2, p3, p4, p5) = usercall.parameters();
+                            result = dispatch(&mut Handler(&mut input), p1, p2, p3, p4, p5);
+                        }
+                        match result {
+                            Ok(ret) => {
+                                if enclave.work_sender.lock().unwrap().as_ref().unwrap()
+                                    .send(Work {tcs:state, entry: CoEntry::Resume(usercall, ret )})
+                                    .is_err() {
+                                    panic!("Work sender couldn't send data to receiver")
```
jethrogb: This is equivalent to calling `Result::expect`

enclave-runner/src/usercalls/mod.rs
```diff
@@ -346,37 +385,228 @@ impl EnclaveState {
...
+                    break 'outer;
+                }
+                match my_result {
+                    Ok(ret) => {
+                        if state.mode == EnclaveEntry::ExecutableNonMain {
+                            let cmd = enclave.kind.as_command().unwrap();
+                            let mut cmddata = cmd.data.lock().unwrap();
+                            cmddata.running_secondary_threads -= 1;
+                            if cmddata.running_secondary_threads == 0 {
+                                cmd.wait_secondary_threads.notify_all();
```
jethrogb: The only thing that could be waiting is the main thread, but this is the main thread and it's executing this. As I mentioned in my comments earlier this week, I'm pretty sure this notification system isn't going to work anymore.

The point of the notification system was to make sure that `Command::run` didn't return as long as any enclave thread was still executing. Now that we have the worker thread system, you can do this differently. I think you can just join all the worker threads before returning from main_entry. I believe worker threads should be exiting automatically after the event loop is closed (because all the work sender channels are closed).

enclave-runner/src/tcs.rs
```diff
@@ -50,7 +20,7 @@ pub enum CoResult<Y, R> {
 
 #[derive(Debug)]
 pub struct Usercall<T: Tcs> {
-    tcs: T,
+    pub tcs: T,
```
jethrogb: Undo this change. The enclave state of a TCS in the middle of the usercall is such that it must only be resumed, you can't use the TCS except for returning from the usercall. The private encapsulation by `Usercall` guarantees this. See individual notes at usage sites on how to handle this.

enclave-runner/src/usercalls/mod.rs
```diff
@@ -346,36 +381,204 @@ impl EnclaveState {
...
+                    match result {
+                        Ok(ret) => {
+                            work_sender
+                                .send(Work {tcs:state, entry: CoEntry::Resume(usercall, ret )})
+                                .expect("Work sender couldn't send data to receiver");
+                            continue;
+                        },
+                        Err(EnclaveAbort::Exit { panic: true }) => {
+                            #[cfg(unix)]
+                            trap_attached_debugger(usercall.tcs.address().0 as _);
```
jethrogb: It's ok to add a method to `Usercall` to get the TCS address.

enclave-runner/src/usercalls/mod.rs
```diff
@@ -346,36 +381,204 @@ impl EnclaveState {
...
+                            Err(EnclaveAbort::Exit {
+                                panic: EnclavePanic::from(buf.into_inner()),
+                            })
+                        },
+                        Err(EnclaveAbort::Exit { panic: false }) => {
+                            if state.mode != EnclaveEntry::ExecutableMain {
+                                erased_tcs = Some(usercall.tcs);
```
jethrogb: You can't do this, see earlier note about `Usercall.tcs` visibility

enclave-runner/src/usercalls/mod.rs
```diff
@@ -346,36 +381,204 @@ impl EnclaveState {
...
+            }));
+        }
+        return thread_handles;
+    }
+
+    fn  syscall_loop(io_queue_receive: mpsc::Receiver<UsercallSendData>, enclave: Arc<EnclaveState>,
+                     work_sender: crossbeam::crossbeam_channel::Sender<Work>
+    ) -> StdResult<(u64, u64), EnclaveAbort<EnclavePanic>> {
+        while let Some((coresult, mut state, buf)) = io_queue_receive.recv().ok() {
+            let mut erased_tcs = None;
```
jethrogb: I don't think you need to save this in a variable. You can always just put it back on the queue then and there instead of assigning it.

enclave-runner/src/usercalls/mod.rs
```diff
@@ -346,36 +381,204 @@ impl EnclaveState {
...
+            let enclave = enclave.clone();
```
jethrogb: Preemptive clone is probably unnecessary

enclave-runner/src/usercalls/mod.rs
parthsane: I'm trying to do that.
```    
impl<T: Tcs> Usercall<T> {
    ...
    pub fn tcs_address(&self) -> *mut c_void { self.tcs.address() }
}
```
However for some reason, I'm having to change
`trap_attached_debugger(usercall.tcs.address().0 as _);`
to 
`trap_attached_debugger(usercall.tcs_address() as _);`

This is confusing me.

parthsane:
@jethrogb I've resolved the above comments

enclave-runner/src/usercalls/mod.rs
```diff
@@ -267,17 +277,13 @@ enum EnclaveKind {
 }
 
 struct CommandSync {
-    threads: Vec<StoppedTcs>,
     primary_panic_reason: Option<EnclaveAbort<EnclavePanic>>,
     other_reasons: Vec<EnclaveAbort<EnclavePanic>>,
-    running_secondary_threads: usize,
+    threads_queue: crossbeam::queue::SegQueue<StoppedTcs>
```
jethrogb: You have a crossbeam queue inside of a `Mutex`, that seems unnecessary. But also, can you unify the way the available TCSes are kept in the `EnclaveState` between libraries and commands?

enclave-runner/src/usercalls/mod.rs
```diff
@@ -346,36 +381,221 @@ impl EnclaveState {
...
+            match (my_result, state.mode ) {
+                (e @ Err(_), EnclaveEntry::Library ) |
+                (e @ Err(_), EnclaveEntry::ExecutableMain ) |
+                (e @ Ok(_), EnclaveEntry::ExecutableMain)=> {
+                    return e;
+                },
+                (Ok(_), EnclaveEntry::ExecutableNonMain) => {
+                    continue;
+                },
+                (Ok(ret), EnclaveEntry::Library) => {
```
jethrogb: This arm can be merged with the initial three like so:
```rust
(e, EnclaveEntry::Library) | (e, EnclaveEntry::ExecutableMain) => return e
```

enclave-runner/src/usercalls/mod.rs
```diff
@@ -346,36 +381,221 @@ impl EnclaveState {
...
+                    let mut cmddata = cmd.data.lock().unwrap();
+
+                    if cmddata.primary_panic_reason.is_none() {
+                        cmddata.primary_panic_reason = Some(e)
+                    } else {
+                        cmddata.other_reasons.push(e)
+                    }
+                    return Err(EnclaveAbort::Secondary);
+                },
+                (Err(EnclaveAbort::Secondary), EnclaveEntry::ExecutableNonMain) => {},
```
jethrogb: Is an empty block the same as `continue`? If so, merge with `Ok` case above

enclave-runner/src/usercalls/mod.rs
```diff
@@ -346,36 +381,221 @@ impl EnclaveState {
...
+                (Err(e @ EnclaveAbort::InvalidUsercall(_)), EnclaveEntry::ExecutableNonMain) => {
+                    let cmd = enclave.kind.as_command().unwrap();
+                    let mut cmddata = cmd.data.lock().unwrap();
+
+                    if cmddata.primary_panic_reason.is_none() {
+                        cmddata.primary_panic_reason = Some(e)
+                    } else {
+                        cmddata.other_reasons.push(e)
+                    }
+                    return Err(EnclaveAbort::Secondary);
```
jethrogb: Why are we returning here, but not if the original error was already `EnclaveAbort::Secondary`?

enclave-runner/src/usercalls/mod.rs
```diff
@@ -346,36 +381,221 @@ impl EnclaveState {
...
+            thread_handles.push(thread::spawn(move || {
+                while let Ok(work) = work_receiver.recv() {
+                    work.do_work(&io_queue_send);
+                }
+            }));
+        }
+        return thread_handles;
+    }
+
+    fn  syscall_loop(io_queue_receive: mpsc::Receiver<UsercallSendData>, enclave: Arc<EnclaveState>,
```
jethrogb: Instead of taking `enclave: Arc<EnclaveState>`, I think you can make this take `&self`. Same applies to `setup_and_start_work` below

enclave-runner/src/usercalls/mod.rs
```diff
@@ -346,36 +381,221 @@ impl EnclaveState {
             fds: Mutex::new(fds),
             last_fd,
             exiting: AtomicBool::new(false),
-            usercall_ext : usercall_ext,
+            usercall_ext,
         })
     }
 
+    fn create_worker_threads(num_of_worker_threads: usize,
```
jethrogb: It doesn't really make sense this is on `impl EnclaveState`. Move this function so it's defined inside `setup_and_start_work`

enclave-runner/src/usercalls/mod.rs
```diff
@@ -346,36 +381,221 @@ impl EnclaveState {
...
+                    let mut cmddata = cmd.data.lock().unwrap();
+                    cmddata.other_reasons.push(e)
+                },
+
+            }
+        }
+        unreachable!();
+    }
+
+    fn setup_and_start_work(num_of_worker_threads: usize,
```
jethrogb: I think this should just be called `fn run`

enclave-runner/src/usercalls/mod.rs
```diff
@@ -346,36 +381,221 @@ impl EnclaveState {
...
             .collect();
 
+        let start_queue = crossbeam::queue::SegQueue::new();
+        for tcs in threads {
+            start_queue.push(tcs);
+        }
+        let StoppedTcs {
+            tcs: main_tcs,
+            event_queue,
+        } = main;
```
jethrogb: I feel like this destructuring doesn't really add much. Just write `main.event_queue` and `main.tcs` below, where needed. Same inside `library_entry` and `launch_thread`.

parthsane:
finished working on the comments. As the `as_library` is no longer used, I've renamed it to `_as_library`, but it can be removed too.

enclave-runner/src/usercalls/mod.rs
```diff
@@ -661,27 +637,21 @@ impl EnclaveState {
         p5: u64,
     ) -> StdResult<(u64, u64), failure::Error> {
         // There is no other way than `Self::library` to get an `Arc<Self>`
```
jethrogb: This comment no longer makes sense

enclave-runner/src/usercalls/mod.rs
```diff
@@ -432,25 +421,21 @@ impl EnclaveState {
                         Err(EnclaveAbort::Exit { panic: false }) => {
                             match state.mode {
                                 EnclaveEntry::ExecutableNonMain => {
-                                    let cmd = enclave.kind.as_command().unwrap();
-                                    let cmddata = cmd.data.lock().unwrap();
                                     // If the enclave is in the exit-state, threads are no
                                     // longer able to be launched
-                                    if !enclave.exiting.load(Ordering::SeqCst) {
-                                        cmddata.threads_queue.push(StoppedTcs {
+                                    if !self.exiting.load(Ordering::SeqCst) {
```
jethrogb: A library should have the same behavior when `exiting` is set

enclave-runner/src/usercalls/mod.rs
```diff
@@ -605,12 +587,13 @@ impl EnclaveState {
             Err(EnclaveAbort::Secondary) |
             Ok(_) => false,
         };
-        let cmd = enclave.kind.as_command().unwrap();
-        let mut cmddata = cmd.data.lock().unwrap();
+
         //clear the threads_queue
-        cmddata.threads_queue  = crossbeam::queue::SegQueue::new();
+        while enclave.threads_queue.pop().is_ok() {};
         enclave.abort_all_threads();
```
jethrogb: Should call abort before clearing the queue

enclave-runner/src/usercalls/mod.rs
```diff
@@ -279,16 +279,13 @@ enum EnclaveKind {
 struct CommandSync {
```
jethrogb: Rename to `struct PanicReason`

enclave-runner/src/usercalls/mod.rs
```diff
@@ -279,16 +279,13 @@ enum EnclaveKind {
...
     primary_panic_reason: Option<EnclaveAbort<EnclavePanic>>,
     other_reasons: Vec<EnclaveAbort<EnclavePanic>>,
-    threads_queue: crossbeam::queue::SegQueue<StoppedTcs>
 }
 
 struct Command {
     data: Mutex<CommandSync>,
```
jethrogb: Rename field to `panic_reason`

enclave-runner/src/usercalls/mod.rs
```diff
@@ -340,16 +374,181 @@ impl EnclaveState {
...
+                    continue;
+                },
+                (Err(e @ EnclaveAbort::Exit {..}), EnclaveEntry::ExecutableNonMain)
+                | (Err(e @ EnclaveAbort::InvalidUsercall(_)), EnclaveEntry::ExecutableNonMain)
+                => {
+                    let cmd = self.kind.as_command().unwrap();
+                    let mut cmddata = cmd.data.lock().unwrap();
+
+                    if cmddata.primary_panic_reason.is_none() {
+                        cmddata.primary_panic_reason = Some(e)
```
jethrogb: Somewhere, we lost the logic where this gets set for `EnclaveEntry::ExecutableMain`

jethrogb:
Also, run `rustfmt` again

enclave-runner/src/usercalls/mod.rs
parthsane: According to the previous architecture, this was only set for threads. 
I tried to mirror the error handling as it was before, hence, it isn't set for `EnclaveEntry::ExecutableMain`.

Is this variable not meant for understanding what the primary reason might be if a secondary thread panics?

jethrogb: Yes, you are correct. Thanks for paying attention!

parthsane:
@jethrogb I've finished with the comments

enclave-runner/src/usercalls/mod.rs
```diff
@@ -330,78 +370,246 @@ impl EnclaveState {
...
+                            Err(EnclaveAbort::Exit {
+                                panic: EnclavePanic::from(buf.into_inner()),
+                            })
+                        }
+                        Err(EnclaveAbort::Exit { panic: false }) => {
+                            match state.mode {
+                                EnclaveEntry::ExecutableNonMain | EnclaveEntry::Library => {
+                                    // If the enclave is in the exit-state, threads are no
+                                    // longer able to be launched
+                                    if !self.exiting.load(Ordering::SeqCst) {
```
jethrogb: This doesn't make sense. This is handling `EnclaveAbort::Exit` so if you get here you must be exiting. Remember what I said before about not extracting the TCS from a `Usercall`? Remove the `fn get_tcs` function.

enclave-runner/src/usercalls/mod.rs
```diff
@@ -330,78 +370,246 @@ impl EnclaveState {
...
 
+    fn syscall_loop(
+        &self,
+        io_queue_receive: mpsc::Receiver<UsercallSendData>,
+        work_sender: crossbeam::crossbeam_channel::Sender<Work>,
+    ) -> StdResult<(u64, u64), EnclaveAbort<EnclavePanic>> {
+        while let Ok((coresult, mut state, buf)) = io_queue_receive.recv() {
+            let my_result = match coresult {
+                CoResult::Yield(usercall) => {
+                    let result;
```
jethrogb: Write
```rust
let result = {
    ...
    dispatch(...)
};
```

enclave-runner/src/usercalls/mod.rs
```diff
@@ -810,56 +959,31 @@ impl RunningTcs {
...
-        let result = thread::Builder::new().spawn(move || {
-            let tcs = recv.recv().unwrap();
-            let ret = EnclaveState::thread_entry(&enclave, tcs);
-
-            let cmd = enclave.kind.as_command().unwrap();
-            let mut cmddata = cmd.data.lock().unwrap();
-            cmddata.running_secondary_threads -= 1;
-            if cmddata.running_secondary_threads == 0 {
-                cmd.wait_secondary_threads.notify_all();
+        let cmddata = command.panic_reason.lock().unwrap();
```
jethrogb: This is unused

parthsane:
bors r=[jethrogb]

PR #137, opened 2019-05-01
Fix warning that appeared in latest nightly

jethrogb:
See https://github.com/rust-lang/rust/issues/60450


jethrogb:
bors r=jseyfried

PR #138, opened 2019-05-07
added verify to Report, replacing verify_report from libenclave

Goirad:
Added an inherent method that allows for local verification of the report

### DIFF
--- a/sgx-isa/src/lib.rs
+++ b/sgx-isa/src/lib.rs
@@ -574,6 +574,11 @@ impl Report {
         Report::try_copy_from(&out.0).unwrap()
     }
 
+    /// This function verifies the report's MAC using the provided
+    /// implementation of the verifying function.
+    ///
+    /// Care should be taken that `check_mac` prevents timing attacks,
+    /// in particular that the comparison happens in constant time. 
     #[cfg(all(feature = "sgxstd", target_env = "sgx"))]
     pub fn verify<F, R>(&self, check_mac: F) -> R
     where
--- a/sgx-isa/src/lib.rs
+++ b/sgx-isa/src/lib.rs
@@ -32,7 +32,9 @@ use core::{convert::TryFrom, num::TryFromIntError};
 #[macro_use]
 mod large_array_impl;
 #[cfg(feature = "large_array_derive")]
-macro_rules! impl_default_clone_eq { ($n:ident) => {} }
+macro_rules! impl_default_clone_eq {
+    ($n:ident) => {};
+}
 
 macro_rules! enum_def {
     (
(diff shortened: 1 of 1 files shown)


sgx-isa/src/lib.rs
```diff
@@ -576,6 +576,21 @@ impl Report {
         // unwrap ok, `out` is the correct number of bytes
         Report::try_copy_from(&out.0).unwrap()
     }
+
+    #[cfg(all(feature = "sgxstd", target_env = "sgx"))]
+    pub fn verify<F, R>(&self, check_mac: F) -> R
+        where F: FnOnce(&[u8], &[u8], &mut [u8]) -> R {
+	let req = Keyrequest {
```
jethrogb: Indentation issue. Try `rustfmt`.

sgx-isa/src/lib.rs
```diff
@@ -576,6 +576,21 @@ impl Report {
...
+	    keyname: Keyname::Report as u16,
+	    keyid: self.keyid,
+	    ..Default::default()
+	};
+	let key = req.egetkey().expect("Couldn't get report key");
+        let mut out = [0 as u8; 16];
+        check_mac(&key,
+                  unsafe{::std::slice::from_raw_parts(self as *const _ as *const u8, 384)},
+                  &mut out)
```
jethrogb: This should be self.mac

sgx-isa/src/lib.rs
```diff
@@ -576,6 +576,21 @@ impl Report {
         // unwrap ok, `out` is the correct number of bytes
         Report::try_copy_from(&out.0).unwrap()
     }
+
+    #[cfg(all(feature = "sgxstd", target_env = "sgx"))]
+    pub fn verify<F, R>(&self, check_mac: F) -> R
+        where F: FnOnce(&[u8], &[u8], &mut [u8]) -> R {
```
jethrogb: Please use the function signature I suggested (with fixed-size arrays)

sgx-isa/src/lib.rs
```diff
@@ -576,6 +580,24 @@ impl Report {
         // unwrap ok, `out` is the correct number of bytes
         Report::try_copy_from(&out.0).unwrap()
     }
+
+    #[cfg(all(feature = "sgxstd", target_env = "sgx"))]
+    pub fn verify<F, R>(&self, check_mac: F) -> R
```
VardhanThigle: small suggestion:
Would it be possible to add documentation here to tell users that check_mac should prevent a timing attack?
Another possibility could be that we accept a mac call back and we equate macs of mac.

jethrogb: Let's go with docs

Goirad: Just `// check_mac should prevent timing attacks` ?

jethrogb: Well if you're documenting, might as well write the full docs for the function. Also, you should call out the comparison specifically as an area of concern.

PS. Doc comments are triple-slash

Goirad: Added comments

PR #140, opened 2019-05-16
added missing math functions from intel's libm

Goirad:


rs-libc/src/asm/x86_64/sqrt.S
```diff
@@ -0,0 +1,80 @@
+/*
```
jethrogb: Is this file necessary? That's surprising to me. What happens if you call `f64::sqrt` in Rust?

Goirad: I'll have to try, but it was up there with the other missing math functions.

rs-libc/LICENSE
```diff
@@ -29,6 +29,21 @@ Adapted from musl-libc (license below):
     src/c/strtod.c
     src/c/strtol.c
 
+Adapted from Intel's linux-sgx libm reference (license in files):
```
jethrogb: Can you actually extract the license into this file (like musl below)?

jethrogb: Remove the word "reference"

jethrogb:
r=me after squashing all commits into one

Goirad:
Commits have been squashed

PR #153, opened 2019-07-01
Add sgx pkix crate

Goirad:
Open sources our `sgx-pkix` crate under the name `ftx-sgx-pkix`.
Dependent on `pkix` PR before merging, to update this crate's dependencies.

Unfortunately this crate does not have tests, but it compiles for linux and does not compile for sgx due to `pkix`'s dependency on `rustc-serialize`


Gautamgith:
@Goirad you can use rustc-serialize2 = "0.3.25" from https://crates.io/crates/rustc-serialize2
It is no_std.

Taowyoo:
I close this since:
- this is very old
- this is outdated since `sgx-pkix` already been added to the repo under `intel-sgx/sgx_pkix`

PR #154, opened 2019-07-01
initial copy of proxy functionality

Goirad:
Initial extraction of proxy functionality into a crate called `hyper-proxy2`.


Goirad:
Does not belong in this repo

PR #155, opened 2019-07-02
Adds native-tls-ext crate

Goirad:
Open sources some `native-tls` 0.1 extensions


Goirad:
Not relevant here, will be made into its own repo

PR #157, opened 2019-07-12
fix for newest nightly

Goirad:


sgxs-tools/src/sgx_detect/tests/debug.rs
```diff
@@ -49,26 +49,26 @@ impl DebugSupport for SgxCpuConfiguration {
             if !Paint::is_enabled() {
                 swcontrol_msg = swcontrol_msg.replace('“', r#"""#).replace('”', r#"""#).into();
             }
-            const SGX_DISABLED: Result<EfiSoftwareguardstatus, Rc<Error>> = Ok(EfiSoftwareguardstatus { status: SgxEnableStatus::Disabled });
```
jethrogb: You said this is going to be fixed in tomorrow's nightly? Then don't make these changes

Goirad: Well, I said that but what I meant was there is an open PR that makes the warning allow instead of deny. Jeffery said that it is probably not a false positive, so we would have to deal with in regardless.

Goirad: [Issue](https://github.com/rust-lang/rust/issues/62614) and [PR](https://github.com/rust-lang/rust/pull/62623)

PR #159, opened 2019-07-16
remove definitions since they are now a part of compiler builtins

Goirad:


jethrogb:
`ldexpl` isn't part of compiler-builtins, is it?

Goirad:
You're right, I've added it back in

PR #162, opened 2019-07-19
Avoid panic if ftxsgx-runner is terminated by a signal

mzohreva:
https://doc.rust-lang.org/std/process/struct.ExitStatus.html#method.code


jethrogb:
I thought there was an opportunity for functional code, but `enum CommandFail` is making this too hard.

bors r+

PR #166, opened 2019-08-07
[ZIRC-203] EnclaveOS diagnostics added for pre and post installation

prawalgangwar:


sgxs-tools/src/sgx_detect/tests/mod.rs
```diff
@@ -777,6 +777,106 @@ impl Dependency<RunEnclaveProdWl> for RunEnclaveProd {
...
+#[derive(Clone, DebugSupport)]
+struct NodeAgent {
+    version: Result<String, Rc<Error>>,
+}
+
+impl Update for NodeAgent{
+    fn update(&mut self, support: &SgxSupport) {
+        self.inner = Some(NodeAgentInner {
+            version: match support.nodeagent {
+                Ok(_) => Ok(support.nodeagent.clone().unwrap().current_ver),
```
jseyfried: No need to unwrap -- do `Ok(nodeagent) => nodeagent.current_ver` here

sgxs-tools/src/sgx_detect/interpret.rs
```diff
@@ -354,3 +356,26 @@ mod serde {
...
+
+#[derive(Debug, PartialEq)]
+pub enum BuildType {
+    Generic,
+    EnclaveOSPreInstall,
+    EnclaveOSPostInstall,
+}
+
+lazy_static! {
+    pub static ref BUILD_FOR: Mutex<BuildType> = Mutex::new(BuildType::Generic);
```
jseyfried: I think passing in `BuildType` into `fn detect` as well as `print` (add it beside the `verbose` argument) would be sufficient.

sgxs-tools/src/sgx_detect/tests/mod.rs
```diff
@@ -777,6 +777,106 @@ impl Dependency<RunEnclaveProdWl> for RunEnclaveProd {
...
+#[derive(Default, DebugSupport, Update)]
+struct Eor {
+    graphene : Status,
+    nodeagent : Status,
+    permdaemon : Status,
+}
+
+impl Print for Eor {
+    fn supported(&self) -> Status {
+        if {*BUILD_FOR.lock().unwrap() == BuildType::EnclaveOSPostInstall} {
```
jseyfried: Probably moot point, but the `{ }`  are not needed here

sgxs-tools/src/sgx_detect/tests/mod.rs
```diff
@@ -819,6 +919,21 @@ impl Tests {
             self.functions.lookup::<Psw>().supported() == Status::Supported {
             println!("\nYou're all set to start running SGX programs!");
         }
+
+        // Creating two variables as mutex lock could not be taken up twice in a single expression
+        let eos_pre_install_check = {*BUILD_FOR.lock().unwrap() == BuildType::EnclaveOSPreInstall};
+        let eos_post_install_check= {*BUILD_FOR.lock().unwrap() == BuildType::EnclaveOSPostInstall};
```
jseyfried: Probably a moot point, but
 - the `{ }` are not needed on the above two lines
 - you don't need to lock `BUILD_FOR` twice -- you can instead just assign a local variable of type `BuildType` to `*BUILD_FOR.lock().unwrap()`.

sgxs-tools/src/sgx_detect/tests/mod.rs
```diff
@@ -905,6 +1020,14 @@ impl Tests {
                     "Production mode (Intel whitelisted)" => Test(RunEnclaveProdWl),
                 }),
             }),
+            "EnclavesOS system requirements" => Category(Eor, tests: {
+                @[update_supported = graphene]
+                "Graphene kernel device" => Test(GrapheneDevice),
+                @[update_supported = nodeagent]
+                "Node Agent " => Test(NodeAgent),
```
jseyfried: nit: space between `Agent` and `"` (not sure if this matters).

sgxs-tools/src/sgx_detect/tests/mod.rs
```diff
@@ -777,6 +777,106 @@ impl Dependency<RunEnclaveProdWl> for RunEnclaveProd {
...
+    service: Result<(), Rc<Error>>,
+}
+
+impl Update for PermDaemon{
+    fn update(&mut self, support: &SgxSupport) {
+        self.inner = Some(PermDaemonInner {
+            service: match support.permdaemon {
+                Ok(v) => Ok(v),
+                Err(ref e) => Err(e.clone()),
+            },
```
jseyfried: No need for match -- just do `service: support.permdaemon.clone()`

sgxs-tools/src/sgx_detect/interpret.rs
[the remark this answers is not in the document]
jseyfried: I think passing in `BuildType` into `fn detect` as well as `fn print` would be sufficient.

sgxs-tools/src/sgx_detect/tests/mod.rs
```diff
@@ -819,6 +919,21 @@ impl Tests {
...
+        let eos_post_install_check= {*BUILD_FOR.lock().unwrap() == BuildType::EnclaveOSPostInstall};
+        if  eos_pre_install_check || eos_post_install_check {
+            if self.functions.lookup::<Isa>().supported() &
+                self.functions.lookup::<Psw>().supported() &
+                self.functions.lookup::<Eor>().supported() == Status::Supported {
+                println!("The system is configured for EnclaveOS applications!");
+            } else {
+                println!("The system is not configured for EnclaveOS applications!");
+                // Using a non standard exit code for the failure case
+                *EXIT_CODE.lock().unwrap() = 10;
```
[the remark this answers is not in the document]
jethrogb: Yeah or just return a code from this function -- seems fine to add an exit code in other error cases.

sgxs-tools/src/sgx_detect/interpret.rs
```diff
@@ -354,3 +356,26 @@ mod serde {
         }
     }
 }
+
+#[derive(Debug, Clone, Copy, Serialize, Deserialize)]
+pub enum DSStatus {
```
jethrogb: I don't think these and the types below belong in this module

sgxs-tools/src/sgx_detect/main.rs
```diff
@@ -248,6 +265,52 @@ impl SgxSupport {
...
+        })();
+        let permdaemon_status = (|| -> Result<(), Error> {
+            let mut isgx = false;
+            let mut gsgx = false;
+            let mut installed = false;
+            let mut system = sysinfo::System::new();
+            system.refresh_all();
+            for proc in system.get_process_by_name("bash") {
+                if (proc.cmd().len() >= 2) && (installed == false) {
+                    installed = proc.cmd()[1].contains("sgx_perm_daemon");
```
jethrogb: Is this really the best way to check this? How about `systemctl status perm_daemon` or so?

sgxs-tools/src/sgx_detect/main.rs
```diff
@@ -248,6 +265,52 @@ impl SgxSupport {
...
+                .filter_map(|line| line.ok())
+                .filter(|line| line.find("sgx_perm_daemon").is_some())
+                .for_each(|line| {
+                if line.contains("gsgx: 10:54") {
+                    gsgx = true;
+                } else if line.contains("isgx: 10:55"){
+                    isgx = true;
+                }
+            });
+            if isgx && gsgx && installed {
```
jethrogb: You shouldn't be checking the dependencies here. There's the dependency mechanism for that.

sgxs-tools/src/sgx_detect/main.rs
prawalgangwar: will use `systemctl is-active sgx_perm_daemon`

sgxs-tools/src/sgx_detect/interpret.rs
prawalgangwar: @jethrogb Aren't these similar to `KmodStatus`, `SgxEnableStatus` and `AesmStatus` which have also been defined in the same module ?

jethrogb: Yeah you're right. Then maybe only `BuildType` shouldn't be here.

sgxs-tools/src/sgx_detect/main.rs
prawalgangwar: this is different from the dependency check in the `tests/mod.rs` where there is a separate test for each dependency. In this case the 3 conditions for `perm_daemon` are being checked. It makes little sense to write 3 separate tests just to check this.

sgxs-tools/src/sgx_detect/interpret.rs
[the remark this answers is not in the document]
prawalgangwar: @jseyfried  passing `BuildType` in `fn print` would require me to pass it to `supported()` function and to `try_supported()` as well in the `trait Print` (`scaffold.rs`).  
There is a call to `try_supported` in `fn update`( `tests/mod.rs`) which goes on to be passed to a macro `define_dependencies`. Adding a new parameter to `fn update` is breaking something in the macro which I don't understand. Could you help me with that?

sgxs-tools/src/sgx_detect/tests/scaffold.rs
```diff
@@ -24,20 +24,31 @@ pub trait Name {
...
+    Generic,
+    EnclaveOSPreInstall,
+    EnclaveOSPostInstall,
+}
+
 pub trait Print: Name {
-    fn try_supported(&self) -> Option<Status> {
-        Some(self.supported())
+    fn try_supported(&self, build_type: &BuildType) -> Option<Status> {
+            Some(self.supported_for_config(build_type))
```
jseyfried: nit: fix indent

sgxs-tools/src/sgx_detect/main.rs
jethrogb: It's not different. perm_daemon is just another component of the whole system. You already have a `PermDaemon` test in the tree, so it's easy to add isgx and gsgx as a dependency there.

sgxs-tools/src/sgx_detect/tests/mod.rs
```diff
@@ -912,6 +1027,14 @@ impl Tests {
                     "Production mode (Intel whitelisted)" => Test(RunEnclaveProdWl),
                 }),
             }),
+            "EnclavesOS system requirements" => Category(Eor, tests: {
```
jethrogb: Typo: EnclavesOS

jethrogb: Please ensure this category is hidden when not running in EOS mode. One way to do this would be to add a `Dependency<...> for Eor` with `CONTROL_VISIBILITY = true`.

sgxs-tools/src/sgx_detect/tests/mod.rs
```diff
@@ -954,7 +1077,7 @@ fn update<T: DetectItem, U: Dependency<T>>(
     dependent.update_dependency(dependency, support);
 
     let hiddenval = if U::CONTROL_VISIBILITY {
-        dependency.try_supported() == Some(Status::Fatal)
+        dependency.try_supported(&BuildType::Generic) == Some(Status::Fatal)
```
jethrogb: It seems odd to pass a constant here

sgxs-tools/src/sgx_detect/tests/scaffold.rs
```diff
@@ -24,20 +24,31 @@ pub trait Name {
...
+    fn try_supported(&self, build_type: &BuildType) -> Option<Status> {
+        Some(self.supported_for_config(build_type))
     }
 
     fn supported(&self) -> Status {
-        panic!("Unable to answer supported query for {}", self.name())
+        panic!("Unable to answer supported query for {}\n", self.name())
+    }
+
+    fn supported_for_config(&self, _build_type: &BuildType) -> Status {
```
jethrogb: I think it makes more sense to merge this into `fn supported`, like you did for `try_supported`. I don't think ignoring `build_type` is a big deal for implementors of this trait.

sgxs-tools/src/sgx_detect/main.rs
jethrogb: I now see this relates to the checking of the logs, not the loading of the drivers. Please ignore my previous comment.

sgxs-tools/src/sgx_detect/main.rs
```diff
@@ -308,8 +358,17 @@ fn main() {
         (@arg EXPORT:   --export                                                "Export detected support information as YAML")
         (@arg PLAIN:    --plaintext                                             "Disable color and UTF-8 output")
         (@arg VERBOSE:  --verbose -v                                            "Print extra information when encountering issues")
+        (@arg ENCLAVEOS_PRE:  --enclaveos_pre_install                           "Run extra diagnostics tests for enclaveos pre installation")
+        (@arg ENCLAVEOS_POST: --enclaveos_post_install                          "Run extra diagnostics tests for enclaveos post installation")
```
jethrogb: I think dashes are more commonly used for command-line arguments. I think you can specify them like so in this macro: `--("enclaveos-post-install")`

sgxs-tools/src/sgx_detect/tests/scaffold.rs
prawalgangwar: @jethrogb  There are several calls to `fn suported` in the code. A lot of the calls are in functions where `BuildType` is not a parameter. It would become a very big change and maybe not even possible in some cases to pass `BuildType` recursively to all those parent functions. Is there a better way to do so ?

jethrogb: See https://github.com/fortanix/rust-sgx/pull/166#discussion_r353163256

sgxs-tools/src/sgx_detect/tests/mod.rs
```diff
@@ -784,8 +785,110 @@ impl Dependency<RunEnclaveProdWl> for RunEnclaveProd {
     const CONTROL_VISIBILITY: bool = true;
 }
 
+// EnclaveOS requirements
+#[derive(Default, DebugSupport, Update)]
+struct Eor {
+    graphene: Status,
+    node_agent: Status,
+    perm_daemon: Status,
```
aandyl: Only the gsgx device is truly an EnclaveOS requirement. The others are Data Shield requirements, or more precisely:

node_agent is a requirement to run Enclave Manager-enabled applications

perm_daemon is a "run containers that use SGX under a container runtime that limits device access with cgroups" requirement (meaning, not required for plain docker, but usually required for kubernetes)

This distinction may not matter for the time being, but noting it. @banne01 please chime in if you have thoughts.

sgxs-tools/src/sgx_detect/main.rs
```diff
@@ -263,6 +269,47 @@ impl SgxSupport {
...
+            let mut isgx = false;
+            let mut gsgx = false;
+            let daemon_status = Command::new("systemctl")
+                .arg("is-active")
+                .arg("sgx_perm_daemon")
+                .output()?
+                .stdout;
+            let daemon_active = str::from_utf8(&daemon_status)?
+                .eq("active\n");
+            let stdout = Command::new("journalctl")
```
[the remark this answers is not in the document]
aandyl: Maybe we could change perm_daemon to write a file `/run/sgx_perm_daemon.devices` listing the devices it is managing, and have sgx-detect check that the systemd service is running and that the device list file contains both devices?

sgxs-tools/src/sgx_detect/main.rs
prawalgangwar: @jethrogb  I don't really understand this. Do you suggest that I should replace 

```
(@arg ENCLAVEOS_PRE:  --enclaveos_pre_install
``` 

with 

``` 
(@arg ENCLAVEOS_POST: --("enclaveos-pre-install")
```

?

jethrogb: yes

sgxs-tools/src/sgx_detect/main.rs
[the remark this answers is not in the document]
prawalgangwar: I think it is better to do it later as it would involve some work on the perm_daemon side as well. I will create a JIRA ticket to track this.

sgxs-tools/src/sgx_detect/tests/mod.rs
```diff
@@ -784,8 +785,160 @@ impl Dependency<RunEnclaveProdWl> for RunEnclaveProd {
...
+            println!( "{:width$}{}{}", "", self.supported().paint(), self.name(), width = level * 2);
+        } else {
+            println!("{:width$}{}{} {}", "", self.supported().paint(), self.name(), "(Okay if container runtime is CRI-O (openshift))", width = level * 2);
+        }
+    }
+}
+
 impl Tests {
-    fn print_recurse(&self, test: TypeIdIdx, level: usize, path: &mut Vec<TypeIdIdx>, debug: &mut Vec<Vec<TypeIdIdx>>) {
+    fn print_recurse(&self, test: TypeIdIdx, level: usize, path: &mut Vec<TypeIdIdx>, debug: &mut Vec<Vec<TypeIdIdx>>, build_type: &BuildType) {
```
jethrogb: You don't need to pass BuildType here

sgxs-tools/src/sgx_detect/main.rs
```diff
@@ -263,6 +270,49 @@ impl SgxSupport {
...
+                .arg("-u")
+                .arg("sgx_perm_daemon")
+                .stdout(Stdio::piped())
+                .spawn()?
```
jethrogb: I think you can use `.output()` instead of `.stdout(Stdio::piped()).spawn()`.

sgxs-tools/src/sgx_detect/tests/mod.rs
jethrogb: I want to resolve this question before merging

sgxs-tools/src/sgx_detect/tests/mod.rs
```diff
@@ -784,8 +785,160 @@ impl Dependency<RunEnclaveProdWl> for RunEnclaveProd {
...
+                Err(_) => Status::Fatal,
+            };
+        self.node_agent = match support.node_agent {
+                Ok(_) => Status::Supported,
+                Err(_) => Status::Fatal,
+            };
+        self.perm_daemon = match support.perm_daemon {
+                Ok(_) => Status::Supported,
+                Err(_) => Status::Fatal,
+            };
```
jethrogb: I think these are all unnecessary since you have these flagged with `update_supported` below

sgxs-tools/src/sgx_detect/tests/mod.rs
```diff
@@ -784,8 +785,160 @@ impl Dependency<RunEnclaveProdWl> for RunEnclaveProd {
...
+        if self.build_type == BuildType::EnclaveOSPreInstall {
+            self.graphene & self.perm_daemon
+        } else {
+            self.graphene & self.perm_daemon & self.node_agent
+        }
+    }
+}
+
+#[derive(Clone, Debug, Default, DebugSupport)]
+struct BuildConfig {
```
jethrogb: The config or type doesn't really relate to the build anymore. Maybe a better term is `EnvConfig` or `EnvType`

sgxs-tools/src/sgx_detect/tests/mod.rs
prawalgangwar: @aandyl How should we label this? Should we use DataShield instead of EnclaveOS here?

aandyl: Given that perm daemon should be going away soon, let's overload perm daemon on EnclaveOS for now. I would propose the following changes:
* remove the EnclaveOS pre-install mode. just have one EnclaveOS mode that includes gsgx and perm daemon checks but not node agent checks
* add Enclave Manager mode, that checks the node agent
* maybe add a Data Shield command line argument that activates both EOS and EM modes.

sgxs-tools/src/sgx_detect/tests/mod.rs
```diff
@@ -784,6 +785,151 @@ impl Dependency<RunEnclaveProdWl> for RunEnclaveProd {
...
+            return Status::Fatal
+        }
+        Status::Supported
+    }
+}
+
+#[dependency]
+impl Dependency<EnvType> for DataShield {
+    const CONTROL_VISIBILITY: bool = true;
+    fn update_dependency(&mut self, _dependency: &EnvType, support: &SgxSupport) {
```
jethrogb: You don't need to define this, since you don't do anything in this function anyway.

sgxs-tools/src/sgx_detect/tests/mod.rs
```diff
@@ -784,6 +785,151 @@ impl Dependency<RunEnclaveProdWl> for RunEnclaveProd {
...
+}
+
+impl Print for EnclaveOS {
+    fn supported(&self) -> Status {
+        self.graphene & self.perm_daemon
+    }
+}
+
+#[derive(Clone, Debug, Default, DebugSupport)]
+struct EnvType {
```
jethrogb: Name of this type should be `EnvTypeNotGeneric`

sgxs-tools/src/sgx_detect/tests/mod.rs
```diff
@@ -784,6 +785,151 @@ impl Dependency<RunEnclaveProdWl> for RunEnclaveProd {
...
+
+impl Update for EnvType {
+    fn update(&mut self, support: &SgxSupport) {
+        self.build_type = support.build_type;
+    }
+}
+
+impl Print for EnvType {
+    fn supported(&self) -> Status {
+        if self.build_type == BuildType::Generic {
```
jethrogb: `(self.build_type != BuildType::Generic).as_req()`

sgxs-tools/src/sgx_detect/tests/mod.rs
```diff
@@ -784,6 +785,151 @@ impl Dependency<RunEnclaveProdWl> for RunEnclaveProd {
     const CONTROL_VISIBILITY: bool = true;
 }
 
+// Data Shield requirements
+#[derive(Debug, Default, DebugSupport, Update)]
+struct DataShield {
+    enclave_os: Status,
+    enclave_manager: Status,
+    build_type: BuildType,
```
jethrogb: This field appears unused

sgxs-tools/src/sgx_detect/tests/scaffold.rs
```diff
@@ -24,6 +24,18 @@ pub trait Name {
     fn name(&self) -> &'static str;
 }
 
+#[derive(Copy, Clone, Deserialize, Serialize, Debug, PartialEq)]
+pub enum BuildType {
```
jethrogb: As previously mentioned “build” doesn't convey the right meaning. Maybe a better term is `EnvConfig` or `EnvType`

jethrogb:
r=me after CI is passing

prawalgangwar:
bors r=jethrogb

PR #169, opened 2019-09-02
Updating protobuf version to fix CVE-2019-15544

VardhanThigle:
Updating protobuf crate version in order to fix [CVE-2019-15544](https://nvd.nist.gov/vuln/detail/CVE-2019-15544)
This also needs a corresponding update in protoc compiler.

### DIFF
--- a/aesm-client/build.rs
+++ b/aesm-client/build.rs
@@ -18,6 +18,7 @@ fn main() {
         out_dir: out_dir.to_str().expect("OUT_DIR must be valid UTF-8"),
         input: &["src/aesm_proto.proto"],
         includes: &[],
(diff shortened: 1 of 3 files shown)


aesm-client/build.rs
```diff
@@ -18,6 +19,9 @@ fn main() {
         out_dir: out_dir.to_str().expect("OUT_DIR must be valid UTF-8"),
         input: &["src/aesm_proto.proto"],
         includes: &[],
+        customize: Customize {
```
jethrogb: just write `customize: Default::default()`

VardhanThigle: Thanks. Fixed in current push.

PR #171, opened 2019-09-06
Async enclave-runner

jethrogb:


enclave-runner/src/usercalls/interface.rs
```diff
@@ -12,150 +12,278 @@ use std::slice;
...
     fn is_exiting(&self) -> bool {
         self.0.is_exiting()
     }
 
-    fn read(&mut self, fd: Fd, buf: *mut u8, len: usize) -> UsercallResult<(Result, usize)> {
-        unsafe {
-            Ok(from_raw_parts_mut_nonnull(buf, len)
-                .and_then(|buf| self.0.read(fd, buf))
-                .to_sgx_result())
+    fn read(
```
jethrogb: It's unfortunate how much more verbose these implementations have become, mainly because `?` can no longer be used. I tried writing some macros to make it nicer, but they all failed. Maybe @jseyfried has some ideas.

examples/usercall-extension/test.sh
```diff
@@ -1,17 +1,24 @@
 set -e
 
+APPNAME=app
```
jethrogb: Are the changes to this file necessary?

Instead of the changes to this file, I do think this example needs to be updated to support `AsyncStream/AsyncListener`.

enclave-runner/src/usercalls/mod.rs
```diff
@@ -76,148 +83,307 @@ impl<T> Write for ReadOnly<T> {
...
+        return Err(IoErrorKind::BrokenPipe.into());
+    }
 }
 
-trait SharedStream<'a> {
-    type Inner: Read + Write + 'a;
+impl<T> AsyncRead for ReadOnly<T> where T: std::io::Read {}
 
-    fn lock(&'a self) -> Self::Inner;
+impl<T> Read for WriteOnly<T> {
```
jethrogb: Move this and `impl<W: Write> Write for WriteOnly<W>` so it doesn't show up as changes in the diff.

enclave-runner/src/usercalls/mod.rs
```diff
@@ -76,148 +83,307 @@ impl<T> Write for ReadOnly<T> {
...
+}
 
-    fn lock(&'a self) -> Self::Inner {
-        WriteOnly(io::Stdout::lock(self))
+fn notify_other_tasks(queue: &mut VecDeque<tokio::prelude::task::Task>) {
+    // NOTE: if you replace `is_current()` with `will_notify_current()`,
+    // you should also change the `if` to `while` in the next line because
+    // will_notify_current() is not required to return true.
+    queue.retain(|task| !task.will_notify_current());
+    while let Some(task) = queue.pop_front() {
```
jethrogb: This is inefficient: first you delete some elements at random indices from the queue, requiring lots of memory shuffling. Then, you clear the queue. Why don't you just do this?
```rust
for task in queue.drain(..) {
    if !task.will_notify_current() {
        task.notify();
    }
}
```
Also I think the comment at the top of the function is obsolete.

enclave-runner/src/usercalls/mod.rs
```diff
@@ -76,148 +83,307 @@ impl<T> Write for ReadOnly<T> {
...
+            Async::NotReady => Ok(Async::NotReady),
+            Async::Ready(mut adapter) => adapter.poll_flush(),
+        })
+        .compat()
+        .await
+    }
 }
 
+struct AsyncListenerContainer {
+    inner: Mutex<Box<dyn AsyncListener>>,
```
jethrogb: Is there a reason this is a `futures::Mutex` instead of a `tokio::Lock`? I'd prefer to only use one of them throughout the file.

enclave-runner/src/usercalls/mod.rs
```diff
@@ -76,148 +83,307 @@ impl<T> Write for ReadOnly<T> {
...
+            Async::Ready(mut adapter) => adapter.poll_read(buf),
+        })
+        .compat()
+        .await
     }
 
-    fn flush(&self) -> IoResult<()> {
-        Write::flush(&mut { self })
+    async fn async_read_alloc(&self) -> IoResult<Vec<u8>> {
+        let mut vec: Vec<u8> = vec![0; 8192];
```
jethrogb: Don't default to a vec at this level. Implementors of `SyncStream` used to be able to implement `read_alloc`. We should make sure implementors of `AsyncStream` can do the same.

enclave-runner/src/usercalls/mod.rs
```diff
@@ -330,124 +522,327 @@ impl EnclaveState {
...
+    ) -> Arc<Self> {
         let mut fds = FnvHashMap::default();
-        fds.insert(FD_STDIN, Arc::new(FileDesc::stream(Shared(io::stdin()))));
-        fds.insert(FD_STDOUT, Arc::new(FileDesc::stream(Shared(io::stdout()))));
-        fds.insert(FD_STDERR, Arc::new(FileDesc::stream(Shared(io::stderr()))));
+
+        fds.insert(
+            FD_STDIN,
+            Arc::new(AsyncFileDesc::stream(ReadOnly(
+                //tokio_io::io::AllowStdIo::new(io::stdin()),
```
jethrogb: Delete stray comment

enclave-runner/src/usercalls/mod.rs
```diff
@@ -330,124 +522,327 @@ impl EnclaveState {
...
+                        cmddata.other_reasons.push(e);
+                        continue;
+                    }
+                };
+                return res;
+            }
+            // !!! confirm if this is correct;
+            unreachable!();
+        };
+        let tx_clone = tx_return_channel.clone();
```
jethrogb: Unnecessary clone, I think.

enclave-runner/src/usercalls/mod.rs
```diff
@@ -330,124 +522,327 @@ impl EnclaveState {
...
+                    (Err(e), EnclaveEntry::ExecutableNonMain) => {
+                        let cmd = enclave_clone.kind.as_command().unwrap();
+                        let mut cmddata = cmd.panic_reason.lock().await;
+                        cmddata.other_reasons.push(e);
+                        continue;
+                    }
+                };
+                return res;
+            }
+            // !!! confirm if this is correct;
```
jethrogb: Probably ok for now

enclave-runner/src/usercalls/mod.rs
```diff
@@ -330,124 +522,327 @@ impl EnclaveState {
...
+                };
+            }
+            unreachable!();
+        };
+
+        let select_fut =
+            futures::future::select(return_future.boxed(), io_future.boxed()).map(|either| {
+                match either {
+                    Either::Left((x, _)) => x,
+                    _ => unreachable!(),
```
jethrogb: This seems wrong... Why wouldn't `io_future` return here?

enclave-runner/Cargo.toml
```diff
@@ -29,3 +30,8 @@ lazy_static = "1.2.0"    # MIT/Apache-2.0
 libc = "0.2.48"          # MIT/Apache-2.0
 nix = "0.13.0"           # MIT
 openssl = "0.10"         # Apache-2.0
+crossbeam = "0.7.1"      # MIT/Apache-2.0
+num_cpus = "1.10.0"      # MIT/Apache-2.0
+tokio = "0.1.22"         # MIT
+futures-preview = { version = "0.3.0-alpha.17", features = ["compat", "io-compat"] }
+tokio-io = "0.1.12"
```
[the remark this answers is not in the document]
parthsane: `AllowStdIo` is not in `tokio::io`. Once the `stdout` and `stderr` thread based implementations are done, we can remove the tokio-io dependency

enclave-runner/src/usercalls/mod.rs
parthsane: futures::Mutex was the thing that least amount of changes, hence I went for that. But for some things tokio::Lock was better (eg our stdin implementation). 
I've changed all the mutex to Lock and used a macro to unwrap.

enclave-runner/src/usercalls/mod.rs
parthsane: `io_future` is written in such a way that it never returns. It keeps on looping. 
If I understand the way select works, it means that only the `return_future` can reach this point in the code

enclave-runner/src/usercalls/mod.rs
parthsane: so, as I understand this, we should have a function like `poll_read_alloc` in the trait `AsyncStream` which can then also be overridden by the implementers of AsyncStream.

jethrogb: Yes

enclave-runner/src/usercalls/mod.rs
```diff
@@ -605,94 +1067,42 @@ pub trait UsercallExtension : 'static + Send + Sync + std::fmt::Debug {
...
         Box::new(value)
     }
 }
 
 #[derive(Debug)]
 struct UsercallExtensionDefault;
-impl UsercallExtension for UsercallExtensionDefault{}
+impl UsercallExtension for UsercallExtensionDefault {}
 
 #[allow(unused_variables)]
```
mzohreva: Please remove this annotation and fix all unused variable warnings in the impl block.

enclave-runner/src/usercalls/mod.rs
```diff
@@ -811,56 +1242,30 @@ impl RunningTcs {
...
+                pending_event_set: 0,
+                event_queue: new_tcs.event_queue,
+                mode: EnclaveEntry::ExecutableNonMain,
+            },
+            entry: CoEntry::Initial(new_tcs.tcs, 0, 0, 0, 0, 0),
+        });
+        match ret {
+            Ok(()) => Ok(()),
+            Err(err) => {
+                let y = err.into_inner();
```
mzohreva: Maybe put this back in the `threads_queue`?

enclave-runner/Cargo.toml
```diff
@@ -29,6 +30,11 @@ lazy_static = "1.2.0"                           # MIT/Apache-2.0
 libc = "0.2.48"                                 # MIT/Apache-2.0
 nix = "0.13.0"                                  # MIT
 openssl = { version = "0.10", optional = true } # Apache-2.0
+crossbeam = "0.7.1"                             # MIT/Apache-2.0
+num_cpus = "1.10.0"                             # MIT/Apache-2.0
+tokio = "0.1.22"                                # MIT
+futures-preview = { version = "0.3.0-alpha.17", features = ["compat", "io-compat"] }
```
mzohreva: Is it possible to use the latest version of futures now without much change to the code? If so, please update this to use https://crates.io/crates/futures version 0.3.4

enclave-runner/src/usercalls/mod.rs
```diff
@@ -82,143 +107,353 @@ impl<W: Write> Write for WriteOnly<W> {
...
-        self.0.lock().flush()
+pub trait AsyncStream: AsyncRead + AsyncWrite + 'static + Send + Sync {
+    fn poll_read_alloc(&mut self) -> tokio::prelude::Poll<Vec<u8>, std::io::Error>
+    {
+        let mut v: Vec<u8> = vec![0; 8192];
+        self.poll_read(v.as_mut_slice()).map(move |size| {
+            match size {
+                Async::NotReady => Async::NotReady,
+                Async::Ready(size) => {
+                    v.resize(size, 0);
```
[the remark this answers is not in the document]
parthsane: What is the reason to use truncate over resize? I understand that truncate will not increase the size of the vector
Why do we want to limit the size to `8192`?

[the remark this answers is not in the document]
mzohreva: Because `poll_read()` is only given a mutable slice (not the vec) which means it is limited to read up to 8K. Using truncate is better because we don't want to allocate unnecessarily in case the function returns  a number > 8K

mzohreva:
bors merge

PR #172, opened 2019-09-15
Move Quote types and utilities to crate

nhynes:
This PR moves what was previously `dcap_ql::quote` to its own crate so to allow other packages to use the `Quote` types without pulling in the dependencies of (the completely independent) `dcap` libraries. This is particularly useful for utilities around quote generation and verification that do not want to use the C++ implementation.


jethrogb:
Thanks! lgtm if you can get CI passing

nhynes:
Cool, thanks!

Also, I saw that the Cargo.tomls didn't follow a consistent format, so I formatted them according to the LSB-upgradeable (i.e. [carat](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#caret-requirements)) pattern that's now very common.

nhynes:
nvm. I ended up [rewriting it](https://github.com/nhynes/sgx-quote). Still need to add tests and fuzzing, though..

PR #173, opened 2019-09-16
support for pck cert chain-based attesation

lkatalin:
This is a first attempt at adding some additional support for attestation based on the PCK certificate chain (rather than PPID) in the `dcap-ql`. The relevant changes are:

- An implementation of `Qe3CertData` for `CertificationDataType::PckCertificateChain`
- An associated function on `Quote` to retrieve the material signed by the Attestation Key, i.e. the Quote Header || Report Body. This does not utilize the `parse` function or other data types in the `dcap-ql` because the byte vector required for successful attestation needs to include the Quote Version info (which is abstracted to `V3` in the `QuoteHeader` type).
- An associated function on `Quote` to retrieve the hash of the Attestation Key || QE Auth Data, which also must be verified for attestation.

I used these in outside attestation code and they were successful; however, I have not written additional tests for the `dcap_ql` crate.

I'm happy to make any suggested changes.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -84,7 +84,7 @@ dependencies = [
 
 [[package]]
 name = "autocfg"
-version = "0.1.1"
+version = "0.1.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 
 [[package]]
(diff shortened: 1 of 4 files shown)


jethrogb:
Superseded by #237

PR #177, opened 2019-09-20
Quote serde

lkatalin:
This PR is meant to provide `serde` support to serialize and deserialize Quote, Report, and Targetinfo structs if the (new) "serde_support" feature is enabled.

### DIFF
--- a/dcap-ql/Cargo.toml
+++ b/dcap-ql/Cargo.toml
@@ -41,6 +41,7 @@ lazy_static = "1"   # MIT/Apache-2.0
 libc = "0.2"        # MIT/Apache-2.0
 num-derive = "0.2"  # MIT/Apache-2.0
 num-traits = "0.2"  # MIT/Apache-2.0
+serde = { version = "1.0.104", features = ["derive"], optional = true } # MIT/Apache-2.0
 
 [dev-dependencies]
 "report-test" = { version = "0.2.0", path = "../report-test" }
--- a/sgx-isa/Cargo.toml
+++ b/sgx-isa/Cargo.toml
@@ -22,6 +22,7 @@ categories = ["hardware-support"]
 [dependencies]
 # External dependencies
 bitflags = "1" # MIT/Apache-2.0
+serde = { version = "1.0.104", features = ["derive"], optional = true } # MIT/Apache-2.0
 
 [features]
 large_array_derive = []
--- a/sgx-isa/src/lib.rs
+++ b/sgx-isa/src/lib.rs
@@ -23,11 +23,90 @@ extern crate std;
 
 #[macro_use]
 extern crate bitflags;
+
+#[cfg(feature = "serde")]
+extern crate serde;
+
+#[cfg(feature = "serde")]
+use serde::{Serialize, Deserialize};
+
 #[cfg(all(feature = "sgxstd", target_env = "sgx"))]
 use std::os::fortanix_sgx::arch;
 #[cfg(all(feature = "nightly", target_env = "sgx", not(feature = "sgxstd")))]
 mod arch;
 use core::{convert::TryFrom, num::TryFromIntError, slice};
+
+
+#[cfg(feature = "serde")]
+mod array_64 {
+    use serde::{
+        de::{Deserializer, Error, SeqAccess, Visitor},
+        ser::{SerializeTuple, Serializer},
+    };
+    use core::fmt;
+
+    const LEN: usize = 64;
+
+    pub fn serialize<S: Serializer>(array: &[u8; LEN], serializer: S) -> Result<S::Ok, S::Error> {
+        let mut seq = serializer.serialize_tuple(LEN)?;
+        for elem in &array[..] {
+            seq.serialize_element(elem)?;
+        }
+        seq.end()
+    }
+
+    pub fn deserialize<'de, D: Deserializer<'de>>(deserializer: D) -> Result<[u8; LEN], D::Error> {
+        struct ArrayVisitor;
+        impl<'de> Visitor<'de> for ArrayVisitor {
+            type Value = [u8; LEN];
+            fn expecting(&self, formatter: &mut fmt::Formatter) -> fmt::Result {
+                write!(formatter, "an array of length 64")
+            }
+            fn visit_seq<A: SeqAccess<'de>>(self, mut seq: A) -> Result<[u8; LEN], A::Error> {
+                let mut arr = [0; LEN];
+                for i in 0..LEN {
+                    arr[i] = seq
+                        .next_element()?
+                        .ok_or_else(|| A::Error::invalid_length(i, &self))?;
+                }
+                Ok(arr)
+            }
+        }
+        deserializer.deserialize_tuple(64, ArrayVisitor)
+    }
+}
+
+// These helper functions implement defaults for structs' reserved fields,
+// which is necessary for serde support.
+#[cfg(feature = "serde")]
+fn report_reserved1() -> [u8; 28] {
+    [0u8; 28]
+}
+
+#[cfg(feature = "serde")]
+fn report_reserved2() -> [u8; 32] {
+    [0u8; 32]
+}
+
+#[cfg(feature = "serde")]
+fn report_reserved3() -> [u8; 96] {
+    [0u8; 96]
+}
+
+#[cfg(feature = "serde")]
+fn report_reserved4() -> [u8; 60] {
+    [0u8; 60]
+}
+
+#[cfg(feature = "serde")]
+fn ti_reserved1() -> [u8; 4] {
+    [0u8; 4]
+}
+
+#[cfg(feature = "serde")]
+fn ti_reserved2() -> [u8; 456] {
+    [0u8; 456]
+}
 
 #[cfg(not(feature = "large_array_derive"))]
 #[macro_use]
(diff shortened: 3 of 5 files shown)


dcap-ql/Cargo.toml
```diff
@@ -25,6 +25,7 @@ required-features = ["test-sgx-flc"]
 # If set, will link to libsgx_dcap_ql. Otherwise, will use the dynamic loader
 # to access the library at runtime.
 link = ["dcap-ql-sys/link"]
+serde_support = ["serde"]
```
nickmooney: Per the [manifest docs](https://doc.rust-lang.org/cargo/reference/manifest.html), this should probably be `serde-support` (dashes, not underscores).

dcap-ql/src/lib.rs
```diff
@@ -4,9 +4,11 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-#![doc(html_logo_url = "https://edp.fortanix.com/img/docs/edp-logo.svg",
-       html_favicon_url = "https://edp.fortanix.com/favicon.ico",
-       html_root_url = "https://edp.fortanix.com/docs/api/")]
+#![doc(
```
nickmooney: Noise from `cargo fmt` -- maybe better as a separate PR?

dcap-ql/src/lib.rs
```diff
@@ -20,6 +22,8 @@ extern crate libc;
 #[macro_use]
 extern crate num_derive;
 extern crate num_traits;
+#[cfg(feature = "serde_support")]
+extern crate serde;
```
nickmooney: `extern crate` is no longer needed in most cases in Rust 2018, but this is an issue separate from your PR :). In the future it might be worth another PR to change the `edition` so we can get rid of the instances `extern crate` and `macro_use`.

sgx-isa/src/lib.rs
```diff
@@ -12,9 +12,11 @@
 //! [isdm]: https://www-ssl.intel.com/content/www/us/en/processors/architectures-software-developer-manuals.html
 
 #![no_std]
-#![doc(html_logo_url = "https://edp.fortanix.com/img/docs/edp-logo.svg",
-       html_favicon_url = "https://edp.fortanix.com/favicon.ico",
-       html_root_url = "https://edp.fortanix.com/docs/api/")]
+#![doc(
```
nickmooney: ditto `cargo fmt`

sgx-isa/src/lib.rs
```diff
@@ -23,11 +25,50 @@ extern crate std;
...
+#[macro_use]
+#[cfg(feature = "serde_support")]
+extern crate serde_big_array;
+
 #[cfg(all(feature = "sgxstd", target_env = "sgx"))]
 use std::os::fortanix_sgx::arch;
 
 use core::{convert::TryFrom, num::TryFromIntError, slice};
 
+// big_array! is used to support serilization for large arrays in structs.
```
nickmooney: nitpick: should have newlines between these functions

dcap-ql/Cargo.toml
jethrogb: Actually, I think the feature should just be called `serde`. And specifying an optional dependency does this already. So, there's no need to specifically create a feature.

lkatalin: I had to leave this as an explicitly stated `serde-support` feature because otherwise I can't specify feature dependencies.

lkatalin:
@jethrogb @nickmooney I've made the changes; let me know what you think. 

The build fails now due to this error:
`error: could not compile 'syntex_syntax'.` 
but I think this has nothing to do with my code. Will this stop it from being merged?

jethrogb:
Yeah that's unfortunately an issue with syntex_syntax on newer Rust compilers. Has nothing to do with your code We're working on a fix. However, we can't land things that don't pass CI.

sgx-isa/src/lib.rs
```diff
@@ -552,18 +602,24 @@ struct_def! {
     feature = "large_array_derive",
     derive(Clone, Debug, Default, Eq, PartialEq)
 )]
+#[cfg_attr(feature = "serde_support", derive(Serialize, Deserialize))]
 pub struct Report {
     pub cpusvn: [u8; 16],
     pub miscselect: Miscselect,
+    #[cfg_attr(feature = "serde_support", serde(default = "report_reserved1"), serde(skip))]
```
jethrogb: nit: you can combine these like so: `serde(default = "report_reserved1", skip)`

sgx-isa/Cargo.toml
```diff
@@ -23,7 +23,10 @@ categories = ["hardware-support"]
 [dependencies]
 # External dependencies
 bitflags = "1" # MIT/Apache-2.0
+serde = { version = "1.0.84", features = ["derive"], optional = true } #MIT/Apache-2.0
+serde-big-array = { version = "0.1.5", optional = true } # MIT/Apache-2.0
 
 [features]
 large_array_derive = []
 sgxstd = []
+serde-support = ["serde", "serde-big-array"]
```
jethrogb: I see. I think it's not worth using this dependency right now, since it's only used in one place and it prevents us from having the feature name we want. It will also make fixing the situation in the future (when const generics become a thing) harder.

How about we just add this in `lib.rs`?
```rust
#[cfg(feature = "serde")]
mod array_64 {
    use serde::{
        de::{Deserializer, Error, SeqAccess, Visitor},
        ser::{SerializeTuple, Serializer},
    };
    use core::fmt;

    const LEN: usize = 64;

    pub fn serialize<S: Serializer>(array: &[u8; LEN], serializer: S) -> Result<S::Ok, S::Error> {
        let mut seq = serializer.serialize_tuple(LEN)?;
        for elem in &array[..] {
            seq.serialize_element(elem)?;
        }
        seq.end()
    }

    pub fn deserialize<'de, D: Deserializer<'de>>(deserializer: D) -> Result<[u8; LEN], D::Error> {
        struct ArrayVisitor;
        impl<'de> Visitor<'de> for ArrayVisitor {
            type Value = [u8; LEN];
            fn expecting(&self, formatter: &mut fmt::Formatter) -> fmt::Result {
                write!(formatter, "an array of length 64")
            }
            fn visit_seq<A: SeqAccess<'de>>(self, mut seq: A) -> Result<[u8; LEN], A::Error> {
                let mut arr = [0; LEN];
                for i in 0..LEN {
                    arr[i] = seq
                        .next_element()?
                        .ok_or_else(|| A::Error::invalid_length(i, &self))?;
                }
                Ok(arr)
            }
        }
        deserializer.deserialize_tuple(64, ArrayVisitor)
    }
}
```

jethrogb:
@lkatalin are you still interested in landing this PR?

sgx-isa/Cargo.toml
lkatalin: Could you give an example of how to use this?
`#[cfg_attr(feature = "serde", serde(default = "array_64::serialize"))]` is clearly not right...

lkatalin: `Serialize` is fine, but trying to add `Deserialize` on `TargetInfo`, for example, leads to 
`the trait 'core::default::Default' is not implemented for '[u8; 456]'` once I remove `serde-big-array`.

lkatalin:
@jethrogb I was waiting on the `syntex_syntax` issue, which is resolved now. If you have some guidance you can give on how to use the above serialization method, I can update the code. So far, removing `serde-big-array` breaks everything.

sgx-isa/Cargo.toml
jethrogb: Same way you were using big array before, so `serde(with = "array_64")`

jethrogb: I'm not sure what your point is about `TargetInfo`, that's already working correctly in the current PR?

lkatalin: Ah, thanks, you're right. This helps. I think I can get it working now.

lkatalin:
@jethrogb I got this working, but then I rebased it onto master and I get `failed to run custom build command for `mbedtls-sys-auto v2.18.2` with all kinds of bindgen errors. This happens even in the master branch without my changes. Unsure if it's my system; is there a fix for this that you know? I haven't pushed the rebased version yet.

lkatalin:
@jethrogb It should be good now. Let me know if not.

sgx-isa/src/lib.rs
```diff
@@ -23,12 +23,92 @@ extern crate std;
...
+
 #[cfg(all(feature = "sgxstd", target_env = "sgx"))]
 use std::os::fortanix_sgx::arch;
 #[cfg(all(feature = "nightly", target_env = "sgx", not(feature = "sgxstd")))]
 mod arch;
 use core::{convert::TryFrom, num::TryFromIntError, slice};
 
+
+#[cfg(feature = "serde")]
+#[allow(dead_code)]
```
jethrogb: I'm surprised this is necessary

sgx-isa/src/lib.rs
```diff
@@ -67,6 +147,7 @@ macro_rules! struct_def {
     (
         #[repr(C $(, align($align:tt))*)]
         $(#[cfg_attr(feature = "large_array_derive", derive($($cfgderive:meta),*))])*
+	$(#[cfg_attr(feature = "serde", derive($($serdederive:meta),*))])*
```
jethrogb: Used tab instead of spaces in a couple places here

jethrogb:
bors r+

Sorry it took so long to merge this!

PR #181, opened 2019-10-11
Support upstream driver

jethrogb:
See also the [Linux kernel documentation for __vdso_sgx_enter_enclave](https://github.com/torvalds/linux/blob/84292fffc2468125632a21c09533a89426ea212e/arch/x86/include/uapi/asm/sgx.h#L123-L166)


sgxs-loaders/src/isgx/mod.rs
```diff
@@ -207,10 +310,27 @@ impl EnclaveLoad for InnerDevice {
     }
 }
 
+#[derive(Copy, Clone, Debug, PartialEq, Eq, Hash)]
+pub enum DriverVersion {
+    External,
+    Upstream,
```
parthsane: I don't think Upstream is the best name for this.
Today it may be upstream but it might not be tomorrow

enclave-runner/src/tcs.rs
```diff
@@ -83,6 +83,50 @@ pub(crate) fn coenter<T: Tcs>(
...
+        // Strongly link to another symbol in the VDSO, so that the linker will
+        // include a DT_NEEDED entry for `linux-vdso.so.1`. This doesn't happen
+        // automatically because rustc passes `--as-needed` to the linker. This
+        // is never executed because of the unconditional jump above.
+.global __vdso_clock_gettime
+        call __vdso_clock_gettime@PLT
+
+2:      pushq $$0                                            // push argument: handler = NULL
+        pushq $$0                                            // push argument: e = NULL
+        push %rbx                                            // push argument: tcs
```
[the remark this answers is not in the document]
jethrogb: We don't support 32-bit platforms

sgxs-loaders/src/isgx/mod.rs
jethrogb: Any suggestions?

parthsane: I remember that I was confused about what Upstream means.
Does it mean the driver that is in the master branch in the linux repo? In that case the name Upstream is fine

If it's a specific version of the driver then the name should reflect that. 

But I think my concerns come from not understanding the context. Maybe a comment would help understand it

jethrogb: Yes upstream here means part of the Linux kernel

jethrogb:
Ready for review, tested against Linux 5.11-rc2

enclave-runner/src/tcs.rs
```diff
@@ -54,31 +57,145 @@ pub(crate) fn coenter<T: Tcs>(
...
+                }
+            }
+
+            let mut run = SgxEnclaveRun {
+                tcs: tcs.address() as _,
+                ..Default::default()
+            };
+            let ret: i32;
+            asm!("
+                    sub $0x8, %rsp                   // align stack
```
raoulstrackx, line 139: Why is this required?

jethrogb: Call stack alignment is discussed in section 3.2.2 of the [ABI](https://github.com/hjl-tools/x86-psABI/wiki/x86-64-psABI-1.0.pdf). According to [`asm!` documentation](https://doc.rust-lang.org/beta/unstable-book/library-features/asm.html#options-1):

> If [the `nostack`] option is not used then the stack pointer is guaranteed to be suitably aligned (according to the target ABI) for a function call.

However, we are passing a single argument on the stack, so we need to make sure the stack is properly aligned after the stack arguments.

raoulstrackx: Oh I didn't know about that requirement. Good to know!

jethrogb:
Non-rebase part of my latest push:
```patch
diff --git a/.travis.yml b/.travis.yml
index 48980cc..83d2c28 100644
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,7 +29,7 @@ matrix:
       before_script:
       script:
-        - cargo test --verbose --all --exclude sgxs-loaders
+        - cargo test --verbose --all --exclude sgxs-loaders && [ "$(echo $(nm -D target/debug/sgx-detect|grep __vdso_sgx_enter_enclave))" = "w __vdso_sgx_enter_enclave" ]
         - cargo test --verbose -p sgx-isa --features sgxstd --target x86_64-fortanix-unknown-sgx --no-run
         - cargo test --verbose -p sgxs-tools --features pe2sgxs --bin isgx-pe2sgx
         - cargo test --verbose -p dcap-ql --features link
diff --git a/enclave-runner/src/tcs.rs b/enclave-runner/src/tcs.rs
index a890867..35c8676 100644
--- a/enclave-runner/src/tcs.rs
+++ b/enclave-runner/src/tcs.rs
@@ -75,6 +75,7 @@ pub(crate) fn coenter<T: Tcs>(
                 // `--as-needed` to the linker. This is never executed because
                 // of the unconditional jump above.
 .global __vdso_clock_gettime
+.type __vdso_clock_gettime, function
                 call __vdso_clock_gettime@PLT
 
 1:
@@ -138,6 +139,8 @@ pub(crate) fn coenter<T: Tcs>(
             asm!("
                     sub $0x8, %rsp                   // align stack
                     push {}                          // push argument: run
+.weak __vdso_sgx_enter_enclave
+.type __vdso_sgx_enter_enclave, function
                     call __vdso_sgx_enter_enclave@PLT
                     add $0x10, %rsp                  // restore stack pointer
                 ",
@@ -163,7 +166,7 @@ pub(crate) fn coenter<T: Tcs>(
                 match sgx_result.try_into() {
                     Ok(Enclu::EExit) => { /* normal case */ },
                     Ok(Enclu::EResume) => {
-                       if let Some(mut debug_buf) = debug_buf {
+                        if let Some(mut debug_buf) = debug_buf {
                             let _ = write!(&mut debug_buf[..], "Enclave triggered exception: {:?}\0", run);
                         } else {
                             eprintln!("Enclave triggered exception, treating as panic: {:?}", run);
```

PR #185, opened 2019-10-25
Pvs/no_std_nightly (Add nightly feature for no_std use of ereport function from libstd)

parthsane:
I have solved the review comments. 
I am unsure of how the egetkey function would be used though

### DIFF
--- a/sgx-isa/src/arch.rs
+++ b/sgx-isa/src/arch.rs
@@ -1,0 +1,66 @@
+/* Copyright (c) Fortanix, Inc.
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+use core::mem::MaybeUninit;
+use super::Enclu;
+
+/// Wrapper struct to force 16-byte alignment.
+#[repr(align(16))]
+pub struct Align16<T>(pub T);
+
+/// Wrapper struct to force 128-byte alignment.
+#[repr(align(128))]
+pub struct Align128<T>(pub T);
+
+/// Wrapper struct to force 512-byte alignment.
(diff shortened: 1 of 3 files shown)


sgx-isa/src/lib.rs
```diff
@@ -725,4 +724,4 @@ fn test_eq() {
 
     b.miscmask = 0xdeadbeef;
     assert!(a == b);
-}
```
jethrogb: Revert whitespace changes to this file

sgx-isa/src/arch.rs
```diff
@@ -0,0 +1,74 @@
+#[cfg(all(feature = "nightly", target_env = "sgx"))]
```
jethrogb: You shouldn't need to have any `cfg` in this file. This file should only be included when needed, and then everything in this file should be defined.

sgx-isa/src/lib.rs
```diff
@@ -22,10 +23,10 @@ extern crate std;
 
 #[macro_use]
 extern crate bitflags;
-
 #[cfg(all(feature = "sgxstd", target_env = "sgx"))]
 use std::os::fortanix_sgx::arch;
-
+#[cfg(all(feature = "nightly", target_env = "sgx"))]
```
jethrogb: If you define both `sgxstd` and `nightly`, you'll get a namespace conflict. You should make it such that `sgxstd` takes precedence.

sgx-isa/src/lib.rs
```diff
@@ -22,10 +23,10 @@ extern crate std;
 
 #[macro_use]
 extern crate bitflags;
-
 #[cfg(all(feature = "sgxstd", target_env = "sgx"))]
 use std::os::fortanix_sgx::arch;
-
+#[cfg(all(feature = "nightly", target_env = "sgx", not(feature = sgx)))]
```
jethrogb: Typo

PR #190, opened 2019-11-08
Allow using external crypto for dummy signature

mzohreva:
and make `Command` `Send`

### DIFF
--- a/enclave-runner/Cargo.toml
+++ b/enclave-runner/Cargo.toml
@@ -17,15 +17,18 @@ categories = ["os", "hardware-support"]
 
 [dependencies]
 # Project dependencies
-sgxs = { version = "0.7.0", path = "../sgxs", features = ["crypto-openssl"] }
+sgxs = { version = "0.7.0", path = "../sgxs" }
 fortanix-sgx-abi = { version = "0.3.0", path = "../fortanix-sgx-abi" }
 sgx-isa = { version = "0.3.0", path = "../sgx-isa" }
 
 # External dependencies
-failure = "0.1.1"        # MIT/Apache-2.0
-failure_derive = "0.1.1" # MIT/Apache-2.0
-fnv = "1"                # MIT/Apache-2.0
-lazy_static = "1.2.0"    # MIT/Apache-2.0
-libc = "0.2.48"          # MIT/Apache-2.0
-nix = "0.13.0"           # MIT
-openssl = "0.10"         # Apache-2.0
+failure = "0.1.1"                               # MIT/Apache-2.0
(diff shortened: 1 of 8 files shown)


jethrogb:
Can you please rebase on master?

enclave-runner/src/loader.rs
```diff
@@ -136,20 +149,31 @@ impl<'a> EnclaveBuilder<'a> {
             miscselect: None,
             signature: None,
             usercall_ext : None,
+            load_and_sign: None,
+            hash_enclave: None,
         };
```
jethrogb: Please make a new private method `fn new_with_source(EnclaveSource) -> EnclaveBuilder` that you can call in both `new` and `new_from_memory`

enclave-runner/Cargo.toml
```diff
@@ -17,15 +17,18 @@ categories = ["os", "hardware-support"]
...
+failure = "0.1.1"                               # MIT/Apache-2.0
+failure_derive = "0.1.1"                        # MIT/Apache-2.0
+fnv = "1"                                       # MIT/Apache-2.0
+lazy_static = "1.2.0"                           # MIT/Apache-2.0
+libc = "0.2.48"                                 # MIT/Apache-2.0
+nix = "0.13.0"                                  # MIT
+openssl = { version = "0.10", optional = true } # Apache-2.0
+
+[features]
+default = ["openssl", "sgxs/crypto-openssl"]
```
jethrogb: Since the code requires a combination of `openssl` and `sgxs/crypto-openssl`, I think you should also specify the features here as follows:
```toml
[features]
default = ["crypto-openssl"] 
crypto-openssl = ["openssl", "sgxs/crypto-openssl"] 
```
And then use `cfg(feature = "crypto-openssl")` throughout

enclave-runner/src/loader.rs
```diff
@@ -163,15 +174,34 @@ impl<'a> EnclaveBuilder<'a> {
...
+            None => Err(format_err!("either compile with default features or use with_dummy_signature_signer()"))
+        }
     }
 
     pub fn dummy_signature(&mut self) -> &mut Self {
         self.signature = None;
         self
     }
 
+    pub fn with_dummy_signature_signer<H, E, S, T, F>(&mut self, load_key: F)
```
jethrogb: Please add a doc comment describing at least the signature of `F`.

I think the appropriate order for the type parameters is H, S, F, E, T. This is based on what I think someone might want to specify vs. what someone might want to have inferred using `_`, with inference-preferred parameters last.

mzohreva:
bors r=jethrogb

PR #191, opened 2019-11-11
add an option to override the runner

aandyl:
remove `--signature dummy` (to allow operations with runners that do not
implement it) because it is anyways the default in ftxsgx-runner


Goirad:
Why do you say that it is the default in ftxsgx-runner? Checking there the `None` variant does not call a signature method on the builder.

aandyl:
> Why do you say that it is the default in ftxsgx-runner? Checking there the `None` variant does not call a signature method on the builder.

From enclave-runner/src/loader.rs:

    pub fn dummy_signature(&mut self) -> &mut Self {
        self.signature = None;
        self
    }

    fn load<T: Load>(mut self, loader: &mut T) -> Result<(Vec<ErasedTcs>, *mut c_void, usize), Error> {
        let signature = match self.signature {
            Some(sig) => sig,
            None => self
                .generate_dummy_signature()
                .context("While generating dummy signature")?,
        };

Goirad:
I see, it is the default in enclave-runner, not in ftxsgx-runner.

PR #194, opened 2019-11-21
Pvs/async queues alt

parthsane:
I have created the PR. 

I know of two things that I am unsure of
1) I don't know how exactly to test the implementation
2) the `fn async_queues` on being called the second time should return something "equivalent to calling exit(true)" and I am unsure if the current code does that.


enclave-runner/src/usercalls/mod.rs
```diff
@@ -793,12 +865,98 @@ impl EnclaveState {
...
+            thread::spawn(move || {
+
+                tokio::runtime::current_thread::block_on_all(tokio::prelude::future::poll_fn(|| -> Poll<(),()> {
+                    loop {
+                        let val = unsafe {enclave_clone.async_queue_usercall.recv()};
+                        match val {
+                            None => {
+                                // sleep
+                                // retry after waking up
+                                reader_waker_clone.register();
```
mzohreva: How is this supposed to be woken up when the enclave sends a usercall?

fortanix-sgx-abi/src/lib.rs
```diff
@@ -702,10 +703,10 @@ pub mod async {
     pub struct FifoDescriptor<T> {
         /// Pointer to the queue memory. Must have a size of
         /// `len * size_of::<T>()` bytes and have alignment `align_of::<T>`.
```
mzohreva: please update the doc comments to refer to the correct field names

fortanix-sgx-abi/src/lib.rs
```diff
@@ -702,10 +703,10 @@ pub mod async {
...
-        pub data: *mut T,
+        pub data: *const Slot<T>,
         /// The number of elements pointed to by `data`. Must be a power of two
         /// less than or equal to 2³¹.
```
mzohreva: should be 2 ^ 30 now?

fortanix-sgx-abi/src/lib.rs
```diff
@@ -702,10 +703,10 @@ pub mod async {
...
-        pub len: usize,
+        pub capacity: u32,
```
jethrogb: Please don't change the ABI

fortanix-sgx-abi/src/lib.rs
jethrogb: Why?

async-queues/src/lib.rs
```diff
@@ -0,0 +1,270 @@
...
+    /// `0` indicates this slot is empty.
+    id: AtomicUsize,
+    data: UnsafeCell<T>,
+}*/
+
+pub struct FifoDescriptorContainer<T> {
+    pub inner: FifoDescriptor<T>,
+}
+
+unsafe impl <T:Copy> Sync for FifoDescriptorContainer<T> {}
```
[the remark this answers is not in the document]
parthsane: i've replace the Copy with Send

enclave-runner/src/usercalls/mod.rs
parthsane: :man_facepalming: 
I have changed the code for both of the threads to be woken up on any synchronous usercall according to the following
https://edp.fortanix.com/docs/api/fortanix_sgx_abi/async/index.html#enclaveuserspace-synchronization

async-queues/src/lib.rs
```diff
@@ -0,0 +1,249 @@
...
+    }
+
+    fn len(&self, capacity: u32) -> usize {
+        ((if self.read <= self.write { 0 } else { 2 * capacity }) + self.write - self.read) as usize
+    }
+}
+
+impl<T: Copy> FifoDescriptorContainer<T> {
+    pub fn new(capacity: usize) -> Self {
+        assert!(capacity <= usize::pow(2, 31));
```
Goirad: In the documentation below you specify that the capacity must be a power of 2 less than `2^31` but here and I think in one other place you check only that it is a number less than `2^31`.

jethrogb:
Closing in favor of #246

PR #203, opened 2019-12-09
elf2sgx: Remove .text_no_sgx section during conversion

Pagten:
This pull request is related to [rust-lang pull request #67084](https://github.com/rust-lang/rust/pull/67084), which places the ELF entry point in a separate section of the ELF file. This pull request removes/overwrites that section during the conversion from ELF to SGXS, because that code is not needed in the enclave. To avoid changing the relative position of other code, the section is only removed if it is at the end of a program segment. It is overwritten with NOPs otherwise.

The pull request also includes some general refactoring of the code related to Splices.

### DIFF
--- a/fortanix-sgx-tools/src/bin/ftxsgx-elf2sgxs.rs
+++ b/fortanix-sgx-tools/src/bin/ftxsgx-elf2sgxs.rs
@@ -384,6 +384,14 @@ impl<'a> LayoutInfo<'a> {
                 u64_to_bytes(enclave_size),
             ));
         }
+        if let Some(sec_no_sgx) = self.elf.find_section_by_name(".text_no_sgx") {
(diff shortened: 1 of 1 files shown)


fortanix-sgx-tools/src/bin/ftxsgx-elf2sgxs.rs
```diff
@@ -66,22 +66,43 @@ fn u64_to_bytes(v: u64) -> Vec<u8> {
     }
 }
 
-struct Splice(u64, Vec<u8>);
+struct Splice {
+    address: u64,
+    value: Vec<u8>,
+    truncate: bool, /* remove the splice if it is at the end of a segment */
```
jethrogb: Use a doc comment:
```rust
/// Remove the splice if it is at the end of a segment
truncate: bool,
```

PR #211, opened 2020-01-17
Add Windows CI

parthsane:
Adding all the tests which pass on Windows

As and when other tests pass, they can be added


parthsane:
@Goirad @jethrogb @mzohreva 
Can anyone approve this?

PR #215, opened 2020-02-17
eagerly print panics based on a new config option

Goirad:
Add an EnclaveBuilder method to set a panic_on_abort flag.

### DIFF
--- a/enclave-runner/src/loader.rs
+++ b/enclave-runner/src/loader.rs
@@ -66,6 +66,7 @@ pub struct EnclaveBuilder<'a> {
     usercall_ext: Option<Box<dyn UsercallExtension>>,
     load_and_sign: Option<Box<dyn FnOnce(Signer) -> Result<Sigstruct, Error>>>,
     hash_enclave: Option<Box<dyn FnOnce(&mut EnclaveSource<'_>) -> Result<EnclaveHash, Error>>>,
+    forward_panics: bool,
 }
 
 #[derive(Debug, Fail)]
@@ -138,6 +139,7 @@ impl<'a> EnclaveBuilder<'a> {
             usercall_ext: None,
             load_and_sign: None,
             hash_enclave: None,
+            forward_panics: false,
         };
 
         let _ = ret.coresident_signature();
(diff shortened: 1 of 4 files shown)


enclave-runner/src/usercalls/mod.rs
```diff
@@ -714,9 +717,13 @@ impl EnclaveState {
                                     println!("Attaching debugger");
                                     #[cfg(all(unix, not(target_abi = "musl")))]
                                     trap_attached_debugger(usercall.tcs_address() as _).await;
-                                    Err(EnclaveAbort::Exit {
-                                        panic: EnclavePanic::from(buf.into_inner()),
-                                    })
+                                    let panic = EnclavePanic::from(buf.into_inner());
+                                    println!("{}", &panic);
```
jethrogb: This print shouldn't be here

Goirad: I think I'm missing something. I thought the point was to print the panic message as soon as the runner received one from the enclave.

Goirad: Unless you mean that it should be inside the if

jethrogb: I'm ok with it inside the if. But then it should be an `eprintln` and it should print something like `Enclave panicked: {}`

Goirad: That is perfectly reasonable. What do you think about panicking with that message instead of aborting the process?

jethrogb: That seems fine but the option should be called `forward_panics`

mzohreva: The prefix `Enclave panicked:` is already in the message produced by `EnclavePanic::from(buf.into_inner());` I think.

Goirad: Good point Mohsen, I will omit that then

enclave-runner/src/loader.rs
```diff
@@ -248,10 +250,19 @@ impl<'a> EnclaveBuilder<'a> {
         self.usercall_ext = Some(extension.into());
     }
 
+    /// Whether to abort the runner if any thread panics.
```
jethrogb: Update docs to new name

enclave-runner/src/usercalls/mod.rs
```diff
@@ -714,9 +717,12 @@ impl EnclaveState {
                                     println!("Attaching debugger");
                                     #[cfg(all(unix, not(target_abi = "musl")))]
                                     trap_attached_debugger(usercall.tcs_address() as _).await;
-                                    Err(EnclaveAbort::Exit {
-                                        panic: EnclavePanic::from(buf.into_inner()),
-                                    })
+                                    let panic = EnclavePanic::from(buf.into_inner());
+                                    if enclave_clone.forward_panics {
+                                        panic!("{}", &panic);
```
mzohreva: I think just using eprintln! is better as this one produces messages like the following:
```
thread 'main' panicked at 'Enclave panicked: thread '<unnamed>' panicked at 'spawned thread panicking...', src/main.rs:6:9
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
', enclave-runner/src/usercalls/mod.rs:722:41
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

Goirad: @jethrogb Did you see this?

jethrogb: I'm indifferent about printing but I do like panicking instead of plain exiting.

PR #216, opened 2020-02-26
disable the debugger trap when building for musl

aandyl:
when targeting musl, the rust libc crate does not have a compatible ucontext_t definition. (musl itself will provide one if _GNU_SOURCE is defined, but the libc crate does not do that.)

This change by itself does not result in a working musl build of the enclave runner, due to openssl build issues. When I add the "vendored" openssl feature, I am able to produce a working statically-linked musl build of the enclave runner.

I also tried building with the `rust:alpine` docker image, but that failed with proc macro build problems.

### DIFF
--- a/enclave-runner/src/usercalls/mod.rs
+++ b/enclave-runner/src/usercalls/mod.rs
@@ -4,8 +4,9 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
+#[cfg(all(unix, not(target_abi = "musl")))]
 extern crate libc;
-#[cfg(unix)]
+#[cfg(all(unix, not(target_abi = "musl")))]
 extern crate nix;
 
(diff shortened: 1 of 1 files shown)


jethrogb:
This PR did nothing because `target_abi` is never set. I think you mean `target_env`.

jethrogb:
Also, the musl definition of [ucontext_t](https://rust-lang.github.io/libc/x86_64-unknown-linux-musl/libc/struct.ucontext_t.html) seems to be compatible to me.

aandyl:
> Also, the musl definition of [ucontext_t](https://rust-lang.github.io/libc/x86_64-unknown-linux-musl/libc/struct.ucontext_t.html) seems to be compatible to me.

It did not always have gregs. https://github.com/rust-lang/libc/commit/216d8008b661d9a13aa0ccc31f44e35e4ac857b5

Not sure how I came up with `target_abi`. I had originally added a cargo feature and later tried to eliminate it, apparently replacing it with something broken.

PR #217, opened 2020-02-29
Add Async queues

parthsane:
Added functionality for async queues to improve performance of usercalls
Added logic for sleep and wake up of async queue polling threads
Added AsyncQueueWaiter for waking up Enclave TCS


jethrogb:
Closing in favor of #246

Issue #31, opened 2018-11-20
Add timeout support


jethrogb:
The following operations in `std` require some form of timeout support:

* TcpStream::connect_timeout
* TcpStream::set_read_timeout
* TcpStream::set_write_timeout
* thread::sleep
* Condvar::wait_timeout

Currently, the first just doesn't use a timeout, and the rest errors/panics.

The ABI currently does not support waiting a specified amount (only no wait or indefinite wait): https://docs.rs/fortanix-sgx-abi/0.1.0-rc1/fortanix_sgx_abi/struct.Usercalls.html#method.wait . It's also not clear if timeouts make sense in SGX setting.

jethrogb:
Should the API use deadlines or timeouts?

PR #222, opened 2020-03-19
 Add implementation of AESM client for SGX 

parthsane:
The implementation of AESM for SGX takes in a TcpStream for its constructor.
It omits the get_launch_token functionality

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,6 +32,7 @@ matrix:
         - cargo test --verbose -p sgx-isa --features sgxstd -Z package-features --target x86_64-fortanix-unknown-sgx --no-run
         - cargo test --verbose -p sgxs-tools --features pe2sgxs --bin isgx-pe2sgx -Z package-features
         - cargo test --verbose -p dcap-ql --features link -Z package-features
+        - cargo build --verbose -p aesm-client --target=x86_64-fortanix-unknown-sgx
 
     - os: windows
       before_install:
@@ -55,4 +56,3 @@ matrix:
         - cargo test --verbose -p sgxs
         - cargo test --verbose -p sgxs-loaders
         - cargo test --verbose -p sgxs-tools --features pe2sgxs --bin isgx-pe2sgx -Z package-features
-  
--- a/aesm-client/src/imp/aesm_protobuf/mod.rs
+++ b/aesm-client/src/imp/aesm_protobuf/mod.rs
@@ -1,0 +1,106 @@
+use imp::AesmClient;
+pub use error::{AesmError, Error, Result};
+use protobuf::Message;
+use std::io::{Read, Write};
+use byteorder::{LittleEndian, NativeEndian, ReadBytesExt, WriteBytesExt};
+use {
+    AesmRequest, FromResponse, QuoteInfo, QuoteResult, QuoteType,
+    Request_GetQuoteRequest, Request_InitQuoteRequest,
+};
+// FIXME: remove conditional compilation after resolving https://github.com/fortanix/rust-sgx/issues/31
+#[cfg(not(target_env = "sgx"))]
+use imp::{LOCAL_AESM_TIMEOUT_US, REMOTE_AESM_TIMEOUT_US};
+// FIXME: remove conditional compilation after resolving https://github.com/fortanix/rust-sgx/issues/31
+#[cfg(not(target_env = "sgx"))]
+use std::time::Duration;
+
+impl AesmClient {
+    pub fn try_connect(&self) -> Result<()> {
+        self.open_socket().map(|_| ())
+    }
+
+    pub(super) fn transact<T: AesmRequest>(&self, req: T) -> Result<T::Response> {
+        let mut sock = self.open_socket()?;
+
+        // FIXME: remove conditional compilation after resolving https://github.com/fortanix/rust-sgx/issues/31
+        #[cfg(not(target_env = "sgx"))]
+        let _ = sock.set_read_timeout(req.get_timeout().map(|t| Duration::from_micros(t as _)))?;
+
+        let req_bytes = req
+            .into()
+            .write_to_bytes()
+            .expect("Failed to serialize protobuf");
+        sock.write_u32::<NativeEndian>(req_bytes.len() as u32)?;
+        sock.write_all(&req_bytes)?;
+
+        let res_len = sock.read_u32::<NativeEndian>()?;
+        let mut res_bytes = vec![0; res_len as usize];
+        sock.read_exact(&mut res_bytes)?;
+
+        let res = T::Response::from_response(protobuf::parse_from_bytes(&res_bytes))?;
+        Ok(res)
+    }
+
+    /// Obtain target info from QE.
+    pub fn init_quote(&self) -> Result<QuoteInfo> {
+        #[allow(unused_mut)]
+        let mut req = Request_InitQuoteRequest::new();
+        // FIXME: remove conditional compilation after resolving https://github.com/fortanix/rust-sgx/issues/31
+        #[cfg(not(target_env = "sgx"))]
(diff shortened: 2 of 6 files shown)


aesm-client/src/imp/sgx.rs
```diff
@@ -0,0 +1,119 @@
...
+}
+
+impl AesmClient {
+    pub fn new(tcp_stream: TcpStream) -> Self {
+        AesmClient{ tcp_stream }
+    }
+
+    fn open_socket(&self) -> Result<TcpStream> {
+        let sock = self.tcp_stream.try_clone()?;
+        let _ = sock.set_write_timeout(Some(Duration::from_micros(LOCAL_AESM_TIMEOUT_US as _)))?;
```
parthsane: I've kept set_write_timeout and set_read_timeout, but I remember that they don't function correctly. 
Should I remove them?

aesm-client/src/lib.rs
```diff
@@ -192,6 +212,7 @@ impl AesmClient {
         )
     }
 
+    #[cfg(not(target_env = "sgx"))]
     pub fn get_launch_token(
```
parthsane: as of now, the crate doesnt compile for unix (in master branch) without using the `feature="sgxs"` flag.
Is that a something we want?
I believe one solution would be to add a `#[cfg(feature = "sgxs")]` for `get_launch_token`

aesm-client/src/imp/sgx.rs
jethrogb: You can comment them out with a `FIXME` pointing to #31

aesm-client/src/lib.rs
jethrogb: Oops. That's probably an oversight. Please add the feature yes

.travis.yml
```diff
@@ -32,6 +32,7 @@ matrix:
         - cargo test --verbose -p sgx-isa --features sgxstd -Z package-features --target x86_64-fortanix-unknown-sgx --no-run
         - cargo test --verbose -p sgxs-tools --features pe2sgxs --bin isgx-pe2sgx -Z package-features
         - cargo test --verbose -p dcap-ql --features link -Z package-features
+        - pushd aesm-client/ && cargo build --target=x86_64-fortanix-unknown-sgx && popd
```
jethrogb: Please use `-p` syntax as in the lines above

aesm-client/src/imp/unix.rs
```diff
@@ -150,6 +153,7 @@ impl AesmClient {
     }
 
     /// Obtain launch token
+    #[cfg(feature = "sgxs")]
```
jethrogb: Please test building without this feature in CI

aesm-client/src/lib.rs
```diff
@@ -150,15 +157,29 @@ impl QuoteResult {
     }
 }
 
+#[cfg(not(target_env = "sgx"))]
 #[derive(Default, Debug, Clone)]
 pub struct AesmClient {
     inner: imp::AesmClient
 }
 
+// TODO: Find a better alternative to removing Clone and Default trait
```
jethrogb: use `cfg_attr`

aesm-client/src/lib.rs
```diff
@@ -150,15 +157,29 @@ impl QuoteResult {
...
+}
+
+
 impl AesmClient {
+    #[cfg(not(target_env = "sgx"))]
     pub fn new() -> Self {
         AesmClient { inner: imp::AesmClient::new() }
     }
+    #[cfg(target_env = "sgx")]
+    pub fn new(tcp_stream: TcpStream) -> Self {
```
jethrogb: Don't do this, use an extension trait (see unix with_path)

aesm-client/src/lib.rs
```diff
@@ -31,10 +31,14 @@ extern crate sgx_isa;
...
 #[cfg(feature = "sgxs")]
 use std::result::Result as StdResult;
+#[cfg(target_env = "sgx")]
+use std::net::TcpStream;
 
 use protobuf::ProtobufResult;
 #[cfg(feature = "sgxs")]
 use sgxs::einittoken::{Einittoken, EinittokenProvider};
+#[cfg(not(target_env = "sgx"))]
+#[cfg(feature = "sgxs")]
```
jethrogb: Consolidate these into a single attribute using `all(...)` (also below)

aesm-client/src/imp/unix.rs
jethrogb: Same thing needed for Windows?

aesm-client/src/imp/sgx.rs
```diff
@@ -0,0 +1,129 @@
...
+    tcp_stream: TcpStream,
+}
+
+impl AesmClient {
+    pub fn new(tcp_stream: TcpStream) -> Self {
+        AesmClient { tcp_stream }
+    }
+
+    fn open_socket(&self) -> Result<TcpStream> {
+        let sock = self.tcp_stream.try_clone()?;
```
jethrogb: `try_clone` never fails on SGX

aesm-client/src/imp/sgx.rs
```diff
@@ -0,0 +1,129 @@
...
+/// remote servers, such as provisioning EPID.
+const REMOTE_AESM_TIMEOUT_US: u32 = 30_000_000;
+*/
+
+#[derive(Debug)]
+pub struct AesmClient {
+    tcp_stream: TcpStream,
+}
+
+impl AesmClient {
```
jethrogb: A lot of this code is shared with unix, it would be good to use a common implementation somehow. i.e. create another `mod aesm_protobuf` or so that you depend on in both unix and sgx.

.travis.yml
```diff
@@ -32,6 +32,8 @@ matrix:
         - cargo test --verbose -p sgx-isa --features sgxstd -Z package-features --target x86_64-fortanix-unknown-sgx --no-run
         - cargo test --verbose -p sgxs-tools --features pe2sgxs --bin isgx-pe2sgx -Z package-features
         - cargo test --verbose -p dcap-ql --features link -Z package-features
+        - cargo build --verbose -p aesm-client
```
jethrogb: I'm confused, the sgxs feature is already not set on line 31. Why do we need to test this separately?

aesm-client/src/lib.rs
```diff
@@ -150,12 +162,16 @@ impl QuoteResult {
     }
 }
 
-#[derive(Default, Debug, Clone)]
+// TODO: Find a better alternative to removing Clone and Default trait
+#[cfg_attr(not(target_env = "sgx"), derive(Default,Clone))]
```
jethrogb: Why is clone conditional? You should manually implement `Clone` for imp::AesmClient on SGX since that's guaranteed to work

.travis.yml
parthsane: :man_facepalming: 
My mistake. I overlooked that somehow

aesm-client/src/imp/aesm_protobuf/mod.rs
```diff
@@ -0,0 +1,114 @@
...
+        let res_len = sock.read_u32::<NativeEndian>()?;
+        let mut res_bytes = vec![0; res_len as usize];
+        sock.read_exact(&mut res_bytes)?;
+
+        let res = T::Response::from_response(protobuf::parse_from_bytes(&res_bytes))?;
+        Ok(res)
+    }
+
+    // FIXME: remove conditional compilation after resolving https://github.com/fortanix/rust-sgx/issues/31
+    #[cfg(not(target_env = "sgx"))]
```
jethrogb: Just do this conditional inline (as in `get_quote`). You can apply `#[allow(unused_mut)]` on the binding.

Issue #109, opened 2019-03-07
Command doesn't exit if thread is stuck in blocking I/O


jethrogb:
This program may not terminate:

```rust
use std::{net::TcpListener, sync::mpsc, thread};

fn main() {
    let (tx, rx) = mpsc::channel();
    thread::spawn(move || {
        let listen = TcpListener::bind("0:0").unwrap();
        tx.send(()).unwrap();
        while let Ok(_) = listen.accept() {}
    });
    rx.recv().unwrap();
    println!("Exiting main thread");
}
```

This is because when the main thread exits, the second thread is likely to be blocking in an `accept()` call. Therefore, that thread never terminates and the main thread (out of the enclave) is waiting indefinitely for that second thread to finish.

jethrogb:
I expect this to be fixed alongside #18.

jethrogb:
Closing since #18 is merged. Ok, @mzohreva ?

mzohreva:
Yeah, this was fixed in #223

PR #210, opened 2020-01-14
Port to tokio 0.2

Goirad:
Ports the enclave runner async branch to use tokio 0.2, in the process removing a lot of compat code and shims. I have tested this against the rust standard library tests.

PR #223, opened 2020-03-19
Enable enclave exit when a TCS is blocked on I/O

mzohreva:
This fixes issue #109

### DIFF
--- a/enclave-runner/src/usercalls/mod.rs
+++ b/enclave-runner/src/usercalls/mod.rs
@@ -560,11 +560,11 @@ impl Work {
         let usercall_send_data = match self.entry {
             CoEntry::Initial(erased_tcs, p1, p2, p3, p4, p5) => {
                 let coresult = tcs::coenter(erased_tcs, p1, p2, p3, p4, p5, Some(&buf));
-                ((coresult, self.tcs, buf))
+                (coresult, self.tcs, buf)
             }
             CoEntry::Resume(usercall, coresult) => {
                 let coresult = usercall.coreturn(coresult, Some(&buf));
-                ((coresult, self.tcs, buf))
+                (coresult, self.tcs, buf)
             }
         };
         // if there is an error do nothing, as it means that the main thread has exited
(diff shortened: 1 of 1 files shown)


mzohreva:
Open questions:
- Do we need to add the exit handler for other calls such as `alloc`/`free`/`async_queues`/`insecure_time`, etc?
- Is there a better way to avoid creating an extra future?

enclave-runner/src/usercalls/mod.rs
```diff
@@ -1083,6 +1092,23 @@ impl<T: UsercallExtension> From<T> for Box<dyn UsercallExtension> {
 struct UsercallExtensionDefault;
 impl UsercallExtension for UsercallExtensionDefault {}
 
+macro_rules! with_exit_handler {
```
jethrogb: This seems complicated, I'd much rather have something where we make use of tokio's task cancellation mechanisms such as https://docs.rs/tokio/0.2.13/tokio/runtime/struct.Runtime.html#method.shutdown_timeout

jethrogb: Also, if this is wrapping all usercalls, couldn't it be done one level down in the stack (in usercall dispatch)?

jethrogb:
Can we add some tests? Maybe they could be part of the rust-lang/rust test suite.

enclave-runner/src/usercalls/mod.rs
mzohreva: > use of tokio's task cancellation mechanisms such as https://docs.rs/tokio/0.2.13/tokio/runtime/struct.Runtime.html#method.shutdown_timeout

That would require upgrading to tokio 0.2, I'll explore that option using PR #210

> couldn't it be done one level down in the stack (in usercall dispatch)?

Good idea, I'll give that a shot

mzohreva:
Just found out that the parallel future approach implemented here is not necessary. The blocking issue appears to be due to the following line in `syscall_loop()`:
```rust
tokio::runtime::current_thread::block_on_all(select_fut.unit_error().compat()).unwrap()
```
The documentation for `tokio::runtime::current_thread::block_on_all()` says:

> This first creates a new Runtime, and calls Runtime::block_on with the provided future, which blocks the current thread until the provided future completes. It then calls Runtime::run to wait for any other spawned futures to resolve.

The problematic part is the `It then calls Runtime::run to wait for any other spawned futures to resolve.` and that is not what we need in syscall_loop. We want to return from syscall_loop as soon as the select future returns.

I'll update this PR accordingly.

enclave-runner/src/usercalls/mod.rs
```diff
@@ -776,14 +776,15 @@ impl EnclaveState {
         };
 
         let select_fut =
-            futures::future::select(return_future.boxed(), io_future.boxed()).map( |either| {
+            futures::future::select(return_future.boxed_local(), io_future.boxed_local()).map( |either| {
```
jethrogb: Please add a comment that you know `io_future` isn't going to resolve, but that's intentional.

PR #224, opened 2020-03-24
Check enclave page 0 is not executable

raoulstrackx:
LVI mitigations require the first enclave page to not be executable.
https://software.intel.com/security-software-guidance/insights/deep-dive-load-value-injection#adhocLBmitigation


fortanix-sgx-tools/src/bin/ftxsgx-elf2sgxs.rs
```diff
@@ -460,6 +460,12 @@ impl<'a> LayoutInfo<'a> {
                 }
             }
 
+            // To defend against LVI attacks, the first page of an enclave should not be executable.
```
jethrogb: I think line 429 is a more appropriate place for this check.

PR #225, opened 2020-03-27
write entire request at once

aandyl:
It seems like each write turns into a separate enclave I/O, so, even if the helper service ought to be able to tolerate that, there is a small performance gain from combining them.

### DIFF
--- a/aesm-client/src/imp/aesm_protobuf/mod.rs
+++ b/aesm-client/src/imp/aesm_protobuf/mod.rs
@@ -2,6 +2,7 @@ use imp::AesmClient;
 pub use error::{AesmError, Error, Result};
 use protobuf::Message;
 use std::io::{Read, Write};
+use std::mem::size_of;
 use byteorder::{LittleEndian, NativeEndian, ReadBytesExt, WriteBytesExt};
 use {
     AesmRequest, FromResponse, QuoteInfo, QuoteResult, QuoteType,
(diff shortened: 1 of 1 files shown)


aesm-client/src/imp/aesm_protobuf/mod.rs
```diff
@@ -26,11 +26,14 @@ impl AesmClient {
         #[cfg(not(target_env = "sgx"))]
         let _ = sock.set_read_timeout(req.get_timeout().map(|t| Duration::from_micros(t as _)))?;
 
-        let req_bytes = req
-            .into()
-            .write_to_bytes()
+        let mut req_bytes = vec![0u8; 4];
+        let mut cursor = Cursor::new(&mut req_bytes);
+        cursor.set_position(4);
```
jethrogb: You don't need a cursor, [`impl io::Write for Vec<u8>` appends to the vector](https://doc.rust-lang.org/1.42.0/std/io/trait.Write.html#impl-Write-9).

aesm-client/src/imp/aesm_protobuf/mod.rs
```diff
@@ -26,11 +26,12 @@ impl AesmClient {
         #[cfg(not(target_env = "sgx"))]
         let _ = sock.set_read_timeout(req.get_timeout().map(|t| Duration::from_micros(t as _)))?;
 
-        let req_bytes = req
-            .into()
-            .write_to_bytes()
+        let mut req_bytes = vec![0u8; 4];
```
jethrogb: Do you want to use `size_of::<u32>()` in place of `4` everywhere? It's ok if not.

aandyl: Yeah, I'll update that.

jethrogb:
bors retry

PR #226, opened 2020-04-02
Port enclave-runner to tokio 0.2

mzohreva:
This is mostly the same as PR #210.

cc @Goirad

### DIFF
--- a/enclave-runner/src/library.rs
+++ b/enclave-runner/src/library.rs
@@ -17,7 +17,6 @@ use std::fmt;
 use std::os::raw::c_void;
 
 pub struct Library {
-    rt: tokio::runtime::Runtime,
     enclave: Arc<EnclaveState>,
     address: *mut c_void,
     size: usize,
@@ -50,9 +49,8 @@ impl Library {
         usercall_ext: Option<Box<dyn UsercallExtension>>,
         forward_panics: bool,
     ) -> Library {
-        let (rt, enclave) = EnclaveState::library(tcss, usercall_ext, forward_panics);
+        let enclave = EnclaveState::library(tcss, usercall_ext, forward_panics);
         Library {
-            rt,
             enclave,
             address,
             size,
@@ -71,13 +69,13 @@ impl Library {
     /// The caller must ensure that the parameters passed-in match what the
     /// enclave is expecting.
     pub unsafe fn call(
-        &mut self,
+        &self,
         p1: u64,
         p2: u64,
         p3: u64,
         p4: u64,
         p5: u64,
     ) -> Result<(u64, u64), Error> {
-        EnclaveState::library_entry(&mut self.rt, &self.enclave, p1, p2, p3, p4, p5)
+        EnclaveState::library_entry(&self.enclave, p1, p2, p3, p4, p5)
     }
 }
--- a/enclave-runner/src/library.rs
+++ b/enclave-runner/src/library.rs
@@ -49,9 +49,8 @@ impl Library {
         usercall_ext: Option<Box<dyn UsercallExtension>>,
         forward_panics: bool,
     ) -> Library {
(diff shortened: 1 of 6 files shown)


enclave-runner/src/library.rs
```diff
@@ -68,13 +71,13 @@ impl Library {
     /// The caller must ensure that the parameters passed-in match what the
     /// enclave is expecting.
     pub unsafe fn call(
-        &self,
+        &mut self,
```
mzohreva, line 74: This is a public API change, so we need a minor version bump

enclave-runner/src/usercalls/mod.rs
```diff
@@ -162,236 +139,243 @@ impl Read for Stdin {
...
+        self.poll_read(cx, v.as_mut_slice()).map(move |size| {
+            size.map(|size| { v.resize(size, 0); v })
         })
     }
 }
 
 impl<S: AsyncRead + AsyncWrite + Sync + Send + 'static> AsyncStream for S {}
 
 /// AsyncListener lets an implementation implement a slightly modified form of `std::net::TcpListener::accept`.
-pub trait AsyncListener: 'static + Send {
```
mzohreva, line 209: Why remove `'static`?

enclave-runner/src/library.rs
Goirad: Are there docs somewhere (maybe the EDP website?) that also need to be updated?

jethrogb: This API change is not appropriate. Multiple threads should be able to call into the enclave.

enclave-runner/src/usercalls/mod.rs
```diff
@@ -1226,18 +1216,7 @@ impl<'tcs> IOHandlerInput<'tcs> {
             return Ok(self.alloc_fd(AsyncFileDesc::stream(stream_ext)).await);
         }
 
-        // try to connect to all socket addresses one by one
```
mzohreva, line 1229: This is now unnecessary since `tokio::net::TcpStream::connect()` takes care of it in tokio 0.2 (similar to `std::net::TcpStream`)

Goirad: Wow is there anything tokio 0.2 doesn't do now?

enclave-runner/src/usercalls/mod.rs
```diff
@@ -644,8 +634,15 @@ impl EnclaveState {
     ) -> StdResult<(u64, u64), EnclaveAbort<EnclavePanic>> {
         let (tx_return_channel, mut rx_return_channel) = tokio::sync::mpsc::unbounded_channel();
         let enclave_clone = enclave.clone();
+        let mut rt = tokio::runtime::Builder::new()
+            .basic_scheduler()
```
parthsane, line 638: As I understand basic_scheduler means everything runs on a single thread.
It's what_ we are aiming for.
But do we want to see if a (multi) threaded_scheduler improves performance?

mzohreva: We can experiment with that, but I don't expect that would help. This might actually explain why we saw lower performance for tokio 0.2 implementation over 0.1 before, as we were using multi-threaded scheduler during that test for 0.2.

enclave-runner/src/usercalls/mod.rs
```diff
@@ -148,7 +124,7 @@ impl Read for Stdin {
         }
 
         lazy_static::lazy_static! {
-            static ref STDIN: Lock<AsyncStdin> = {
+            static ref STDIN: futures::lock::Mutex<AsyncStdin> = {
```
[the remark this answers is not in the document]
mzohreva: We need to use `futures::lock::Mutex` in some case because it allows calling `lock()` for `T: ?Sized` while tokio's Mutex does not. I replaced the tokio Mutex use with futures one

PR #228, opened 2020-04-21
Libunwind update + toolchain verification

AdrianCX:
Changes are needed to:

1. Allow libunwind update.
- We now have 4 symbols to reference start and size of sections .eh_frame and .eh_frame_hdr
- We need to support previous versions as well.
- Symbols are not reused from previous version to cause failure at compile time if old tool is used with new libunwind. (also added version check code to prevent this issue in the future and provide nice messages)

2. Fail gracefully if toolchain version is unsupported (give user a nice message telling to update tools)

### DIFF
--- a/fortanix-sgx-tools/src/bin/ftxsgx-elf2sgxs.rs
+++ b/fortanix-sgx-tools/src/bin/ftxsgx-elf2sgxs.rs
@@ -48,9 +48,9 @@ struct Symbols<'a> {
     EH_FRM_HDR_SIZE: Option<&'a DynSymEntry>,
     TEXT_BASE: &'a DynSymEntry,
     TEXT_SIZE: &'a DynSymEntry,
-    EH_FRM_ADDR: Option<&'a DynSymEntry>,
+    EH_FRM_OFFSET: Option<&'a DynSymEntry>,
     EH_FRM_LEN: Option<&'a DynSymEntry>,
-    EH_FRM_HDR_ADDR: Option<&'a DynSymEntry>,
+    EH_FRM_HDR_OFFSET: Option<&'a DynSymEntry>,
     EH_FRM_HDR_LEN: Option<&'a DynSymEntry>,
 }
 struct SectionRange {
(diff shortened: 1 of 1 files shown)


fortanix-sgx-tools/src/bin/ftxsgx-elf2sgxs.rs
```diff
@@ -154,11 +166,62 @@ macro_rules! check_size {
...
+                }
+
+                let desc = header.desc(ptr);
+                
+                if desc.len() != 4 {
+                    bail!("Expecting 32 bit 'toolchain-version' in .note.x86_64-fortanix-unknown-sgx");
+                }
+
+                // According to entry.S in rust-lang 'desc - toolchain version number, 32-bit LE'
+                let version : u32 = ((desc[0] as u32) << 0) + ((desc[1] as u32) << 8) + ((desc[2] as u32) << 16) + ((desc[3] as u32) << 24);
```
Pagten, line 205: You could write this as `let version = u32::from_le_bytes(desc.try_into().unwrap());`. The unwrap is justified by the check in line 200.

Goirad: Absolutey, definitely use that function instead of manually bitshifting

fortanix-sgx-tools/src/bin/ftxsgx-elf2sgxs.rs
```diff
@@ -201,6 +258,20 @@ impl<'a> LayoutInfo<'a> {
             bail!("ENCLAVE_SIZE symbol is not naturally aligned")
         }
 
+        if let (Some(EH_FRM_HDR_BASE), Some(EH_FRM_HDR_SIZE)) = (syms.EH_FRM_HDR_BASE, syms.EH_FRM_HDR_SIZE) {
+            check_size!(EH_FRM_HDR_BASE == 8);
+            check_size!(EH_FRM_HDR_SIZE == 8);
+        }
```
Goirad, line 264: Minor thing, but I believe it is standard practice to put the else on the same line as the closing bracket

fortanix-sgx-tools/src/bin/ftxsgx-elf2sgxs.rs
```diff
@@ -201,6 +258,20 @@ impl<'a> LayoutInfo<'a> {
...
+        else if let (Some(EH_FRM_OFFSET), Some(EH_FRM_LEN), Some(EH_FRM_HDR_OFFSET), Some(EH_FRM_HDR_LEN)) = (syms.EH_FRM_OFFSET, syms.EH_FRM_LEN, syms.EH_FRM_HDR_OFFSET, syms.EH_FRM_HDR_LEN) {
+            check_size!(EH_FRM_OFFSET == 8);
+            check_size!(EH_FRM_LEN == 8);
+            check_size!(EH_FRM_HDR_OFFSET == 8);
+            check_size!(EH_FRM_HDR_LEN == 8);
+        }
+        else {
+            bail!("Missing EH Frame header symbols, application must either have (EH_FRM_HDR_BASE/EH_FRM_HDR_SIZE) or (EH_FRM_OFFSET, EH_FRM_LEN, EH_FRM_HDR_OFFSET, EH_FRM_HDR_LEN");
```
Goirad, line 272: I don't like that this message isn't more specific about what is missing, but I can't think of an elegant way of adding that specificity

AdrianCX: I tried to improve on it, if you have other suggestions let me know

fortanix-sgx-tools/src/bin/ftxsgx-elf2sgxs.rs
AdrianCX: fixed, thanks

fortanix-sgx-tools/src/bin/ftxsgx-elf2sgxs.rs
AdrianCX: fixed, thanks

jethrogb:
bors r=pagten

PR #229, opened 2020-04-21
Fix exit bug

mzohreva:
`fn exit()` calling `abort_all_threads()` created a race condition:
return future in `fn syscall_loop()` would detect the exit and break
the syscall loop by receiving either the EnclaveAbort from the thread
that caused the exit or an EnclaveAbort::Secondary from other threads
that observed the exit in `fn dispatch()`.

To fix the issue:
- Avoid calling `abort_all_threads()` in `fn exit`. Do that only
  after the syscall loop is broken.
- Ignore EnclaveAbort::Secondary from non-main threads in return
  future which is not strictly needed after the previous change,
  but would make the code easier to read.

Also, when a non-main thread calls `std::process::exit()` with any
exit status, it should terminate the enclave execution. This was
not handled correctly for exit status 0.

### DIFF
--- a/enclave-runner/src/usercalls/mod.rs
+++ b/enclave-runner/src/usercalls/mod.rs
@@ -646,17 +646,17 @@ impl EnclaveState {
                 rx_return_channel = stream;
                 let (my_result, mode) = work;
                 let res = match (my_result, mode) {
-                    (e, EnclaveEntry::Library)
-                    | (e, EnclaveEntry::ExecutableMain)
-                    | (e @ Err(EnclaveAbort::Secondary), EnclaveEntry::ExecutableNonMain) => e,
+                    (e, EnclaveEntry::Library) |
+                    (e, EnclaveEntry::ExecutableMain)
+                        => e,
+
+                    (Err(EnclaveAbort::Secondary), EnclaveEntry::ExecutableNonMain) |
                     (Ok(_), EnclaveEntry::ExecutableNonMain) => {
                         continue;
                     }
-                    (Err(e @ EnclaveAbort::Exit { .. }), EnclaveEntry::ExecutableNonMain)
-                    | (
-                        Err(e @ EnclaveAbort::InvalidUsercall(_)),
-                        EnclaveEntry::ExecutableNonMain,
-                    ) => {
+
+                    (Err(e @ EnclaveAbort::Exit { .. }), EnclaveEntry::ExecutableNonMain) |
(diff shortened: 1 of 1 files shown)


parthsane:
LGTM. 
But since I suggested the change for removing `abort_all_threads()` from `fn exit`, maybe someone else can also take a look
@Goirad @jethrogb

enclave-runner/src/usercalls/mod.rs
```diff
@@ -719,9 +720,14 @@ impl EnclaveState {
...
                                     }
                                     Err(EnclaveAbort::Exit{ panic })
-
                                 }
-                                Err(EnclaveAbort::Exit { panic: false }) => Ok((0, 0)),
+                                Err(EnclaveAbort::Exit { panic: false }) => match state.mode {
+                                    EnclaveEntry::ExecutableMain => Ok((0, 0)),
+                                    EnclaveEntry::Library |
+                                    EnclaveEntry::ExecutableNonMain => Err(EnclaveAbort::Exit{
+                                        panic: EnclavePanic::DebugStr("Execution aborted by a thread".to_owned()),
```
mzohreva, line 728: We should probably change this so that calling `std::process::exit(0)` from a secondary thread is similar to the main thread doing the same (i.e. not generating a panic). What do you think @jethrogb?

Goirad: I think any non main thread exiting is a violation of the sgx spec

mzohreva: I don't think it is. Do you have a reference?

jethrogb: In fact this is explicitly allowed (because Rust allows it)

mzohreva: I have already addressed this in b8a65d89e72f8922210fe6d0ec5ffe817c44cccc. @jethrogb could you do a final review?

enclave-runner/src/usercalls/mod.rs
```diff
@@ -1285,7 +1287,6 @@ impl<'tcs> IOHandlerInput<'tcs> {
 
     #[inline(always)]
     fn exit(&mut self, panic: bool) -> EnclaveAbort<bool> {
-        self.enclave.abort_all_threads();
```
jethrogb, line 1288: I don't think we should remove this. We should avoid scheduling other threads ASAP after exit is called. Instead, can we have special logic in `syscall_loop` that waits for an exit that isn't `EnclaveAbort::Secondary`?

mzohreva: That's cleaner, I updated with the PR.

PR #230, opened 2020-04-22
aesm-client: Add timeout values to send to quoting enclave for sgx

parthsane:
Fixes bug where calls to aesm from sgx module hangs indefinitely

Also make imported failure crate public


parthsane:
I want to make the failure crate public so that its Compat functionality is easier to use by the user of the aesm crate

PR #236, opened 2020-05-08
Add strrchr to rs-libc

nhynes:
This is useful for compiling sqlite3. The functions were swiped from [musl](https://git.musl-libc.org/cgit/musl/tree/src/string/memrchr.c), like the rest of them.


jethrogb:
Looks good! Can you also update `rs-libc/LICENSE`?

jethrogb:
Released 0.2.1

PR #241, opened 2020-05-18
Acruceru/dcap provider error

AdrianCX:


dcap-provider/src/lib.rs
```diff
@@ -262,9 +262,12 @@ pub extern "C" fn sgx_ql_get_quote_config(
                         // variable, but there is no guarantee that the second
                         // call to this function is going to happen on the same
                         // stack.
+
+                        // We may get called back from aesm service if library injected there.
+                        // In that case we need to return PlatformLibUnavailable in order to have AESM use internal data.
```
jethrogb, line 267: Change comment:
```
// If this is a re-entry, return PlatformLibUnavailable so the raw TCB is used to generate
// quotes instead of certification data provided by us.
```

PR #246, opened 2020-06-03
Implement async queues

mzohreva:
- Implement async queues
- Enable async usercalls in enclave-runner
- Limit memory usage for pending events

This is an alternative to @parthsane's PRs #194 and #217, with the following improvements:
- Has tests for the FIFO implementation
- Fixes the issue with calling async_queues usercall more than once
- Improves memory usage for pending events by removing a `VecDeque` that could grow unbounded


mzohreva:
Also this does not change the ABI.

Here is an example enclave code using async queues:

```rust
#![feature(sgx_platform)]

extern crate async_queue;
extern crate fortanix_sgx_abi;

use async_queue::{
    QueueEvent, Receiver, Sender,
    Synchronizer, SynchronizationError,
};
use fortanix_sgx_abi::{FifoDescriptor, Return, Usercall};
use std::os::fortanix_sgx::usercalls::alloc::User;
use std::os::fortanix_sgx::usercalls::raw::{
    self, UsercallNrs,
    EV_RETURNQ_NOT_EMPTY, EV_USERCALLQ_NOT_FULL,
    WAIT_INDEFINITE,
};
use std::{io, mem, thread};

#[derive(Clone, Debug)]
enum QueueSynchronizer {
    Usercall,
    Return,
}

impl Synchronizer for QueueSynchronizer {
    fn wait(&self, event: QueueEvent) -> Result<(), SynchronizationError> {
        let ev = match (self, event) {
            (Self::Usercall, QueueEvent::NotEmpty) => panic!("enclave should not recv on usercall queue"),
            (Self::Return, QueueEvent::NotFull)    => panic!("enclave should not send on return queue"),
            (Self::Usercall, QueueEvent::NotFull)  => EV_USERCALLQ_NOT_FULL,
            (Self::Return, QueueEvent::NotEmpty)   => EV_RETURNQ_NOT_EMPTY,
        };
        unsafe {
            raw::wait(ev, WAIT_INDEFINITE);
        }
        Ok(())
    }

    fn notify(&self, _event: QueueEvent) {
        // any synchronous usercall would do
        unsafe {
            raw::insecure_time();
        }
    }
}

struct AsyncQueues {
    usercall_queue: FifoDescriptor<Usercall>,
    return_queue: FifoDescriptor<Return>,
}

impl AsyncQueues {
    fn new() -> io::Result<Self> {
        let usercall_queue = User::<raw::FifoDescriptor<raw::Usercall>>::uninitialized().into_raw();
        let return_queue = User::<raw::FifoDescriptor<raw::Return>>::uninitialized().into_raw();

        let r = unsafe { raw::async_queues(usercall_queue, return_queue) };
        if r != 0 {
            return Err(io::Error::from_raw_os_error(r));
        }

        let usercall_queue = unsafe {
            User::<raw::FifoDescriptor<raw::Usercall>>::from_raw(usercall_queue)
        }.to_enclave();

        let return_queue = unsafe {
            User::<raw::FifoDescriptor<raw::Return>>::from_raw(return_queue)
        }.to_enclave();

        // For some reason rust considers std::os::fortanix_sgx::usercalls::raw::FifoDescriptor
        // and fortanix_sgx_abi::FifoDescriptor different types, hence the transmutes here.
        let usercall_queue: FifoDescriptor<Usercall> = unsafe { mem::transmute(usercall_queue) };
        let return_queue: FifoDescriptor<Return> = unsafe { mem::transmute(return_queue) };

        Ok(Self { usercall_queue, return_queue })
    }

    fn sender(&self) -> Sender<Usercall, QueueSynchronizer> {
        Sender::new(self.usercall_queue.clone(), QueueSynchronizer::Usercall)
    }

    // Panics if there are existing receivers
    fn receiver(&self) -> Receiver<Return, QueueSynchronizer> {
        Receiver::new(self.return_queue.clone(), QueueSynchronizer::Return)
    }
}

fn main() {
    let queues = AsyncQueues::new().expect("failed to setup async queues");
    let usercall_queue_tx = queues.sender();
    let return_queue_rx = queues.receiver();

    const N: u64 = 100;
    let h = thread::spawn(move || {
        for i in 0..N {
            usercall_queue_tx.send(&Usercall {
                id: 1 + i,
                args: (UsercallNrs::insecure_time as _, 0, 0, 0, 0),
            }).unwrap();
        }
    });

    for _ in 0..N {
        let t = return_queue_rx.recv().unwrap();
        println!("[{:3}] {:?}", t.id, t.value);
    }
    h.join().unwrap();
}
```

Goirad:
The example enclave code seems awfully verbose. Is this something that most (all?) enclaves will have to do?

mzohreva:
> The example enclave code seems awfully verbose. Is this something that most (all?) enclaves will have to do?

No, this will be abstracted away once we port MIO to SGX and users will use tokio (which relies on MIO for low level operations).

async-queue/Cargo.toml
```diff
@@ -0,0 +1,23 @@
+[package]
+name = "async-queue"
```
mzohreva, line 2: We'll need to rename this crate since `async-queue` was taken by someone else recently.

async-queue/src/fifo.rs
```diff
@@ -0,0 +1,403 @@
...
+    }
+
+    pub(crate) fn drop_receiver(&self, data_ptr: usize) {
+        let mut receivers = self.0.lock().unwrap();
+        receivers.remove(&data_ptr);
+    }
+}
+
+// Note: we cannot have an AtomicU64 id in Usercall/Return types since they
+// need to be Copy due to requirements of `UserSafeSized` (see definition of
```
jethrogb, line 166: I don't think so? `UserSafeSized` just needs `FifoDescriptor<T>` to be `Copy`, but that doesn't require `T: Copy`. So it should be possible to just convert the `id` fields to `AtomicU64`? Even if not, I prefer type casting over transmutes (transmuting a reference is pretty pointless).

This is related to something else I noticed: it would be good if we could enforce some kind of `Copy` bound on `WithAtomicId`. We could perhaps rearchitect the whole thing as follows:

```
#[repr(C)]
struct WithId<T: Copy> {
    id: AtomicU64,
    data: T,
}
```

I have a faint memory that we used to have that, but I don't remember the details.

jethrogb (commented):
Note to self: continue reviewing at `interface_sync::Receiver`

async-queue/src/lib.rs
```diff
@@ -0,0 +1,113 @@
...
+pub trait Synchronizer {
+    /// block execution until the specified event happens.
+    fn wait(&self, event: QueueEvent) -> Result<(), SynchronizationError>;
+
+    /// notify all waiters blocked on the specified event for the same Fifo.
+    fn notify(&self, event: QueueEvent);
+}
+
+pub struct Sender<T, S> {
+    descriptor: FifoDescriptor<T>,
```
jethrogb, line 88: For all these types, I think we should convert FifoDescriptor on construction to more useful types `&AtomicOffsets` and `&UnsafeCell<[T]>`

async-queue/src/fifo.rs
```diff
@@ -0,0 +1,403 @@
...
+        fetch_adjust(
+            &*descriptor.offsets,
+            new.read as isize - current.read as isize,
+            Ordering::SeqCst,
+        )
+    };
+
+    // 8. If the queue was full in step 1, signal the writer to wake up.
+    //    ... or became full during read
+    let became_full = Offsets::new(after, descriptor.len as u32).is_full();
```
jethrogb, line 134: This deviation from the spec doesn't make sense. If it's full now, you shouldn't be waking up any writers.

async-queue/src/fifo.rs
```diff
@@ -0,0 +1,403 @@
...
+        //    expected to be written imminently.
+        if id != 0 {
+            break slot;
+        }
+    };
+
+    // 6. Read the data, then store `0` in the `id`.
+    let mut val = T::empty();
+    val.copy_except_id(slot);
+    val.set_id(slot.get_id());
```
jethrogb, line 120: You already read the `id` in step 4

async-queue/src/fifo.rs
```diff
@@ -0,0 +1,403 @@
...
+    };
+
+    // 8. If the queue was full in step 1, signal the writer to wake up.
+    //    ... or became full during read
+    let became_full = Offsets::new(after, descriptor.len as u32).is_full();
+    Ok((val, was_full || became_full))
+}
+
+lazy_static! {
+    pub(crate) static ref RECEIVER_TRACKER: ReceiverTracker = ReceiverTracker::new();
```
jethrogb, line 139: Is this useful? There's a 1000 things that could go wrong when setting up the queue, why check for this? On the other hand I guess this is pretty cheap.

async-queue/src/fifo.rs
```diff
@@ -0,0 +1,403 @@
...
+    val.set_id(slot.get_id());
+    slot.set_id(0);
+
+    // 7. Store the new read offset.
+    let after = unsafe {
+        fetch_adjust(
+            &*descriptor.offsets,
+            new.read as isize - current.read as isize,
+            Ordering::SeqCst,
+        )
```
jethrogb, line 129: Alternatively, you can cast to a smaller atomic and just do a store (which avoids an atomic read-modify-write here).

async-queue/src/fifo.rs
```diff
@@ -0,0 +1,403 @@
...
+            !receivers.insert(data_ptr)
+        };
+        if already_exists {
+            panic!("Multiple receivers for the same Fifo is not allowed.");
+        }
+    }
+
+    pub(crate) fn drop_receiver(&self, data_ptr: usize) {
+        let mut receivers = self.0.lock().unwrap();
+        receivers.remove(&data_ptr);
```
jethrogb, line 161: Assert you removed something

async-queue/src/fifo.rs
```diff
@@ -0,0 +1,403 @@
...
+    }
+
+    pub fn async_sender<S: AsyncSynchronizer>(&mut self, synchronizer: S) -> AsyncSender<T, S> {
+        AsyncSender::new(self.descriptor(), synchronizer)
+    }
+
+    /// Panics if there is an existing (sync or async) receiver for the same queue.
+    pub fn async_receiver<S: AsyncSynchronizer>(&mut self, synchronizer: S) -> AsyncReceiver<T, S> {
+        AsyncReceiver::new(self.descriptor(), synchronizer)
+    }
```
jethrogb, line 51: These APIs seem not that useful to me. If you want to try to create a safer API, consider adding lifetimes to Sender/Receiver (e.g. using PhantomData). Also, you should be able to create a channel `(Sender, Receiver)` in a single call taking `&mut self`. You should then be able to convert one of them (taken by value) into a FifoDescriptor.

async-queue/src/lib.rs
```diff
@@ -0,0 +1,113 @@
...
+///
+/// **NOTE:** Sender and reciever types use FifoDescriptor internally which
+/// does not hold a reference to the Fifo instance, therefore users of these
+/// types must ensure that the Fifo instance lives at least as long as all
+/// senders and receivers for that queue.
+///
+/// **NOTE:** sync and async sender/receiver types should not be used together.
+/// i.e. either use sync senders/receivers or the async ones, but don't mix
+/// sync and async. The interfaces are designed for use in SGX enclaves (sync)
+/// and enclave runner (async).
```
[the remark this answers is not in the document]
jethrogb: Also, it'd be good if we could somehow leverage the `UserSafe` wrappers when used inside the enclave.

async-queue/src/lib.rs
```diff
@@ -0,0 +1,113 @@
...
+    descriptor: FifoDescriptor<T>,
+    synchronizer: S,
+}
+
+pub trait AsyncSynchronizer {
+    /// block execution until the specified event happens.
+    fn wait(&self, event: QueueEvent) -> Pin<Box<dyn Future<Output = Result<(), SynchronizationError>> + '_>>;
+
+    /// notify all waiters blocked on the specified event for the same Fifo.
+    fn notify(&self, event: QueueEvent) -> Pin<Box<dyn Future<Output = ()> + '_>>;
```
jethrogb, line 102: Is this supposed to return a future?

async-queue/src/fifo.rs
```diff
@@ -0,0 +1,403 @@
...
+        let prev = unsafe {
+            (*descriptor.offsets).compare_and_swap(current, new.as_usize(), Ordering::SeqCst)
+        };
+        if prev == current {
+            break (new, was_empty);
+        }
+    };
+
+    // 4. Write the data, then the `id`.
+    let slot = unsafe { &mut *descriptor.data.add(new.write_offset()) };
```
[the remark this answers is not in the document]
mzohreva: The safety requirements of `pointer::offset` and `pointer::add` are exactly the same. In fact:
> `pub unsafe fn add(self, count: usize) -> *mut T`
> Calculates the offset from a pointer (**convenience for .offset(count as isize)**).

Since in this case offsets are always 32 bits long, the cast to isize does not add any additional concern either.

async-queue/src/fifo.rs
mzohreva: I like the idea of re-architecting as you pointed out. I'll try to do that.

The reason for not being able to just change the id fields in `Usercall`/`Return` types should be apparent by looking at the following definitions:
```rust
pub unsafe trait UserSafeSized: Copy + Sized {}
//...
unsafe impl UserSafeSized for Usercall {}
unsafe impl UserSafeSized for Return {}
```

async-queue/src/fifo.rs
```diff
@@ -0,0 +1,403 @@
...
+    //    ... or became full during read
+    let became_full = Offsets::new(after, descriptor.len as u32).is_full();
+    Ok((val, was_full || became_full))
+}
+
+lazy_static! {
+    pub(crate) static ref RECEIVER_TRACKER: ReceiverTracker = ReceiverTracker::new();
+}
+
+pub(crate) struct ReceiverTracker(Mutex<HashSet<usize>>);
```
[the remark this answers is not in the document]
mzohreva: `lazy_static` definition above requires the type to be `Send`. Is this somehow safer?

I think the better alternative is to keep this as is and change the signature of `new_receiver`/`drop_receiver` to take a real pointer type instead of `usize`

async-queue/src/fifo.rs
mzohreva: But it's not full now. It became full while we were reading, now it's not full anymore because we just read something.

async-queue/src/fifo.rs
mzohreva: But we do need to know the write offset to find out if the queue got full while read was in progress.

async-queue/src/lib.rs
```diff
@@ -0,0 +1,113 @@
...
+
+#[derive(Debug, PartialEq, Eq)]
+pub enum TrySendError {
+    QueueFull,
+}
+
+#[derive(Debug, PartialEq, Eq)]
+pub enum TryRecvError {
+    QueueEmpty,
+}
```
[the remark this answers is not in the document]
mzohreva: Unfortunately I don't have a way of detecting that channel has been closed in `try_send`/`try_recv`

async-queue/src/fifo.rs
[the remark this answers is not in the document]
mzohreva: I ended up removing this whole thing

async-queue/src/lib.rs
mzohreva: That's probably just for symmetry. I think we can remove the return type

async-queue/src/lib.rs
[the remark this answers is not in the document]
jethrogb: I see. But shouldn't we at least account for this future possibility in the API?

ipc-queue/Cargo.toml
```diff
@@ -0,0 +1,26 @@
...
+"""
+repository = "https://github.com/fortanix/rust-sgx"
+documentation = "https://edp.fortanix.com/docs/api/ipc_queue/"
+homepage = "https://edp.fortanix.com/"
+keywords = ["sgx", "fifo", "queue", "ipc"]
+categories = ["asynchronous"]
+
+[dependencies]
+fortanix-sgx-abi = { version = "0.3.0", path = "../fortanix-sgx-abi" }
+static_assertions = "1.1.0"
```
jethrogb, line 18: This should be a dev depedency?

ipc-queue/src/fifo.rs
```diff
@@ -0,0 +1,404 @@
...
+                    fifo.offsets.as_ref() as *const AtomicUsize as *mut AtomicUsize
+                ),
+                storage: Storage::Shared(fifo),
+            }
+        }
+    }
+
+    /// Consumes `self` and returns a FifoDescriptor. If `self` was created
+    /// using `from_arc`, it leaks the internal `Arc` copy to ensure the
+    /// resulting descriptor is valid for `'static` lifetime.
```
jethrogb, line 119: This is not great. I think you should return some type that can produce a `FifoDescriptor<T>` but that the caller needs to hold on to for as long as it knows that the other end is using it (e.g. for as long as the enclave lives).

ipc-queue/src/fifo.rs
```diff
@@ -0,0 +1,404 @@
...
+        let after = fetch_adjust(
+            self.offsets(),
+            new.read as isize - current.read as isize,
+            Ordering::SeqCst,
+        );
+
+        // 8. If the queue was full in step 1, signal the writer to wake up.
+        //    ... or became full during read
+        let became_full = Offsets::new(after, self.data_len() as u32).is_full();
+        Ok((val, was_full || became_full))
```
jethrogb, line 216: Ok I now think this makes sense. But I also think `became_full` is the only thing that should be checked. Please update the abi spec as well (“If the queue was full before step 7").

ipc-queue/src/sealed.rs
```diff
@@ -0,0 +1,103 @@
...
+use std::os::fortanix_sgx::usercalls::alloc::UserSafeSized;
+
+#[cfg(not(target_env = "sgx"))]
+pub trait UserSafeSized: Copy + Sized {}
+
+#[cfg(not(target_env = "sgx"))]
+impl<T> UserSafeSized for T where T: Copy + Sized {}
+
+#[repr(C)]
+pub struct WithId<T: Copy> {
```
jethrogb, line 13: Actually I meant for this to go in `fortanix-sgx-abi`. Then you can also get rid of most of this module.

ipc-queue/src/fifo.rs
```diff
@@ -0,0 +1,404 @@
...
+    fn clone(&self) -> Self {
+        match self {
+            Storage::Shared(arc) => Storage::Shared(arc.clone()),
+            Storage::Static => Storage::Static,
+        }
+    }
+}
+
+pub(crate) struct FifoInner<T> {
+    data: NonNull<UnsafeCell<[T]>>,
```
jethrogb, line 74: Maybe an `[UnsafeCell<T>]` makes more sense. What do you think?

ipc-queue/src/interface_sync.rs
```diff
@@ -0,0 +1,347 @@
...
+    ///
+    /// * The `data` and `len` fields in `FifoDescriptor` must adhere to all
+    ///   safety requirements described in `std::slice::from_raw_parts_mut()`
+    ///
+    /// * The `offsets` field in `FifoDescriptor` must be non-null and point
+    ///   to a valid memory location holding an `AtomicUsize`.
+    ///
+    /// * The synchronizer must somehow know how to correctly synchronize with
+    ///   the other end of the channel.
+    pub unsafe fn from_descriptor(d: FifoDescriptor<T>, synchronizer: S) -> Self {
```
jethrogb, line 38: Continuing the discussion from https://github.com/fortanix/rust-sgx/pull/246#discussion_r490111910 I still don't see the SGX safe user memory types being used here.

jethrogb (commented):
github now requires me to put some text here

ipc-queue/src/fifo.rs
mzohreva: Right, `was_full` implies `became_full`

ipc-queue/src/sealed.rs
mzohreva: I'm not sure what you intend here, can you elaborate more? Do you just want the `WithId` and `Identified` to be in the abi crate? or do you want a different restructuring?

ipc-queue/src/fifo.rs
mzohreva: Not sure if that makes any difference. Do you think `UnsafeCell<[T]>` could imply anything beyond the interior mutability of slice elements?

ipc-queue/src/interface_sync.rs
mzohreva: I initially assumed you only meant the traits (`UserSafe`, `UserSafeSized`), but yeah we could also use `User<[T]>` here to check the pointer and len.

ipc-queue/src/fifo.rs
jethrogb: With `[UnsafeCell<T>]` you can index/get the length without going through the unsafe wrapper. With `UnsafeCell<[T]>` you can (unsafely) change the pointer/length. Both of these points make me think `[UnsafeCell<T>]` is better.

ipc-queue/src/sealed.rs
jethrogb: `WithId` in the abi crate, I don't think you need the `Identified` trait.

mzohreva: What bound do you put on T for Sender/Receiver types then?

jethrogb: `Copy` (or `UserSafeSized` in the enclave)

```
pub(crate) struct FifoInner<T> {
    data: NonNull<[UnsafeCell<WithId<T>>]>,
    offsets: NonNull<AtomicUsize>,
    storage: Storage<T>,
}

```

mzohreva: This would mean that send/recv functions would be changed to take/return id explicitly. Also ultimately we need to transmute `FifoDescriptor<Usercall>` to `FifoDescriptor<WithId<(u64, u64, u64, u64, u64)>>` (or the reverse) somewhere. That does not seem like a better interface to me.

enclave-runner/src/usercalls/mod.rs
```diff
@@ -1341,26 +1445,17 @@ impl<'tcs> IOHandlerInput<'tcs> {
...
-                .tcs
-                .pending_events
-                .iter()
-                .position(|ev| (ev & event_mask) != 0)
-            {
-                ret = self.tcs.pending_events.remove(pos);
-                self.tcs.pending_event_set = self.tcs.pending_events.iter().fold(0, |m, ev| m | ev);
-            }
-        }
+        // TODO: the ABI allows for calling wait() asynchronously with specific semantics
```
jethrogb, line 1448: Please fix or file a GH issue and link to it

jethrogb (commented):
Note to self: continue review at `struct QueueSynchronizer`

enclave-runner/src/usercalls/mod.rs
```diff
@@ -693,61 +817,44 @@ impl EnclaveState {
...
+            };
+
+            *enclave_clone.fifo_guards.lock().await = Some(fifo_guards);
+            *enclave_clone.return_queue_tx.lock().await = Some(return_queue_tx);
+
+            tokio::task::spawn_local(async move {
+                while let Ok(usercall) = usercall_queue_rx.recv().await {
+                    let _ = io_queue_send.send(UsercallSendData::Async(usercall));
+                }
+            });
```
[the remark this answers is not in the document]
mzohreva: That's possible, but it would complicate the already too complicated loop below. What's your main concern with having it as a separate task?

enclave-runner/src/usercalls/mod.rs
```diff
@@ -491,14 +503,51 @@ struct StoppedTcs {
...
 struct IOHandlerInput<'tcs> {
-    tcs: &'tcs mut RunningTcs,
+    tcs: Option<&'tcs mut RunningTcs>,
     enclave: Arc<EnclaveState>,
     work_sender: &'tcs crossbeam::crossbeam_channel::Sender<Work>,
 }
 
+struct PendingEvents {
+    counts: [u32; Self::MAX_EVENT],
+}
```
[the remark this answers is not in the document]
mzohreva: The impetus for this type is that some threads may never be interested in a particular event, and therefore we could be using unbounded amount of memory storing those events. For example most enclave threads are not interested in `EV_RETURNQ_NOT_EMPTY` since the return queue is probably owned by just one thread, but the enclave runner needs to send that event to all threads since it does not know which thread owns that queue.

enclave-runner/src/usercalls/mod.rs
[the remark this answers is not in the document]
jethrogb: Primarily because the current flow is obfuscating what the input is to the while loop below and a lesser concern might be performance.

enclave-runner/src/usercalls/mod.rs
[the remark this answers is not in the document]
jethrogb: > The impetus for this type is that some threads may never be interested in a particular event, and therefore we could be using unbounded amount of memory storing those events.

Makes sense.

> the enclave runner needs to send [EV_RETURNQ_NOT_EMPTY] to all threads since it does not know which thread owns that queue

Well it only needs to be sent to all threads if no thread is currently waiting on that specific event:

> Userspace will wake any or all threads *waiting on the appropriate event* when it is triggered.

(emphasis mine)

enclave-runner/src/usercalls/mod.rs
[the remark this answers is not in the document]
mzohreva: The input to that while loop comes from `fn do_work` and this task, I don't see that being obfuscated, specially because `UsercallSendData` has two variants. I don't think performance will be impacted much by having an extra task considering that doing `select` will also introduce an extra future.

We could also implement some form of back pressure on the usercall queue more easily in this form, but that might not be needed in practice.

enclave-runner/src/usercalls/mod.rs
```diff
@@ -1466,11 +1584,111 @@ impl<'tcs> IOHandlerInput<'tcs> {
...
+    subscription_maker: async_pubsub::Sender<()>,
+}
+
+impl QueueSynchronizer {
+    fn new(enclave: Arc<EnclaveState>) -> (Self, Self, async_pubsub::Sender<()>) {
+        // This broadcast channel is used to notify enclave-runner of any
+        // synchronous usercalls made by the enclave for the purpose of
+        // synchronizing access to usercall and return queues.
+        // The size of this channel should not matter since recv() can
+        // return RecvError::Lagged.
```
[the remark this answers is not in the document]
mzohreva: It can't be 0 (will panic) but no reason to just set it to 1 I guess.

enclave-runner/src/usercalls/mod.rs
```diff
@@ -8,33 +8,37 @@ extern crate libc;
...
-use fnv::FnvHashMap;
-
-use futures::StreamExt;
-use futures::lock::Mutex;
-use futures::future::{Either, FutureExt, Future, poll_fn};
+use std::{cmp, fmt, str};
 use tokio::prelude::{AsyncRead, AsyncWrite};
-use tokio::sync::mpsc as async_mpsc;
 use tokio::stream::Stream as TokioStream;
+use tokio::sync::broadcast as async_pubsub;
```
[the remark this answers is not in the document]
mzohreva: For some reason that clicked better for me when I was developing it, but you are right, that might just confuse readers.

jethrogb (approved):
Approving this but not yet merging as I want to make sure we publish our existing updates first.

Issue #136, opened 2019-04-25
Allow passing args to Command


jethrogb:
The ABI will let you pass arguments to `main_entry`, but neither `enclave_runner::Command` nor `ftxsgx-runner` will let you do this at the moment.

When this feature is added, it should be clearly documented that this is not an appropriate channel for passing secrets or security configuration.

kaimast:
[This site](https://edp.fortanix.com/docs/concepts/rust-std/) says _"Configuration data that isn't relevant to security may be provided as a command-line argument, read from stdin, or appended as unmeasured data."_ 
But there is indeed no way to pass arguments to the enclave right now. 

I was hoping to run multiple enclaves on the same machine and pass which port they are listening on as a command line argument, but this seems impossible right now?

kaimast:
Any ideas how to work around this? 

If not, would patches for the `ftxsgx-runner` to add this feature be welcome?

jethrogb:
PRs never expected, always appreciated :)

PR #251, opened 2020-06-10
Pass arguments to executable enclaves

mzohreva:
This should resolve issue #136

### DIFF
--- a/enclave-runner/src/loader.rs
+++ b/enclave-runner/src/loader.rs
@@ -67,7 +67,7 @@ pub struct EnclaveBuilder<'a> {
     load_and_sign: Option<Box<dyn FnOnce(Signer) -> Result<Sigstruct, Error>>>,
     hash_enclave: Option<Box<dyn FnOnce(&mut EnclaveSource<'_>) -> Result<EnclaveHash, Error>>>,
     forward_panics: bool,
-    cmd_args: Option<Vec<String>>,
+    cmd_args: Option<Vec<Vec<u8>>>,
 }
 
 #[derive(Debug, Fail)]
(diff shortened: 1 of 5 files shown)


Goirad:
To reflect our out of band discussion here, please add a doc comment that most applications expect the first (0th) argument to be the name of the executable, and care should be taken by whoever calls `run_with_args`

fortanix-sgx-tools/src/bin/ftxsgx-runner-cargo.rs
```diff
@@ -98,6 +98,10 @@ fn run() -> Result<(), Error> {
 
     let mut ftxsgx_runner_command = Command::new(runner);
     ftxsgx_runner_command.arg(args[1].clone() + ".sgxs");
+    if args.len() > 2 {
+        ftxsgx_runner_command.arg("--");
+        ftxsgx_runner_command.args(&args[2..]);
```
Goirad, line 103: What happens if someone runs `runner <enclave> --`?

jethrogb:
As discussed please move this to a builder-like interface similar to `std::process::Command`

fortanix-sgx-tools/src/bin/ftxsgx-runner-cargo.rs
jethrogb: Then `&args[2..]` returns the empty slice so I think this is fine

Goirad: Also when run with cargo, as is the intention, what does this look like?
`cargo run -- -- <enclave arg>`? IE do we need one `--` to get arguments to the runner, and another `--` to get arguments to the enclave?

mzohreva: Just one `--` would be enough in that case. That's actually why I added `--` in the code here so you don't have to repeat it twice.

kaimast:
Any chance this is getting merged soon?

enclave-runner/src/loader.rs
```diff
@@ -260,6 +262,21 @@ impl<'a> EnclaveBuilder<'a> {
...
+    /// **NOTE:** This is not an appropriate channel for passing secrets or
+    /// security configurations to the enclave.
+    ///
+    /// **NOTE:** This is only applicable to [`Command`] enclaves.
+    /// Setting command arguments and then calling [`build_library`] will cause
+    /// a panic.
+    ///
+    /// [`Command`]: struct.Command.html
+    /// [`build_library`]: struct.EnclaveBuilder.html#method.build_library
+    pub fn command_arguments(&mut self, cmd_args: Vec<String>) -> &mut Self {
```
jethrogb, line 275: See https://doc.rust-lang.org/std/process/struct.Command.html#method.args

And arguments should be of type `Vec<u8>`, not `String`

jethrogb:
Public API LGTM now, can someone else review the rest?

Goirad:
I'm on it

fortanix-sgx-tools/src/bin/ftxsgx-runner.rs
```diff
@@ -59,6 +64,13 @@ fn main() -> Result<(), Error> {
         None => (),
     }
 
+    enclave_builder.arg(file);
+    if let Some(enclave_args) = args.values_of("enclave-args") {
+        for a in enclave_args {
```
Goirad, line 69: Can't you use the `args` method here after mapping `enclave_args`?
`args.values_of("enclave-args").map(|args| enclave_builders.args(args))`

mzohreva: I'm not a fan of writing things with side effect is a functional style, but I can definitely use the `args` method instead of the for loop.

PR #255, opened 2020-06-24
Fix doc generation

parthsane:
replaced rust-std-edp-doc.patch with a newer version

The doc generation failed because the patch could not be applied.


jethrogb:
bors retry

PR #258, opened 2020-06-24
disable windows build

AdrianCX:


jethrogb:
bors r+

@parthsane please revert this when you fix Windows

PR #259, opened 2020-06-24
Introduce EM certificate library and SGX-PKIX

AdrianCX:
Commit introduces em-app certificate library and sgx-pkix


AdrianCX:
bors r=aandyl,pagten

PR #262, opened 2020-06-24
Rename sgx_pkix to sgx-pkix

AdrianCX:


Goirad:
Do you plan to republish sgx-pkix after this pr and not before?

PR #264, opened 2020-07-03
Add support to write EFI variables to the environment

shreevari:
I plan to work on enabling SGX via software with `sgx-detect` interactively. It requires that I update the EPCSW EFI variable, hence this change.

I also need some help with the default attributes for the EFI variables.


shreevari:
I'm having issues writing EFI varibles to the environment. There's an OS error (Error number 5 - EIO).
This [thread](https://unix.stackexchange.com/questions/152144/how-to-write-edit-update-the-osindications-efi-variable-from-command-line) suggests that I remove the immutable attribute and write out the file, but I'm still getting the same error.
I am unable to set `e` attribute on the file through command line too.

Any help is welcome

Taowyoo:
Closed this since long time no update. Please be free to reopen this if needed.

PR #268, opened 2020-07-12
Print specific error message on SIGBUS

jethrogb:


fortanix-sgx-tools/src/bin/ftxsgx-runner.rs
```diff
@@ -30,6 +30,19 @@ arg_enum!{
     }
 }
 
+#[cfg(unix)]
+fn catch_sigbus() {
+    unsafe {
+        extern "C" fn handle_bus(_signo: c_int, _info: *mut siginfo_t, _context: *mut c_void) {
+            eprintln!("SIGBUS triggered: likely caused by stack overflow in enclave.")
```
AdrianCX, line 37: This might crash itself as println inside signal handlers is not a good idea...


Also, since we're already crashing there's the risk that stderr is not flushed and you won't see message as well. (can we fflush on stderr?)

jethrogb: We're about to crash due to SIGBUS so we don't really have another option. I added the flush.

Issue #233, opened 2020-05-01
Remove dependency on protobuf on Windows


jethrogb:
aesm-client doesn't use protobufs on Windows, yet it still requires the protobuf compiler to build

Issue #144, opened 2019-05-22
Setup Windows CI


jethrogb:
With #89 landing, we should really be setting CI for Windows. Travis supposedly supports this now, and it looks like you might be able to install OpenSSL with chocolatey.

PR #271, opened 2020-07-18
Remove dependency on protobuf on Windows

kgauravkumar:
This will address issue #233


parthsane:
This looks good. Unfortunately we don't have the Windows CI running currently. So I think we should get Windows CI working again before we can merge this #144

PR #280, opened 2020-08-26
Don't pass filename to enclave as first arg

mzohreva:


fortanix-sgx-tools/src/bin/ftxsgx-runner.rs
```diff
@@ -81,7 +81,7 @@ fn main() -> Result<(), Error> {
         None => (),
     }
 
-    enclave_builder.arg(file);
+    enclave_builder.arg("enclave");
```
jethrogb, line 84: When you launch a process through `std::process::Command`, you don't get to specify arg 0. Should be the same here, this should be handled in `enclave-runner`.

enclave-runner/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "enclave-runner"
-version = "0.4.1"
+version = "0.4.2"
```
mzohreva, line 3: Is this enough, or do we need a minor version bump?

enclave-runner/src/loader.rs
```diff
@@ -262,6 +262,13 @@ impl<'a> EnclaveBuilder<'a> {
         self
     }
 
+    fn initialized_args_mut(&mut self) -> &mut Vec<Vec<u8>> {
+        if let None = self.cmd_args {
+            self.cmd_args = Some(vec!["enclave".as_bytes().to_vec()]);
+        }
+        self.cmd_args.as_mut().unwrap()
```
jethrogb, line 269:
```rust
self.cmd_args.get_or_insert_with(|| vec![b"enclave".to_vec()])
```

enclave-runner/Cargo.toml
jethrogb: I guess this is technically a breaking change, yes.

fortanix-sgx-tools/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "fortanix-sgx-tools"
-version = "0.4.3"
+version = "0.5.0"
```
jethrogb, line 3: This can be a patch bump

mzohreva: The behavior is changed though. It now doesn't pass the filename as first arg

enclave-runner/src/loader.rs
```diff
@@ -263,6 +263,10 @@ impl<'a> EnclaveBuilder<'a> {
         self
     }
 
+    fn initialized_args_mut(&mut self) -> &mut Vec<Vec<u8>> {
```
TorfsD, line 266: @mzohreva  We came to this commit after investigating some troubles related to command arguments parsing. It seems the `initialized_args_mut()` would be only called when there are actually arguments being passed (when `arg()` or `args()` is called). So in case no arguments are passed, the `cmd_args` would still remain `None`, `enclave` would not be inserted and the `args` of the enclave process as read by its `main` would be empty. Is this intentional ?

mzohreva: I think this is a bug, I should have changed `EnclaveBuilder::build()` to call `initialized_args_mut`. I'll create a tracking issue for it.

PR #283, opened 2020-09-14
Support different API families for the Linux driver loader

jethrogb:


sgxs-loaders/tests/test_linux_drivers.sh
```diff
@@ -0,0 +1,100 @@
+#!/bin/bash -e
+#
+# Prior to running this script, make sure the system has the following
+# configuration:
+#
+# * Linux: 4.15 generic kernel (e.g. not the Azure kernel)
+# * Test dependencies: build-essential dkms docker.io
+# * EDP dependencies: pkg-config libssl-dev protobuf-compiler
+# * Rust: latest nightly installed with x86_64-fortanix-unknown-sgx target
+# * No other SGX drivers or software installed (from any vendor)
```
jethrogb, line 10: Build log on a properly prepared Azure VM: [test-output.txt](https://github.com/fortanix/rust-sgx/files/5218723/test-output.txt)

sgxs-loaders/tests/test_linux_drivers.sh
```diff
@@ -0,0 +1,100 @@
...
+#
+# Note: you can't run this script multiple times because the FLC drivers make
+# CPU configuration changes that the non-FLC drivers don't support.
```
raoulstrackx, line 13: How do you recover from running this test once? Is rebooting the only option?

sgxs-loaders/src/isgx/mod.rs
```diff
@@ -132,13 +207,74 @@ impl EnclaveLoad for InnerDevice {
...
+                    srcpage: &data,
+                    secinfo: &secinfo,
+                    chunks: chunks.0,
+                };
+                ioctl_unsafe!(Add, ioctl::montgomery::add(mapping.device.fd.as_raw_fd(), &adddata))
+            },
+            Augusta => {
+                let flags = match chunks.0 {
+                    0 => ioctl::augusta::SgxPageFlags::empty(),
+                    0xffff => ioctl::augusta::SgxPageFlags::SGX_PAGE_MEASURE,
```
raoulstrackx, line 224: Where do these constants come from?

sgxs-loaders/tests/test_linux_drivers.sh
jethrogb: Rebooting is easiest. You could manually restore the PUBKEYHASH MSRs but I didn't invest time in figuring out how.

sgxs-loaders/src/isgx/mod.rs
jethrogb: It's just “no bits set” and “all bits set” meaning all or no 256-byte chunks in a page, see also the Windows loader.

raoulstrackx (approved):

PR #286, opened 2020-09-18
Cancel queue

mzohreva:
This is on top of PR #246 and adds a third queue for cancelling blocked usercalls.

### DIFF
--- a/fortanix-sgx-abi/src/lib.rs
+++ b/fortanix-sgx-abi/src/lib.rs
@@ -629,9 +629,13 @@ impl Usercalls {
 /// Additionally, userspace may choose to ignore cancellations for non-blocking
 /// usercalls. Userspace should be able to cancel a usercall that has been sent
 /// by the enclave but not yet received by the userspace, i.e. if cancellation
-/// is received before the usercall itself. However, userspace should not keep
-/// cancellations forever since that would prevent the enclave from re-using
-/// usercall ids.
+/// is received before the usercall itself. To avoid keeping such cancellations
+/// forever and preventing the enclave from re-using usercall ids, userspace
+/// should synchronize cancel queue with the usercall queue such that the
+/// following invariant is maintained: whenever the enclave writes an id to the
+/// usercall or cancel queue, the enclave will not reuse that id until the
+/// usercall queue's read pointer has advanced to the write pointer at the time
+/// the id was written.
 ///
 /// *TODO*: Add diagram.
 ///
@@ -721,8 +725,8 @@ pub mod async {
     #[derive(Copy, Clone, Default)]
     #[cfg_attr(feature = "rustc-dep-of-std", unstable(feature = "sgx_platform", issue = "56975"))]
     pub struct Cancel {
-        /// This must be the same value as `Usercall.0`.
-        pub usercall_nr: u64,
+        /// Reserved for future use.
+        pub reserved: u64,
     }
 
     /// A circular buffer used as a FIFO queue with atomic reads and writes.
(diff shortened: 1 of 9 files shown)


fortanix-sgx-abi/src/lib.rs
```diff
@@ -611,18 +615,36 @@ impl Usercalls {
...
+/// cancel the usercall's execution and return an [`Interrupted`] error on the
+/// return queue to notify the enclave of the cancellation. Note that usercalls
+/// that do not return [`Result`] cannot be cancelled and if the enclave sends
+/// a cancellation for such a usercall, the userspace should simply ignore it.
+/// Additionally, userspace may choose to ignore cancellations for non-blocking
+/// usercalls. Userspace should be able to cancel a usercall that has been sent
+/// by the enclave but not yet received by the userspace, i.e. if cancellation
+/// is received before the usercall itself. However, userspace should not keep
+/// cancellations forever since that would prevent the enclave from re-using
+/// usercall ids.
```
jethrogb, line 631: This should be more concrete

fortanix-sgx-abi/src/lib.rs
```diff
@@ -453,6 +453,10 @@ pub const EV_RETURNQ_NOT_EMPTY: u64 = 0b0000_0000_0000_0010;
 /// An event that enclaves can use for synchronization.
 #[cfg_attr(feature = "rustc-dep-of-std", unstable(feature = "sgx_platform", issue = "56975"))]
 pub const EV_UNPARK: u64 = 0b0000_0000_0000_0100;
+/// An event that will be triggered by userspace when the cancel queue is not
+/// or no longer full.
```
jethrogb, line 457: I didn't check the exact language when looking at the other queue. Are we correctly handling the “not full” case? i.e. this phrasing suggests that if the enclave calls `wait` on a queue event and there is no event pending but the queue is not empty/full, an event should be generated immediately.

fortanix-sgx-abi/src/lib.rs
jethrogb: For example, you can synchronize the cancel queue with the usercall queue to maintain the following invariant: Whenever the enclave writes an id to the usercall or cancel queue, the enclave will not reuse that id until the usercall queue read pointer has advanced to the write pointer at the time the id was written. The corollary of this is that userspace can forget about an “unused” cancel id once the read pointer has advanced to a certain point. It may be possible to relax the invariant a bit to get less strict synchronization requirements but still have a decent bound on id liveness.

fortanix-sgx-abi/src/lib.rs
mzohreva: I don't think we do that in the implementation. It's based on the assumption that the enclave would only issue a `wait` usercall with `NOT_EMPTY/NOT_FULL` only after trying to send/recv on a queue. But I don't think it would be too difficult to change this. We'd need some way of getting the queue's empty/full state in `fn wait`.

fortanix-sgx-abi/src/lib.rs
mzohreva: That's an interesting idea. Implementing it would require some kind of strictly increasing read/write pointers that could be compared to each other. Here is one way to do it:
- Store an additional atomic "epoch" counter in the Receiver that records how many times the write offset of the ring buffer has wrapped around,
- Combine that value with the current read/write offset (modulo queue length) to form a read/write pointer.

To avoid problems when the produced numbers wrap around, we should make the "epoch" counter big enough and 64 bits seems plenty.

Can you think of a simpler solution?

fortanix-sgx-abi/src/lib.rs
```diff
@@ -690,6 +713,15 @@ pub mod async {
         }
     }
 
+    /// Cancel a usercall peviously sent to userspace.
+    #[repr(C)]
+    #[derive(Copy, Clone, Default)]
+    #[cfg_attr(feature = "rustc-dep-of-std", unstable(feature = "sgx_platform", issue = "56975"))]
+    pub struct Cancel {
+        /// Reserved for future use.
+        pub reserved: u64,
```
jethrogb, line 722: Do you have a use in mind?

fortanix-sgx-abi/src/lib.rs
jethrogb: Can you update the documentation?

fortanix-sgx-abi/src/lib.rs
mzohreva: The timeout idea we discussed in person.

Taowyoo:
Close this since this is done by https://github.com/fortanix/rust-sgx/pull/515

Issue #200, opened 2019-11-27
sgx-detect: Check KSS support


jethrogb:
`sgx-detect` currently doesn't interpret KSS support from the supported attributes in CPUID. See the TODOs in `tests/mod.rs`.

jethrogb:
Related: #277

PR #289, opened 2020-09-29
Add KSS detection to sgx-detect

raoulstrackx:
Adds support to `sgx-detect` to determine if the platform supports KSS (ref #200)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2061,7 +2061,7 @@ dependencies = [
 
 [[package]]
 name = "sgx-isa"
-version = "0.3.2"
+version = "0.3.3"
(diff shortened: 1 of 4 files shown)


jethrogb:
Have you tested this on KSS and non-KSS systems?

raoulstrackx:
Yes I did

jethrogb:
Please bump sgx-isa patch version.

raoulstrackx:
What's the policy on application version? Shouldn't the version of sgx-detect be bumped as well?

jethrogb:
Normally, yes, but that's being done in #287

PR #292, opened 2020-10-13
Support for enclave dynamic memory

raoulstrackx:
SGXv2 supports adding and removing memory pages after the enclave has been constructed. This PR implements the required infrastructure for this.

cc: @jethrogb


Cargo.toml
```diff
@@ -17,3 +17,6 @@ members = [
     "sgxs-tools"
 ]
 exclude = ["examples"]
+
+[patch.crates-io]
+bitflags = { git = "https://github.com/bitflags/bitflags" }
```
raoulstrackx, line 22: Can be removed once `bitflags` publishes a new version

raoulstrackx:
We should add infrastructure to run this in CI as well.

enclave-runner/src/loader.rs
```diff
@@ -22,6 +23,7 @@ use openssl::{
 use sgx_isa::{Attributes, AttributesFlags, Miscselect, Sigstruct};
 use sgxs::crypto::{SgxHashOps, SgxRsaOps};
 use sgxs::loader::{Load, MappingInfo, Tcs};
+use sgxs::loader::{EnclaveControl};
```
raoulstrackx, line 26: Remove '{' and '}'

PR #293, opened 2020-10-13
Augusta SGX loader: map contiguous similar pages at once

jethrogb:
Avoid running into the `/proc/sys/vm/max_map_count` limit by creating fewer larger mappings instead of mapping every page.

This unfortunately requires delaying the last `mmap` call to `einit` time, which is not great UX. This is the same as what Intel does in their SDK though.


jethrogb:
bors r=raoulstrackx

PR #299, opened 2020-11-04
Limit memory usage for broadcast events

mzohreva:
The elusive memory leak that I was chasing was not an actual leak, but was rather caused by pile up of events broadcast to all TCS's. This PR fixes the issue by removing the unbounded channel used for events.

### DIFF
--- a/enclave-runner/src/usercalls/mod.rs
+++ b/enclave-runner/src/usercalls/mod.rs
@@ -14,7 +14,7 @@ use std::sync::atomic::{AtomicBool, AtomicUsize, Ordering};
 use std::sync::Arc;
 use std::task::{Context, Poll, Waker};
 use std::thread::{self, JoinHandle};
-use std::time::{self, Duration, Instant};
+use std::time::{self, Duration};
 use std::{cmp, fmt, str};
 
 use failure::{self, bail};
@@ -31,6 +31,7 @@ use tokio::io::{AsyncRead, AsyncWrite};
 use tokio::stream::Stream as TokioStream;
 use tokio::sync::broadcast;
 use tokio::sync::mpsc as async_mpsc;
+use tokio::sync::Semaphore;
 
 use fortanix_sgx_abi::*;
 use ipc_queue::{self, DescriptorGuard, Identified, QueueEvent};
@@ -496,7 +497,6 @@ impl fmt::Pointer for TcsAddress {
 
 struct StoppedTcs {
     tcs: ErasedTcs,
-    event_queue: futures::channel::mpsc::UnboundedReceiver<u8>,
 }
 
 struct IOHandlerInput<'tcs> {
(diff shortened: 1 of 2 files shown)


enclave-runner/src/usercalls/mod.rs
```diff
@@ -505,8 +505,12 @@ struct IOHandlerInput<'tcs> {
     work_sender: &'tcs crossbeam::crossbeam_channel::Sender<Work>,
 }
 
-struct PendingEvents {
-    counts: [u32; Self::EV_MAX],
+struct PendingEvents(spin::Mutex<PendingEventsInner>);
```
jethrogb, line 508: Instead, can you just use an async semaphore?

enclave-runner/src/usercalls/mod.rs
```diff
@@ -1468,51 +1505,25 @@ impl<'tcs> IOHandlerInput<'tcs> {
...
-                    return Err(IoErrorKind::Other.into());
-                }
+        let ret = match timeout {
+            WAIT_NO => pending_events.take(event_mask),
+            WAIT_INDEFINITE => Some(pending_events.wait_for(event_mask).await),
+            n => match tokio::time::timeout(Duration::from_nanos(n), pending_events.wait_for(event_mask)).await {
+                Ok(ev) => Some(ev),
+                Err(_) => None, // timed out
+            },
+        };
```
jethrogb, line 1517: This got much cleaner now, awesome.

Goirad: Is this match not the same as `.ok()`?

enclave-runner/src/usercalls/mod.rs
```diff
@@ -506,7 +506,8 @@ struct IOHandlerInput<'tcs> {
 }
 
 struct PendingEvents {
-    counts: [u32; Self::EV_MAX],
+    counts: [Semaphore; Self::EV_MAX],
```
Goirad, line 509: Can you add a comment here briefly explaining the meaning of this field? I don't think an array of semaphores is a very common construct (which is to say I have never seen it before)

enclave-runner/src/usercalls/mod.rs
```diff
@@ -506,7 +506,8 @@ struct IOHandlerInput<'tcs> {
...
+    abort: Semaphore,
```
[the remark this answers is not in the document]
mzohreva: Good point. I'd say Semaphore with large number of permits is a practical solution although not the most elegant.

enclave-runner/src/usercalls/mod.rs
mzohreva: It's effectively the same thing as before (a bunch of counters for all possible event sets) but with the added benefit of being able to wait until a counter is non zero (through `Semaphore::acquire`). Although you may be right that this could be confusing for readers, so I'll add a comment.

enclave-runner/src/usercalls/mod.rs
```diff
@@ -520,34 +524,83 @@ impl PendingEvents {
...
+        // [the documentation]: https://docs.rs/tokio/0.2.22/tokio/sync/struct.Semaphore.html#method.add_permits
+        const MAX_PERMITS: usize = usize::MAX >> 5;
+        if self.counts[index].available_permits() < MAX_PERMITS {
+            self.counts[index].add_permits(1);
         }
     }
+
+    fn abort(&self) {
+        // add a "large" number of permits without panicing if `abort` is called
+        // "multiple" times.
```
jethrogb, line 597: You can just check if there are more than 0 permits before adding.

jethrogb (commented):
There's a bunch of `event_number as u8` in the code that is no longer necessary because the u8 channels are gone. Can you go through and see if the native type can be used in relevant places?

PR #302, opened 2020-12-01
Mark ByteBuffer as Send

mzohreva:


jethrogb:
Do we need a version bump? Or do we have one pending?

mzohreva:
Latest published version of `fortanix-sgx-abi` is 0.3.3 but Cargo.toml has `version = "0.4.0"`, so we're good.

PR #306, opened 2021-01-09
add a CsrSigner trait to em-app

aandyl:
the CsrSigner trait is similar to ExternalKey, but instead of allowing signing of arbitrary messages, only signs CSRs.

a default implementation of CsrSigner is provided for anything that implements ExternalKey


AdrianCX (approved):

AdrianCX (approved):

aandyl:
bors r=adriancx

PR #315, opened 2021-02-23
`aesm-client` bug fix

raoulstrackx:
`aesm-client` could not get build with `sgx-isa/sgxstd` feature


jethrogb:
Please bump version

PR #316, opened 2021-02-23
Upgrade `protobuf` to enable publishing `aesm-client`

raoulstrackx:


aesm-client/Cargo.toml
```diff
@@ -32,7 +32,7 @@ sgx-isa = { version = "0.3.0", path = "../sgx-isa"}
 # External dependencies
 byteorder = "1.0"          # Unlicense/MIT
 lazy_static = "1"          # MIT/Apache-2.0
-protobuf = "2.8.0"         # MIT/Apache-2.0
+protobuf = "2.22.0"        # MIT/Apache-2.0
```
jethrogb, line 35: Why are you updating the version requirement? AFAICT Message::parse_from_bytes already existed in 2.8.

raoulstrackx: I don't see it: https://docs.rs/protobuf/2.18.1/protobuf/fn.parse_from_bytes.html Am I missing something?

jethrogb: Sorry, you're right, I didn't look correctly. The minimum version should be 2.19.0.

PR #320, opened 2021-03-24
rs-libc: Fix build determinism

exFalso:
Currently the directory listing order in rs-libc's build.rs is non-deterministic, and this order is fed into the `cc` build directly, making the final binary artifact non-reproducible. This is because link-time relocation is dependent on the order in which inputs are specified.

In certain use-cases of SGX the attested measurement must be reproducible by third parties, otherwise a security audit of the enclave code cannot be tied to attestations.

The fix is to simply sort the listing.

### DIFF
--- a/rs-libc/build.rs
+++ b/rs-libc/build.rs
@@ -34,14 +34,14 @@ fn main() {
     let mut build = cc::Build::new();
 
     #[cfg(unix)]
-    for path in read_dir(p_s).unwrap().filter_map(extension_filter("S")) {
+    for path in sorted(read_dir(p_s).unwrap().filter_map(extension_filter("S"))) {
         build.file(path);
     }
     #[cfg(windows)]
-    for path in read_dir(p_s).unwrap().filter_map(extension_filter("o")) {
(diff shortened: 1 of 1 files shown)


rs-libc/build.rs
```diff
@@ -71,3 +71,9 @@ fn main() {
 
     b.warnings(false).compile(name);
 }
+
+fn sorted<A: Ord, I: Iterator<Item=A>>(iterator: I) -> Vec<A> {
+    let mut items: Vec<_> = iterator.collect();
+    items.sort();
+    items
```
jethrogb, line 78: You can sort while inserting:
```suggestion
fn sorted<A: Ord, I: Iterator<Item=A>>(iterator: I) -> BTreeSet<A> {
    iterator.collect()
```

jethrogb (changes requested):
Thanks for the PR! Reproducible builds are indeed important. Just one minor code comment.

jethrogb:
Probably need to import the type as well.

exFalso:
woops first time I used the GH commit thing, got a bit too excited!

PR #321, opened 2021-03-25
Use remote instead of local timeout for init_quote() in aesm-client

### LINKED ISSUE #179: AesmClient::init_quote() returns "Resource temporarily unavailable" on first call
When aesmd is still unprovisioned, `AesmClient::init_quote()` triggers EPID provisioning, and in this case the client returns an `AesmCommunication` error. Subsequent calls return fine.

Smallest reproducing test case is to call `AesmClient::new().init_quote()` with an unprovisioned aesmd running, e.g. with fresh docker instance.

Workaround is to simply retry.

Pagten:
This should fix #179.

Quoting from that issue:
> When aesmd is still unprovisioned, AesmClient::init_quote() triggers EPID provisioning, and in this case the client returns an AesmCommunication error.

### DIFF
--- a/aesm-client/Cargo.toml
+++ b/aesm-client/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "aesm-client"
-version = "0.5.2"
+version = "0.5.3"
 authors = ["Fortanix, Inc."]
 license = "MPL-2.0"
 description = """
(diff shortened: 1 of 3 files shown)


jethrogb:
This makes every request “remote” can you check the code comments and see if this needs updating?

PR #327, opened 2021-04-01
sgxs-sign support offline signing

ravenac95:
This PR adds support to `sgxs-sign` for an offline signing process for enclaves. This should allow for things like HSMs to be part of the signing ceremony for enclaves created with this sdk. 

Notes:
* The cli was updated in such a way that it should maintain backwards compatibility. 
* `sgxs-sign` now has subcommands that mirror those available with intel's sgx sdk. These subcommands are:
    * `sign` - This is the same as the original sgxs-sign behavior 
    * `gendata` - Generates a hash of the sigstruct data to be signed
    * `catsig` - Adds a signature to a sigstruct
* A new subcommand command `verify` validates a SIGSTRUCT against a given public key (the enclave hash is not validated but is output to stdout).


jethrogb:
Thanks for the PR! I'd certainly like to support signing with key that's not just stored on disk. However, I'm not sure if the proposed flow is the right UX. I'm thinking more along the lines of different implementations of [sgxs::crypto::SgxRsaOps](https://edp.fortanix.com/docs/api/sgxs/crypto/trait.SgxRsaOps.html), for example PKCS#11 or KMIP. Would that work for you?

ravenac95:
@jethrogb Thanks for the quick response! Completely understand this isn't ideal UX. There was a practical reason on my side, I didn't want to install too many things on an airgapped device. That being said, I think you're right that ultimately that's a nicer UX. Would you be open to supporting both experiences?

PR #330, opened 2021-04-06
bump version of sgx-pkix

aandyl:
Forgot to do this with my previous change (https://github.com/fortanix/rust-sgx/pull/328)


Goirad:
I don't see any tags for sgx-pkix, but I imagine that was just an oversight when it was first released. Other than that, looks good to me.

bors r+

PR #334, opened 2021-05-11
rs-libc: resolve symlinks in build script

choco:
Hello, for our build system we use `nix` specifically `naersk` (https://github.com/nmattia/naersk).
During the building process dependencies are linked into a build dir, not copied.
The `rs-libc` build script explicitly checks that the dir entries are files, so symlinks get ignored.
One possible solutions for our use case would be to remove this check.

### DIFF
--- a/rs-libc/build.rs
+++ b/rs-libc/build.rs
@@ -21,12 +21,9 @@ fn main() {
 
     let extension_filter = |ext| {
         move |f: Result<DirEntry, _>| {
-            let f = f.unwrap();
-            if f.file_type().unwrap().is_file() {
-                let path = f.path();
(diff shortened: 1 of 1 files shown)


jethrogb:
Can we resolve the symlink and check if that's a file?

jethrogb (approved):
Approved, but there's a CI issue with the latest Rust nightly so I can't merge it right now.

choco:
I saw that you solved the issue with Rust nightly so I rebased to master. Would it be possible to also release a new `rs-libc` version when this is merged? Thanks!

jethrogb:
Done! Apologies for the delay.

Issue #335, opened 2021-05-11
Register `rbx` is now reserved in `asm!` macros


Goirad:
When trying to `cargo check` enclave-runner:
```
error: invalid register `rbx`: rbx is used internally by LLVM and cannot be used as an operand for inline asm
   --> enclave-runner/src/tcs.rs:190:17
    |
190 |                 inout("rbx") tcs.address() => _,
    |                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^error: invalid register `rbx`: rbx is used internally by LLVM and cannot be used as an operand for inline asm
    --> enclave-runner/src/usercalls/mod.rs:1229:22
     |
1229 |         asm!("int3", in("rbx") tcs, in("r10") debug_buf, options(nomem, nostack, att_syntax));
     |                      ^^^^^^^^^^^^^
```
This is due to this change in rust: https://github.com/rust-lang/rust/pull/84658

This issue is preventing `master` from building.

PR #336, opened 2021-06-07
Avoiding use of LLVM reserved register `rbx`

raoulstrackx:
LLVM reserves register `rbx` in inline assembly (issue: #335). This PR avoids the use of this register until [rust-lang issue 85056](https://github.com/rust-lang/rust/issues/85056) is merged

### DIFF
--- a/enclave-runner/src/tcs.rs
+++ b/enclave-runner/src/tcs.rs
@@ -184,10 +184,12 @@ pub(crate) fn coenter<T: Tcs>(
         } else {
             asm!("
                     lea 1f(%rip), %rcx // set SGX AEP
+                    xchg {0}, %rbx
 1:                  enclu
+                    xchg %rbx, {0}
                 ",
+                inout(reg) tcs.address() => _, // rbx is used internally by LLVM and cannot be used as an operand for inline asm (#84658)
                 inout("eax") Enclu::EEnter as u32 => sgx_result,
-                inout("rbx") tcs.address() => _,
                 out("rcx") _,
                 inout("rdx") p3,
                 inout("rdi") p1,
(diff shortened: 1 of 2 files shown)


enclave-runner/src/tcs.rs
```diff
@@ -184,10 +184,12 @@ pub(crate) fn coenter<T: Tcs>(
         } else {
             asm!("
                     lea 1f(%rip), %rcx // set SGX AEP
+                    xchg %r12, %rbx  // rbx is a reserved LLVM register.
 1:                  enclu
+                    xchg %rbx, %r12
                 ",
                 inout("eax") Enclu::EEnter as u32 => sgx_result,
-                inout("rbx") tcs.address() => _,
+                inout("r12") tcs.address() => _, // rbx is used internally by LLVM and cannot be used as an operand for inline asm
```
jethrogb, line 192: Can you let the compiler assign a register? Same thing below

jethrogb: Also please refer to the Rust issue in a code comment

jethrogb:
I'd approve but this is a draft

enclave-runner/src/tcs.rs
```diff
@@ -184,10 +184,12 @@ pub(crate) fn coenter<T: Tcs>(
         } else {
             asm!("
                     lea 1f(%rip), %rcx // set SGX AEP
+                    xchg {0}, %rbx
 1:                  enclu
+                    xchg %rbx, {0}
```
nodakai, line 189: Looks like `ENCLU[EENTER]` doesn't write to `rbx` so the second `xchg` can perhaps be reduced to `movq {0}, %rbx`?

https://www.felixcloutier.com/x86/eenter

I'm just trying to learn how different projects dealt with https://github.com/rust-lang/rust/pull/84658 and came across with this

jethrogb: EEXIT and AEX do write to RBX. But we were already ignoring the value, so `mov` would've been ok in this case. However, `xchg` is more “obviously correct” (and revertible) to deal with this LLVM nonsense.

PR #345, opened 2021-10-20
[PLAT-182] VME enclave runner

raoulstrackx:
Initial PR with the Fortanixvme runner and fortanix-vme-abi. It only provides support for outgoing connections. It works together with [this](https://github.com/fortanix/rust/pull/2) standard library.


Pagten:
bors d+

raoulstrackx:
bors r=Pagten

PR #349, opened 2021-11-08
Closing `TcpListeners` in runner

raoulstrackx:
When a Nitro enclave closes a `TcpListener`, sockets in the runner need to be cleaned up as well. This PR introduces the required changes to the ABI and runner.

### DIFF
--- a/fortanix-vme/fortanix-vme-runner/src/lib.rs
+++ b/fortanix-vme/fortanix-vme-runner/src/lib.rs
@@ -424,7 +424,7 @@ impl Server {
             // Close `TcpListener`
             drop(listener);
         } else {
-            // Info not found, possibly not a listener socket
+            println!("[warning] Can't close the connection as it can't be located.");
         }
         let response = Response::Closed;
         Self::log_communication(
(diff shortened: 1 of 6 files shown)


fortanix-vme/fortanix-vme-runner/src/lib.rs
```diff
@@ -413,11 +417,34 @@ impl Server {
...
     }
 
+    fn handle_request_close(self: Arc<Self>, enclave_port: u32, enclave: &mut VsockStream) -> Result<(), IoError> {
+        let cid: u32 = enclave.peer().unwrap().parse().unwrap_or(vsock::VMADDR_CID_HYPERVISOR);
+        let addr = VsockAddr::new(cid, enclave_port);
+        if let Some(listener) = self.remove_listener(&addr) {
+            // Close `TcpListener`
+            drop(listener);
+        } else {
+            // Info not found, possibly not a listener socket
```
Pagten, line 427: This is quite unexpected, right? Perhaps it would be good to print a warning here.

Pagten (approved):

fortanix-vme/fortanix-vme-runner/src/lib.rs
raoulstrackx: It'd be better to return an error message here. But returning errors is a bit of a mess at the moment. I created a separate ticket for it. Would you accept a `println` for now, and improve things when errors and logging are added?

PR #353, opened 2021-11-11
Fix doc generation

raoulstrackx:


.travis.yml
```diff
@@ -28,9 +28,8 @@ matrix:
         - RUST_BACKTRACE=1 LLVM_CONFIG_PATH=llvm-3.8-config
       before_script:
         - rustup target add x86_64-fortanix-unknown-sgx x86_64-unknown-linux-musl
-        - cd intel-sgx
       script:
-        - cargo test --verbose --locked --all --exclude sgxs-loaders && [ "$(echo $(nm -D ../target/debug/sgx-detect|grep __vdso_sgx_enter_enclave))" = "w __vdso_sgx_enter_enclave" ]
```
jethrogb, line 33: I don't understand why this change is only needed now and not in a previous PR

raoulstrackx: It's not strictly needed now. Before the scripts, the working directory is changed to `intel-sgx`. That works but would be a bit weird as we'd have to call `../doc/generate-api-docs.sh`

doc/generate-api-docs.sh
```diff
@@ -50,7 +50,8 @@ LIBS_SORTED=$(
 )
 
 for LIB in $LIBS_SORTED; do
-    cd $LIB
+    LIB_DIR=$(find . -maxdepth 2 -name ${LIB} -type d)
+    cd ${LIB_DIR}
```
jethrogb, line 54: Please revert?

raoulstrackx: This is required. Libs like `sgx-isa` have moved to `intel-sgx/sgx-isa`. Doing `cd $LIB` won't work in all cases.

PR #356, opened 2021-11-19
Storing peer address

raoulstrackx:
`TcpStream`s and `TcpListener`s need to be able to return the address of the remote party. This PR:
* Makes the runner return the peer address when a new connection is established
* Returns the full address (not just the port) when a `Bound` response is send to the enclave
* Adds a test for this functionality
* Adds a statement ensuring warnings are treated as errors

### DIFF
--- a/fortanix-vme/fortanix-vme-abi/src/lib.rs
+++ b/fortanix-vme/fortanix-vme-abi/src/lib.rs
@@ -1,3 +1,4 @@
+#![deny(warnings)]
 #![no_std]
 extern crate alloc;
 #[cfg(feature="std")]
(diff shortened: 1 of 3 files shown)


fortanix-vme/fortanix-vme-abi/src/lib.rs
```diff
@@ -67,10 +68,12 @@ pub enum Response {
...
         /// The vsock port the proxy is listening on for an incoming connection
         proxy_port: u32,
+        /// The address of the remote party
+        peer: Addr,
     },
     Bound {
-        /// The TCP port the parent VM is listening on
-        port: u16,
+        /// The local TCP address the parent VM is listening on
+        local: Addr,
```
raoulstrackx, line 76: Warning: This is a breaking change. I think this is ok as we haven't released the code yet.

Pagten (approved):

PR #357, opened 2021-11-19
Pass local addr of connections to the enclave

raoulstrackx:
When handling incoming and outgoing connections, the local address of the TCP connection as used in the runner needs to be passed to the enclave so it can hide the fact from users that the connection is being proxied.


Pagten (approved):

Pagten:
bors d+

PR #361, opened 2021-11-26
[PLAT-345] Refactoring keeping track of `TcpListeners`

raoulstrackx:
`TcpListeners` are being stored by the stdlib in a list together with the file descriptor used for them in the runner. That is suboptimal. A recent change enabled storing additional data in the `Socket`. That enables to get rid of this linked list. However, for that to work we need to reference `TcpListeners` with the port they are bound to, not their file descriptor in the runner. This PR performs the required interface changes.


Pagten (approved):

Pagten:
bors d+

raoulstrackx:
Replaced by #362 for internal bookkeeping

PR #364, opened 2021-12-07
[PLAT-396] Refactor threads

raoulstrackx:
This PR improves on two key areas:
- Information about TCP connections are stored in a hashmap, similar to how information about TCP server sockets is tracked. This will enable this information to be requested from the enclave in a subsequent PR
- The way how connections are proxied, is modified. Instead of starting another thread per TCP connection, a single proxy thread is started. It iterates over all TCP connections and exchanges message once for each connection. When no connections are currently available, the thread yields its CPU time.

### DIFF
--- a/fortanix-vme/fortanix-vme-runner/src/lib.rs
+++ b/fortanix-vme/fortanix-vme-runner/src/lib.rs
@@ -96,8 +96,12 @@ struct Connection {
     remote_name: String,
 }
 
+#[derive(Clone, Debug)]
+struct ConnectionInfo {
+}
+
 impl Connection {
-    fn new(vsock_stream: VsockStream<Std>, tcp_stream: TcpStream, remote_name: String) -> Self {
+    pub fn new(vsock_stream: VsockStream<Std>, tcp_stream: TcpStream, remote_name: String) -> Self {
         Connection {
             tcp_stream,
             vsock_stream,
@@ -105,41 +109,44 @@ impl Connection {
         }
     }
 
-    fn close(&self) {
-        let _ = self.tcp_stream.shutdown(Shutdown::Both);
-        let _ = self.vsock_stream.shutdown(Shutdown::Both);
+    pub fn info(&self) -> ConnectionInfo {
+        ConnectionInfo{}
     }
 
-    /// Exchanges messages between the remote server and enclave. Returns `true` when the
-    /// connection should remain active, false otherwise
-    fn proxy(&mut self) -> bool {
-        fn exchange<S: StreamConnection, D: StreamConnection>(src: &mut S, src_name: &str, dst: &mut D, dst_name: &str) -> bool {
-            // According to the `Read` threat documentation, reading 0 bytes
-            // indicates that the connection has been shutdown correctly. So we
-            // close the proxy service
-            // https://doc.rust-lang.org/std/io/trait.Read.html#tymethod.read
-            match Server::transfer_data(src, src_name, dst, dst_name) {
-                Ok(n) if n == 0 => false,
-                Ok(_)           => true,
-                Err(_)          => false,
+    /// Exchanges messages between the remote server and enclave. Returns on error, or when one of
+    /// the connections terminated
+    pub fn proxy(&mut self) -> Result<(), IoError> {
+        let remote = &mut self.tcp_stream;
+        let enclave = &mut self.vsock_stream;
+
+        let mut golden_set = FdSet::new();
+        golden_set.insert(remote.as_raw_fd());
+        golden_set.insert(enclave.as_raw_fd());
+
+        while golden_set != FdSet::new() {
+            let mut read_set = golden_set.clone();
+
+            if let Ok(_num) = select(None, Some(&mut read_set), None, None, None) {
+                if read_set.contains(remote.as_raw_fd()) {
+                    // According to the `Read` trait documentation, reading 0 bytes
+                    // indicates that the connection has been shutdown (for writes) correctly. We
+                    //  - reflect this change on the other connection
+                    //  - avoid reading from the socket again
+                    // https://doc.rust-lang.org/std/io/trait.Read.html#tymethod.read
+                    if Server::transfer_data(remote, &self.remote_name, enclave, "enclave")? == 0 {
+                        enclave.shutdown(Shutdown::Write)?;
+                        golden_set.remove(remote.as_raw_fd());
+                    }
+                }
+                if read_set.contains(enclave.as_raw_fd()) {
+                    if Server::transfer_data(enclave, "enclave", remote, &self.remote_name)? == 0 {
+                        remote.shutdown(Shutdown::Write)?;
+                        golden_set.remove(enclave.as_raw_fd());
+                    }
+                }
             }
         }
-        let remote = self.tcp_stream.as_raw_fd();
-        let proxy = self.vsock_stream.as_raw_fd();
-
-        let mut read_set = FdSet::new();
-        read_set.insert(remote);
-        read_set.insert(proxy);
-
-        if let Ok(_num) = select(None, Some(&mut read_set), None, None, None) {
-            if read_set.contains(remote) {
-                return exchange(&mut self.tcp_stream, &self.remote_name, &mut self.vsock_stream, "proxy");
-            }
-            if read_set.contains(proxy) {
-                return exchange(&mut self.vsock_stream, "proxy", &mut self.tcp_stream, &self.remote_name);
-            }
-        }
-        true
+        Ok(())
     }
 }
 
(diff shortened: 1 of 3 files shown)


fortanix-vme/fortanix-vme-runner/src/lib.rs
```diff
@@ -89,14 +89,98 @@ impl Listener {
...
+        read_set.insert(remote);
+        read_set.insert(proxy);
+
+        if let Ok(_num) = select(None, Some(&mut read_set), None, None, None) {
+            if read_set.contains(remote) {
+                return exchange(&mut self.tcp_stream, &self.remote_name, &mut self.vsock_stream, "proxy");
+            }
+            if read_set.contains(proxy) {
+                return exchange(&mut self.vsock_stream, "proxy", &mut self.tcp_stream, &self.remote_name);
+            }
```
Pagten, line 140: Is there a risk for starvation of one side of the connection here? Suppose that the `read_set` continually contains `remote` each time `fn proxy()` is called, then we'd never get to exchange data in the other direction. Probably that's okay according to the TCP spec but it might be nicer to still do the exchange from the vsock stream to the TCP stream even if there was an exchange in the other direction.

fortanix-vme/fortanix-vme-runner/src/lib.rs
```diff
@@ -89,14 +89,98 @@ impl Listener {
...
 pub struct Server {
     command_listener: Mutex<VsockListener>,
     /// Tracks information about TCP sockets that are currently listening for new connections. For
     /// every TCP listener socket in the runner, there is a vsock listener socket in the enclave.
     /// When the enclave instructs to accept a new connection, the runner accepts a new TCP
     /// connection. It then locates the ListenerInfo and finds the information it needs to set up a
     /// new vsock connection to the enclave
-    listeners: Mutex<FnvHashMap<VsockAddr, Arc<Mutex<Listener>>>>,
+    listeners: RwLock<FnvHashMap<VsockAddr, Arc<Mutex<Listener>>>>,
+    connections: RwLock<FnvHashMap<ConnectionKey, Arc<Mutex<Connection>>>>,
```
Pagten, line 183: Just an idea that I haven't thought through completely: I think we could get rid of both the outer `RwLock` and the inner `Mutex` by assigning ownership of connections to a specific thread. What I mean is that, instead of giving multiple threads (non-mutable) access to the entire `Arc<Server>`, we could give the thread that handles the proxying ownership of just the `HashMap` containing the connections. Then that thread wouldn't need any locking for accessing the `HashMap` or the `Connection`s contained within. Of course, that thread would have be responsible for adding connections to the `HashMap` as well. I think that could be achieved by sending `Connection`s across a channel from the command thread to the proxying thread(s).

fortanix-vme/fortanix-vme-runner/src/lib.rs
```diff
@@ -286,60 +389,55 @@ impl Server {
...
+            // Exchange messages on every proxy connection
+            // TODO: Store connections as a linked hash map so we don't need to keep a read lock
+            // over the HashMap while every connection is serviced
+            if let Ok(connections) = server.connections.read() {
+                for (key, connection) in connections.iter() {
+                    match connection.try_lock() {
+                        Ok(mut connection) => if !connection.proxy() {
+                                            connection.close();
+                                            closed_connections.push(key.clone());
+                                          }
```
Pagten, line 415: Nit: wrapping the `if`-block into a separate block might be slightly easier to read:
```
Ok(mut connection) => {
    if !connection.proxy() {
        connection.close();
        closed_connections.push(key.clone());
    }
}
```

fortanix-vme/fortanix-vme-runner/src/lib.rs
```diff
@@ -286,60 +389,55 @@ impl Server {
...
-                            eprintln!("transfer from remote failed: {:?}", e);
-                            break;
-                        }
+            // Exchange messages on every proxy connection
+            // TODO: Store connections as a linked hash map so we don't need to keep a read lock
+            // over the HashMap while every connection is serviced
+            if let Ok(connections) = server.connections.read() {
+                for (key, connection) in connections.iter() {
+                    match connection.try_lock() {
+                        Ok(mut connection) => if !connection.proxy() {
```
Pagten, line 412: Since there is no timeout specified on the `select()` calls inside `connection.proxy()`, won't this call block indefinitely in case no data has been sent in either direction? It seems like that would block data from being transferred on other connections.

fortanix-vme/fortanix-vme-runner/src/lib.rs
```diff
@@ -286,60 +389,55 @@ impl Server {
...
+            let mut num_connections = None;
+            if let Ok(mut connections) = server.connections.try_write() {
+                while let Some(k) = closed_connections.pop() {
+                    connections.remove(&k);
                 }
+                num_connections = Some(connections.len());
+            }
+
+            if num_connections == Some(0) {
+                thread::yield_now();
```
Pagten, line 431: Is there a risk of having excessive CPU usage when there are no connections? We basically keep polling in that case. An alternative would be to use a condition variable to block until we have connections again.

fortanix-vme/fortanix-vme-runner/src/lib.rs
```diff
@@ -89,14 +89,98 @@ impl Listener {
...
+    fn close(&self) {
+        let _ = self.tcp_stream.shutdown(Shutdown::Both);
+        let _ = self.vsock_stream.shutdown(Shutdown::Both);
+    }
+
+    /// Exchanges messages between the remote server and enclave. Returns `true` when the
+    /// connection should remain active, false otherwise
+    fn proxy(&mut self) -> bool {
+        fn exchange<S: StreamConnection, D: StreamConnection>(src: &mut S, src_name: &str, dst: &mut D, dst_name: &str) -> bool {
+            // According to the `Read` threat documentation, reading 0 bytes
```
Pagten, line 117: Typo: threat --> trait

fortanix-vme/fortanix-vme-runner/src/lib.rs
raoulstrackx: Yes that could be a problem. Good catch!

fortanix-vme/fortanix-vme-runner/src/lib.rs
```diff
@@ -91,55 +91,43 @@ impl Listener {
...
+            read_set.insert(proxy.as_raw_fd());
+
+            if let Ok(_num) = select(None, Some(&mut read_set), None, None, None) {
+                if read_set.contains(remote.as_raw_fd()) {
+                    // According to the `Read` trait documentation, reading 0 bytes
+                    // indicates that the connection has been shutdown correctly. So we
+                    // close the proxy service
+                    // https://doc.rust-lang.org/std/io/trait.Read.html#tymethod.read
+                    if Server::transfer_data(remote, remote_name, proxy, "proxy")? == 0 {
+                        return Ok(());
```
Pagten, line 121: Similar comment to the previous code: because we `return` here, I think we could in principle have starvation of the proxy --> remote side of the connection.

fortanix-vme/fortanix-vme-runner/src/lib.rs
```diff
@@ -317,10 +305,18 @@ impl Server {
...
 
-    fn add_connection(&self, runner_enclave: VsockStream<Std>, runner_remote: TcpStream, remote_name: String) {
-        let k = ConnectionKey::from_vsock_stream(&runner_enclave);
-        let info = Connection::new(runner_enclave, runner_remote, remote_name);
-        self.connections.write().unwrap().insert(k.clone(), Arc::new(Mutex::new(info)));
+    fn add_connection(self: Arc<Self>, mut runner_enclave: VsockStream<Std>, mut runner_remote: TcpStream, remote_name: String) -> Result<JoinHandle<()>, IoError> {
+        thread::Builder::new().spawn(move || {
+            let k = ConnectionKey::from_vsock_stream(&runner_enclave);
+            let connection = Arc::new(Connection::new(&runner_enclave, &runner_remote));
+            self.connections.write().unwrap().insert(k.clone(), connection.clone());
```
Pagten, line 312: Perhaps it would be better to add the connection to the hash map in the parent thread instead of in the proxying thread.Then we have the guarantee that the connection will have been added to the map when the `add_connection()` function returns.

Pagten: The `connection.clone()` call or even the `Arc` don't seem necessary with the current code. Perhaps they are in preparation of PLAT-367?

Pagten (changes requested):

fortanix-vme/fortanix-vme-runner/src/lib.rs
raoulstrackx: Yes it would. Discussed offline and decided to return to a single thread per connection. This would work nicely and efficiently with async

Pagten (approved):

PR #371, opened 2021-12-16
[PLAT-423] Avoid non-stable feature

raoulstrackx:
Remove non-stable features so CI can move to stable rust compiler 1.59 once that lands.

### DIFF
--- a/fortanix-vme/tests/iron/src/main.rs
+++ b/fortanix-vme/tests/iron/src/main.rs
@@ -1,17 +1,15 @@
-#![feature(once_cell)]
 use iron::prelude::*;
 use iron::{BeforeMiddleware, AfterMiddleware, typemap};
-use std::lazy::SyncOnceCell;
-use std::sync::Mutex;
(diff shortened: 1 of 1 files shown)


fortanix-vme/tests/iron/src/main.rs
```diff
@@ -1,14 +1,17 @@
...
+
 use iron::prelude::*;
 use iron::{BeforeMiddleware, AfterMiddleware, typemap};
-use std::lazy::SyncOnceCell;
 use std::sync::Mutex;
 use time;
 
-static NUM_SUCCEEDING_CONNECTIONS: SyncOnceCell<Mutex<u32>> = SyncOnceCell::new();
+lazy_static! {
+    static ref NUM_SUCCEEDING_CONNECTIONS: Mutex<u32> = Mutex::new(0);
```
jethrogb, line 10: Just use an AtomicU32?

raoulstrackx: Oh that looks interesting!

PR #373, opened 2021-12-17
[PLAT-367] Connection info

raoulstrackx:
When `TcpListener`s or `TcpStream`s are created from a `RawFd`, they lack information about local and/or peer ports used in the runner. When this information is requested, it needs to be fetched from the runner.


Pagten (approved):

Pagten:
bors d+

PR #375, opened 2021-12-23
[PLAT-421] Improve error handling

raoulstrackx:


fortanix-vme/fortanix-vme-abi/src/lib.rs
```diff
@@ -108,6 +112,32 @@ pub enum Response {
...
+
+#[cfg(feature="std")]
+impl From<VsockError> for Error {
+    fn from(error: VsockError) -> Error {
+        match error {
+            VsockError::EntropyError        => Error::Unknown,
+            VsockError::SystemError(errno)  => Error::SystemError(errno),
+            VsockError::WrongAddressType    => Error::Unknown,
+            VsockError::ZeroDurationTimeout => Error::Unknown,
+            VsockError::ReservedPort        => Error::Unknown,
```
raoulstrackx, line 138: This is not ideal. It's unlikely that a `VsockError` will ever be returned to an enclave, and even less that the enclave is able the handle it properly, but still we need to return something and `Unknown` may not be ideal.

fortanix-vme/fortanix-vme-runner/src/lib.rs
```diff
@@ -186,79 +218,87 @@ impl ConnectionKey {
...
+    sender: VsockStream,
+    reader: StreamDeserializer<'de, IoRead<VsockStream>, Request>,
 }
 
-impl Server {
-    fn log_communication(src: &str, src_port: u32, dst: &str, dst_port: u32, msg: &str, arrow: Direction, prot: &str) {
+impl<'de> ClientConnection<'de> {
+    pub fn new(stream: VsockStream) -> Self {
+        // Mimic splitting the stream in a read and send part
+        let sender = unsafe { VsockStream::from_raw_fd(stream.as_raw_fd()) };
```
raoulstrackx, line 229: `tokio::net::TcpStream::split` does exactly this, in an elegant way. When we move to Async such support may be added to `vsock` as well. It used to work without this my enabling read/write on top of a `&VsockStream`.

fortanix-vme/fortanix-vme-abi/src/lib.rs
Pagten: We currently only have three variants of the `Error` enum. Is there a reasoning/strategy behind this?

fortanix-vme/fortanix-vme-runner/src/lib.rs
```diff
@@ -12,6 +14,7 @@ use std::sync::{Arc, Mutex, RwLock};
 use fortanix_vme_abi::{self, Addr, Error as VmeError, Response, Request};
 use vsock::{self, SockAddr as VsockAddr, Std, Vsock, VsockListener, VsockStream};
 
+const MAX_MESSAGE_LEN: usize = 80;
```
Pagten, line 17: Nit: perhaps rename to `MAX_LOG_MESSAGE_LEN`? When I first saw this variable I thought it was the max length of some message communicated between the runner and the enclave.

Pagten (approved):

fortanix-vme/fortanix-vme-abi/src/lib.rs
raoulstrackx: For many errors, we simply want to forward the error to the enclave. When for example binding to a TCP port fails, the error can likely be translated to a `SystemError(errno)`, send back to the enclave, and there transformed again into a similar error.
Logical errors may arise in the runner that require their own error variant. For now there only is `ConnectionNotFound` that is returned when the enclave requests information about a non-existent connection.

PR #378, opened 2022-01-01
[PLAT-351]: Support an easy way to run Nitro enclaves

MihirLuthra:
Currently, to spawn an enclave, one has to do the following 3 steps:
- Convert elf to eif
- (not always required) Start fortanix-vme-runner which is basically a vsock proxy
- nitro-cli run-enclave

This PR adds code for a [custom runner for target `x86_64-unknown-linux-fortanixvme`](https://doc.rust-lang.org/cargo/reference/config.html#targettriplerunner) which will automate the above steps. This is pretty much like `ftxsgx-runner-cargo`. Although, I haven't created a separate `ftxvme-runner` like `ftxsgx-runner` as we can use `nitro-cli` directly here.

This runner spawns `fortanix-vme-runner` (this can be disabled by options in metadata) but doesn't wait on it. So, it doesn't know its exit status. If `fortanix-vme-runner` has been spawned before, it will not cause issues for the runner as when it tries to spawn it again, the address will already be in use and nothing will happen (although, the error message is displayed).
I think doing this should be enough as `fortanix-vme-runner` is being started here just for convenience. If someone wants to be certain of the exit status, they can set `disable-fortanix-vme-runner-start` to true in metadata and start `fortanix-vme-runner` explicitly.

---

I have tested this by setting the following in `~/.cargo/config`:
```toml
[target.x86_64-unknown-linux-fortanixvme]
runner = "/path/to/bin/of/ftxvme-runner-cargo"
```
and then in a simple rust crate, ran:
```
CC=musl-gcc \
  RUSTFLAGS="-Clink-self-contained=yes"\
  cargo run --target ${VME_TARGET} -Zbuild-std
```
This runs the enclave and its output can be further monitored by `nitro-console`.

### DIFF
--- a/fortanix-vme/eif-tools/src/bin/ftxvme-runner-cargo.rs
+++ b/fortanix-vme/eif-tools/src/bin/ftxvme-runner-cargo.rs
@@ -194,7 +194,7 @@ impl FortanixVmeConfig {
                 package.metadata
                     .map(|metadata| metadata.fortanix_vme)
             })
-        .flatten();
+            .flatten();
 
         let config = if let Some(fortanix_vme_metadata) = fortanix_vme_metadata {
 
--- a/fortanix-vme/eif-tools/src/bin/ftxvme-runner-cargo.rs
+++ b/fortanix-vme/eif-tools/src/bin/ftxvme-runner-cargo.rs
@@ -152,12 +152,6 @@ struct FortanixVmeConfig {
     /// `true` by default. This enables debug mode of `nitro-cli run-enclave`.
     debug_mode: bool,
 
-    /// The path to a .json configuration file that specifies the paramaters for the enclave
-    ///
-    /// If this is given, no other entry should be there in the table otherwise
-    /// this will error out.
-    nitro_cli_run_enclave_config: Option<PathBuf>,
-
     /// fortanix-vme-runner starts a vsock proxy server and
     /// is needed if your edp application makes any call to
     /// functions like `TcpStream::connect()`.
@@ -229,7 +223,6 @@ impl Default for FortanixVmeConfig {
             enclave_name: None,
             cpu_ids: None,
             enclave_cid: None,
-            nitro_cli_run_enclave_config: None,
             verbose: false,
             eif_file_path: FortanixVmeConfig::default_eif_path(),
             resource_path: None,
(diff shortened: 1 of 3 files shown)


jethrogb:
> Start fortanix-vme-runner (not always required)
> nitro-cli run-enclave

These should not be separate steps

MihirLuthra:
> These should not be separate steps

IIUC, to run an enclave prior to this PR, we had to start a `vsock-proxy-server` (which is basically `fortanix-vme-runner`) and after that we run the enclave by `nitro-cli run-enclave <args>` which are separate steps.

This PR adds a runner which will do all above mentioned steps in one go i.e. convert from elf to eif, then start vsock proxy and then run enclave.

Please correct me if I am wrong here.

fortanix-vme/eif-tools/src/bin/ftxvme-runner-cargo.rs
```diff
@@ -0,0 +1,302 @@
...
+        // We just try to start fortanix-vme-runner and don't wait on it. So,
+        // we don't know if it errors out. I think this should be enough
+        // as purpose of the runner is to provide easy setup. If someone really wants to
+        // be sure, they can set `disable_fortanix_vme_runner_start` to true and start
+        // it manually.
+        //
+        // If required, other option can be to link fortanix-vme-runner as a library and spawn the
+        // proxy server in another thread. The thing here will be that if fortanix-vme-runner
+        // already running, it will error out will addr in use. To be sure, we could probably
+        // communicate with fortanix-vme-runner.
```
raoulstrackx, line 267: Why would we want to disable the runner from starting? Yes starting multiple runners will currently error out because it won't be able to bind to vsock port 10000, but that'll change in the future; we'll ensure that the runner contacts the enclave, instead of the other way around. I'd remove this option to avoid having to support it in the future.

fortanix-vme/eif-tools/src/bin/ftxvme-runner-cargo.rs
```diff
@@ -0,0 +1,302 @@
...
+    ///
+    /// If this is set to true, this runner won't try to
+    /// start fortanix-vme-runner.
+    disable_fortanix_vme_runner_start: bool,
+}
+
+impl FortanixVmeConfig {
+    const DEFAULT_CPU_COUNT: isize = 2;
+    const DEFAULT_MEMORY: isize = 512;
+    const DEFAULT_DEBUG_MODE: bool = true;
```
raoulstrackx, line 175: I think it's safer to make this opt-in; having `false` by default.

fortanix-vme/eif-tools/src/bin/ftxvme-runner-cargo.rs
MihirLuthra: Makes sense, I will remove this option.

fortanix-vme/eif-tools/src/bin/ftxvme-runner-cargo.rs
```diff
@@ -0,0 +1,263 @@
...
+    /// the name of the .eif file is used.
+    enclave_name: Option<String>,
+
+    /// Specifies the number of vCPUs to allocate to the enclave.
+    /// Shouldn't coexist with `cpu_ids`.
+    cpu_count: Option<isize>,
+
+    /// Specifies the IDs of the vCPUs to allocate to the enclave.
+    /// Shouldn't coexist with `cpu_count`.
+    cpu_ids: Option<String>,
```
raoulstrackx, line 136: Sorry I missed this during the first pass. The `cpu_ids` is something we don't use and will only be used in the future when physical cores may differ. If that ever will be the case. It's safer to remove this option as well.

fortanix-vme/eif-tools/src/bin/ftxvme-runner-cargo.rs
```diff
@@ -0,0 +1,263 @@
...
+
+    /// Basically this field represents the vsock address of
+    /// the enclave.
+    /// This field should always be greater than equal to 4.
+    ///
+    /// The CID of parent instance is always 3 as per:
+    /// https://docs.aws.amazon.com/enclaves/latest/user/nitro-enclave-concepts.html#term-socket
+    ///
+    /// Vsock CID is analogous to an IP address.
+    enclave_cid: Option<String>,
```
raoulstrackx, line 150: Similar as above. We might want to have flexibility in choosing the cid in the future.

raoulstrackx:
bors d+

jethrogb:
> IIUC, to run an enclave prior to this PR, we had to start a vsock-proxy-server (which is basically fortanix-vme-runner) and after that we run the enclave by nitro-cli run-enclave <args> which are separate steps.

That's my point, they shouldn't be.

raoulstrackx:
We all agree the `fortanix-vme-runner` should be in charge of starting and stopping the enclave. Unfortunately the runner doesn't support that just yet. I was hoping to get this PR merged and removing the dependency on `nitro-cli` later.

MihirLuthra:
I will go ahead and merge this PR. Will create follow up PRs if changes required.

PR #381, opened 2022-01-19
[PLAT-66] Nitro nsm

raoulstrackx:
The AWS `nsm-driver` and `nsm-io` crates provide an interface to the AWS Nitro Security Module (NSM), but this is not Rust friendly. This new `nsm` crate avoids having the same conversion everywhere.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -4,11 +4,13 @@ members = [
     "fortanix-vme/eif-tools",
     "fortanix-vme/fortanix-vme-abi",
     "fortanix-vme/fortanix-vme-runner",
+    "fortanix-vme/nsm",
     "fortanix-vme/nitro-attestation-verify",
     "fortanix-vme/tests/hello_world",
     "fortanix-vme/tests/outgoing_connection",
     "fortanix-vme/tests/incoming_connection",
     "fortanix-vme/tests/iron",
+    "fortanix-vme/tests/nsm-test",
     "intel-sgx/aesm-client",
     "intel-sgx/dcap-provider",
     "intel-sgx/dcap-ql-sys",
@@ -34,6 +36,7 @@ exclude = ["examples"]
 [patch.crates-io]
 libc  = { git = "https://github.com/fortanix/libc.git", branch = "fortanixvme" }
 mbedtls = { git = "https://github.com/fortanix/rust-mbedtls", branch = "master" }
+nix   = { git = "https://github.com/fortanix/nix.git", branch = "raoul/fortanixvme_r0.20.2" }
 serde = { git = "https://github.com/fortanix/serde.git", branch = "master" }
 vsock = { git = "https://github.com/fortanix/vsock-rs.git", branch = "fortanixvme" }
 rustc-serialize = { git = "https://github.com/jethrogb/rustc-serialize.git", branch = "portability" }
--- a/fortanix-vme/nsm/src/lib.rs
+++ b/fortanix-vme/nsm/src/lib.rs
@@ -214,6 +214,14 @@ impl Nsm {
     pub fn describe(&self) -> Result<Description, Error> {
         nsm_driver::nsm_process_request(self.0, Request::DescribeNSM).try_into()
     }
+
+    pub fn get_random(&self) -> Result<Vec<u8>, Error> {
+        match nsm_driver::nsm_process_request(self.0, Request::GetRandom) {
+            Response::GetRandom{ random }    => Ok(random),
+            Response::Error(code)            => Err(code.into()),
+            _                                => Err(Error::InvalidResponse),
+        }
+    }
 }
 
 impl Drop for Nsm {
--- a/fortanix-vme/nsm/src/lib.rs
+++ b/fortanix-vme/nsm/src/lib.rs
@@ -4,7 +4,7 @@ pub use serde_bytes::ByteBuf;
 
 pub struct Nsm(i32);
 
-#[derive(Debug)]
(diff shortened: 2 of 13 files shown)


fortanix-vme/nsm/src/lib.rs
```diff
@@ -0,0 +1,275 @@
...
+            Error::AttestationError(ref msg) => write!(fmt, "Attestation error: {}", msg),
+            Error::BufferTooSmall => write!(fmt, "Buffer too small"),
+            Error::CannotOpenDriver => write!(fmt, "CannotOpenDriver"),
+            Error::InputTooLarge => write!(fmt, "InputTooLarge"),
+            Error::InternalError => write!(fmt, "InternalError"),
+            Error::InvalidArgument => write!(fmt, "InvalidArgument"),
+            Error::InvalidOperation => write!(fmt, "InvalidOperation"),
+            Error::InvalidPcrIndex => write!(fmt, "InvalidPcrIndex"),
+            Error::InvalidResponse => write!(fmt, "InvalidResponse"),
+            Error::ReadOnlyPcrIndex => write!(fmt, "ReadOnlyPcrIndex"),
```
Pagten, line 35: Looks like we could write this as
```
impl std::fmt::Display for Error {
    fn fmt(&self, fmt: &mut std::fmt::Formatter) -> std::fmt::Result {
        std::fmt::Debug::fmt(self, fmt)
    }
}
```
as well.

fortanix-vme/nsm/src/lib.rs
```diff
@@ -0,0 +1,275 @@
...
+
+    /// The PCRs that are read-only.
+    pub fn locked_pcrs(&self) -> &BTreeSet<u16> {
+        &self.locked_pcrs
+    }
+
+    /// The digest of the PCR Bank
+    pub fn digest(&self) -> Digest {
+        self.digest
+    }
```
Pagten, line 166: Are these getters needed, since the fields are public?

fortanix-vme/nsm/src/lib.rs
```diff
@@ -0,0 +1,275 @@
...
+            Response::Error(code) => Err(code.into()),
+            _ => Err(Error::InvalidResponse),
+        }
+    }
+
+    pub fn describe_pcr(&mut self, idx_pcr: u16) -> Result<Pcr, Error> {
+        let req = Request::DescribePCR {
+            index: idx_pcr,
+        };
+        nsm_driver::nsm_process_request(self.0, req).try_into()
```
Pagten, line 224: A disadvantage of building a `Pcr` instance from a `Response` like this, is that we do not detect the (admittedly, bizarre) case when the driver would respond with a `ExtendPCR` response to a `DescribePCR` request or vice versa. I think I personally wouldn't do the conversion from `Response`s to `Pcr`s or `Description`s using the `TryFrom` trait.

Pagten (changes requested):

fortanix-vme/nsm/src/lib.rs
```diff
@@ -0,0 +1,275 @@
+pub use nitro_attestation_verify::{AttestationDocument, Unverified, NitroError as AttestationError, Mbedtls};
```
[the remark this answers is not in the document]
raoulstrackx: Yes good idea!

Cargo.toml
```diff
@@ -34,6 +36,7 @@ exclude = ["examples"]
 [patch.crates-io]
```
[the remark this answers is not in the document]
raoulstrackx: Most patches are in there for the `fortanixvme` target. We can't do much about that. I doubt these patches will get upstreamed before the target makes it in the Rust compiler. But at least `mbedtls` could/should get published.
I agree that that should be checked before publishing these crate as a form of QA.

fortanix-vme/nsm/src/lib.rs
raoulstrackx: Oops. I created these getters as clients never need to create their own `Description` value. So having one, must have come from the `nsm` module. I forgot to make the fields private afterwards.

fortanix-vme/nsm/src/lib.rs
raoulstrackx: It's an edge case, but may be worth while to avoid it.

fortanix-vme/nsm/src/lib.rs
```diff
@@ -0,0 +1,263 @@
+//! The nsm_io library provides an interface to the AWS Nitro Security Module (nsm). Unfortunately,
+//! the interface of this crate is not very Rust friendly. To avoid clients of this crate having to
+//! write the same wrapper code over and over again, this crate does this once.
```
Pagten, line 3: Looks like the term "this crate" is used both to refer to the `nsm_io` library and to this `nsm` crate.

Pagten (approved):

Cargo.toml
```diff
@@ -4,11 +4,13 @@ members = [
     "fortanix-vme/eif-tools",
     "fortanix-vme/fortanix-vme-abi",
     "fortanix-vme/fortanix-vme-runner",
+    "fortanix-vme/nsm",
```
jethrogb, line 7: Probably the nitro-enclaves-specific stuff shouldn't be in fortanix-vme.

raoulstrackx: Yes I wanted to create a separate PR just for moving things. The `eif-tools` also shouldn't be directly in `fortanix-vme`

PR #382, opened 2022-01-26
[PLAT-545] Move Nitro related tools in a separate directory

raoulstrackx:
The rust-sgx repo contains a lot of vme/nitro related crates. These are all in the same directory. It is cleaner to place them in separate crates.
This PR does not perform any functional changes.


Pagten (approved):

Pagten:
bors d+

Cargo.toml
```diff
@@ -1,16 +1,16 @@
 [workspace]
 members = [
+    "fortanix-vme/aws-nitro/eif-tools",
```
jethrogb, line 3: It's nitro enclaves. Nitro is something else.

raoulstrackx: Yes you have a point that `aws-nitro-enclaves` is a more suitable name.

PR #383, opened 2022-01-28
[PLAT-355] Avoid automatic cbor derive

raoulstrackx:
The `fortanix-vme-abi` crate provides an interface between the enclave and the enclave-runner. Unfortunately we can't rely on automatic `Serialize` and `Deserialize` macros. See [this conversation](https://rust-lang.zulipchat.com/#narrow/stream/131828-t-compiler/topic/-Z.20dual-proc-macros) for an explanation.
This PR avoids these macros for now by basically expanding the macro. Care has been taken to not introduce any functional changes. Old runners will thus still work with new enclaves and vice versa. Code may be cleaned up when:
* We decide to replace `serde_cbor`
* We introduce changes to accommodate message streams.


raoulstrackx:
Will file another PR based on the [serde_generate crate](https://docs.rs/serde-generate/latest/serde_generate/index.html)

fortanix-vme/fortanix-vme-abi/src/lib.rs
```diff
@@ -54,6 +1250,658 @@ pub enum Addr {
     },
 }
 
+/// Serializes an `Addr` value. We can't rely on the `serde` `Serialize` macro as we wish to use
```
Pagten: Nit: I'd write "We can't rely on the `serde_derive` macro ..." since technically `Serialize` is not a macro.

fortanix-vme/fortanix-vme-abi/src/lib.rs
```diff
@@ -54,6 +1250,658 @@ pub enum Addr {
...
+                SerializeStructVariant::serialize_field(&mut __serde_state, "port", port)?;
+                SerializeStructVariant::serialize_field(&mut __serde_state, "flowinfo", flowinfo)?;
+                SerializeStructVariant::serialize_field(&mut __serde_state, "scope_id", scope_id)?;
+                SerializeStructVariant::end(__serde_state)
+            }
+        }
+    }
+}
+
+/// Deserializes an `Addr` value. We can't rely on the `serde` `Deserialize` macro as we wish to use
```
Pagten: Same comment here

fortanix-vme/fortanix-vme-abi/src/lib.rs
```diff
@@ -109,14 +1957,1403 @@ pub enum Response {
     Failed(Error),
 }
 
-#[derive(Debug, PartialEq, Eq, Serialize, Deserialize)]
+/// Serializes a `Response` value. We can't rely on the `serde` `Serialize` macro as we wish to use
```
Pagten: Same comment here.

fortanix-vme/fortanix-vme-abi/src/lib.rs
```diff
@@ -158,4 +3397,305 @@ mod test {
             panic!("Not IPv4")
         }
     }
+
+    #[test]
+    fn test_error() {
+        let data: Vec<(Error, Vec<u8>)> = Vec::from([
+            (Error::ConnectionNotFound, Vec::from([0x72, 0x43, 0x6f, 0x6e, 0x6e, 0x65, 0x63, 0x74, 0x69, 0x6f,
```
Pagten: An alternative approach to testing this, would be to duplicate the structs and enums defined above here in this `test` module and add a `derive(Serialize, Deserialize)` attribute to them and then test that the result of the "manually" implemented `Serialize`/`Deserialize` traits matches that of the derived ones.

Pagten (commented):
If we decide to go with the approach of manually expanding the `derive(Serialize, Deserialize)` output and pasting it into these source files, then I'm fine with this PR. But, as discussed offline, I'd suggest looking into the using `serde-reflection` and `serde-generate` in a build script as an alternative. I'll hold off approving this PR until we have clarity on that.

PR #384, opened 2022-01-31
[PLAT-268] Test connecting to and binding on `localhost`

raoulstrackx:
An enclave should be able to bind a TCP listener on `localhost`. Likewise, it should be able to establish a local connection to the parent VM. Before this PR both scenarios worked correctly, but CI tests were missing.

### DIFF
--- a/fortanix-vme/aws-nitro-enclaves/tests/incoming_connection/src/main.rs
+++ b/fortanix-vme/aws-nitro-enclaves/tests/incoming_connection/src/main.rs
@@ -3,9 +3,13 @@ use std::io::{ErrorKind, Read, Write};
 use std::net::{IpAddr, Ipv4Addr, Shutdown, SocketAddr, TcpListener, TcpStream};
 use std::os::unix::io::{AsRawFd, FromRawFd};
 
-fn server_run() {
+fn server_run(run: u32) {
     println!("Bind TCP socket to port 3400");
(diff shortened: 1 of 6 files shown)


fortanix-vme/aws-nitro-enclaves/tests/incoming_connection/src/main.rs
```diff
@@ -3,9 +3,13 @@ use std::io::{ErrorKind, Read, Write};
 use std::net::{IpAddr, Ipv4Addr, Shutdown, SocketAddr, TcpListener, TcpStream};
 use std::os::unix::io::{AsRawFd, FromRawFd};
 
-fn server_run() {
+fn server_run(run: u32) {
```
Pagten, line 6: Taking the address to listen on as an argument would be a bit easier to read than taking this `run` argument.
```
fn server_run<A: ToSocketAddrs>(addr: A) {
    println!("Bind TCP socket to port 3400");
    let listener = TcpListener::bind(addr).expect("Bind failed");

   [...]
}

fn main() {
    for addr in &["127.0.0.1:3400", "localhost:3400"] {
        println!("Running server on {}", addr);
        server_run(addr);
    }
    println!("Bye bye");
}
```

Pagten (approved):

Pagten (approved):

PR #395, opened 2022-05-24
[PLAT-896] Verify fifo descriptor pointers

raoulstrackx:
When processing `FifoDescriptor` to turn it into a `ipc_queue`, the offset point isn't verified correctly.

### DIFF
--- a/ipc-queue/src/fifo.rs
+++ b/ipc-queue/src/fifo.rs
@@ -94,7 +94,7 @@ impl<T: Transmittable> Fifo<T> {
             "Fifo len should be a power of two"
         );
         #[cfg(target_env = "sgx")] {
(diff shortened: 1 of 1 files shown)


mzohreva:
@raoulstrackx if there are no more changes needed, then please bump up the ipc-queue crate's version before merge so we can publish it as soon as it gets merged

raoulstrackx:
The last published version is 0.1.0 and the version is currently at 0.1.1. I'll do some refactoring that will require bumping the minor version. So I propose we merge this PR as is, publish the crate to make sure everybody can easily pick up the patch and submit the next PR.

mzohreva:
Thanks, `ipc-queue` 0.1.1 is published now

PR #403, opened 2022-07-27
Version bump `tokio` crate

choco:
I've tried to touch as little as possible since I saw that there's a bigger PR that touches the enclave-runner to add async support to enclave interface.


Taowyoo:
Closed as version is already bumped, see https://github.com/fortanix/rust-sgx/blob/master/Cargo.lock#L4118-L4135

PR #408, opened 2022-09-05
Document `TCS` struct limitations

raoulstrackx:
Document that a TCS struct should never be located as the beginning of an enclave. (ref https://github.com/rust-lang/rust/pull/98391)


intel-sgx/fortanix-sgx-abi/src/lib.rs
```diff
@@ -441,6 +441,8 @@ impl Usercalls {
 }
 
 /// The absolute address of a TCS in the current enclave.
+/// To ensure a TCS struct will never be located at address 0, a TCS struct should not be located
```
jethrogb, line 444: Insert newline before?

PR #415, opened 2022-11-01
Ias certchain

monokles:
Adds verification for the full IAS certificate chain.


monokles:
Converting this to a draft while waiting for https://github.com/fortanix/rust-mbedtls/pull/112 to be merged in rust-mbedtls so we can add revocation list support at the same time.

PR #416, opened 2022-11-14
Change sgx_pkix crate version to 0.2.0. EDP-99.

arai-fortanix:
sgx_pkix version 0.1.2 was incorrectly numbered. This version changed the version of sgx-isa used, which in turn changed the version of the mbedtls dependency. mbedtls 0.8.0 is not compatible with version 0.7.0, so version 0.1.2 is not compatible with code that was using 0.1.1. This release should have been numbered 0.2.0.

Two other crates directly depend on sgx_pkix: em-app and ias. I've updated their dependencies, and bumped the third digit in their versions. We will want to yank sgx_pkix version 0.1.2,  em-app version 3.0, and ias version 1.0 when the new versions are published.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -813,7 +813,7 @@ dependencies = [
 
 [[package]]
 name = "em-app"
-version = "0.3.0"
+version = "0.3.1"
 dependencies = [
  "aws-nitro-enclaves-nsm-api",
  "b64-ct",
@@ -1662,7 +1662,7 @@ dependencies = [
 
 [[package]]
 name = "ias"
-version = "0.1.0"
+version = "0.1.1"
 dependencies = [
  "aesm-client",
  "base64 0.13.0",
@@ -3505,7 +3505,7 @@ dependencies = [
 
 [[package]]
 name = "sgx_pkix"
-version = "0.1.2"
+version = "0.2.0"
 dependencies = [
  "byteorder 1.3.4",
  "lazy_static",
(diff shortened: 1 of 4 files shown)


Pagten:
> mbedtls 0.8.0 is not compatible with version 0.7.0
What does that mean? Is it not possible to have a crate that depends on both versions? Or is the problem that `sgx-isa` and `sgx_pkix` re-export some of the mbedtls types?

In the former case, don't we also need to bump the minor version (second digit) of `em-app` and `ias` instead of their patch version?

arai-fortanix:
I think the versions of `em-app` and `ias` are okay, aside from depending on a bad version of `sgx_pkix`. `em-app` got bumped from `0.2.0` to `0.3.0` when the required version of mbedtls was increased, and with `ias` the only published version of `ias` uses `mbedtls` `0.8.0`. It looks like we never published the older versions of the `ias` code that worked with `mbedtls 0.7.0`.

I'm still working on reproducing the specific build issue with mbedtls. The problem is that some of the library APIs take mbedtls types as parameters, so you end up with incorrect structure versions.

arai-fortanix:
There may not be a specific issue with the mbedtls version, but there are other crate versioning issues that make it so we cannot pick up the latest `sgx_pkix` crate in roche without fixing a bunch of other things. `sgx_pkix` depends on `sgx-api` and some of the APIs in `sgx_pkix` take types from `sgx-api` as parameters. This can ultimately be fixed, but is going to be a lot of work to update everything to newer versions of the crates.

What I want right now is to add a new OID definition to the sgx_pkix crate, without having to update all of these other dependencies.

arai-fortanix:
bors r=[Pagten]

PR #421, opened 2022-12-02
Enable travis CI for sgx-pkix_v0.1.x branch

Pagten:


arai-fortanix (approved):
Look good.

Pagten:
It took quite a number of tries, but CI is finally working now for this `sgx-pkix_v0.1.x` branch.

The strategy I employed is to enable the same builds/tests as we have in `master`, obviously except for the builds/tests of crates that aren't present yet in this branch. An alternative strategy would have been to enable just the build/test of `sgx_pkix`, since that's the only crate we will be modifying in this branch, but I figured we might catch some unintended API backcompat breakages by building the other packages in the repo as well.

raoulstrackx (approved):
Looks good to me!

PR #429, opened 2022-12-21
sgxs crate changes to enable offline signing

arai-fortanix:
This change adds new interfaces to the `sgxs` crate to make it possible to sign enclaves separately from creating the sigstructs. This is useful, for example, for performing signing via an HSM.

The first two commits in this sequence come from this pull request from ravenac95: https://github.com/fortanix/rust-sgx/pull/327. I did not include the changes to the command-line sgx-sign tool from that pull request. I think we can have a separate discussion about whether we want to support that model. The library changes should be less controversial.

I also included a change to expose the hash bytes from an EnclaveHash object. That comes from https://github.com/fortanix/rust-sgx/pull/341 from trevor-crypto.

This change should be backward-compatible with old code using this crate. New code that wants to use the new `cat_sign()` method and is using a custom key implementation will need to provide the new `SgxRsaPubOps()` trait for calculating the Q1 and Q2 values from a signature, instead of doing this during signing.

### DIFF
--- a/intel-sgx/sgxs/CHANGELOG.md
+++ b/intel-sgx/sgxs/CHANGELOG.md
@@ -1,0 +1,15 @@
+# Version 0.7.4 - 2022-12-21
+
+## New Features
+- Refactored the signing APIs to permit signatures to be generated separately (for example using an HSM).
+  The way to use this new functionality is to construct a Sigstruct as usual and call
+  `sigstruct.unsigned_hash()` to produce the hash that must be signed externally. Once the signature is
+  available, reconstruct the sigstruct and call `sigstruct.cat_sign()` with the signature to get the signed
+  and populated Sigstruct.
+- If you are using a custom key implementation, you will need to implement the new `SgxRsaPubOps()`
+  trait for your key. This trait must provide a `calculate_q1_q2()` method that calculates the q1 and q1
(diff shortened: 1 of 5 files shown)


intel-sgx/sgxs/CHANGELOG.md
```diff
@@ -0,0 +1,15 @@
+# Version 0.7.4 - 2022-12-21
+
+## New Features
+- Refactored the signing APIs to permit signatures to be generated separately (for example using an HSM).
+  The way to use this new functionality is to construct a Sigstruct as usual and call
+  `sigstruct.unsigned_hash()` to produce the hash that must be signed externally. Once the signature is
+  available, reconstruct the sigstruct and call `sigstruct.cat_sign()` with the signature to get the signed
+  and populated Sigstruct.
+- If you are using a custom key implementation, you will need to implement the new `SgxRsaPubOps()`
+  trait for your key. This trait must provide a `calculate_q1_q2()` method that calculates the q1 and q1
```
raoulstrackx, line 10: Nitpick: Typo "q1 and q1"

raoulstrackx (approved):

arai-fortanix:
bors r=[Pagten]

PR #431, opened 2023-01-06
refactor dcap-retrieve-pckid

Taowyoo:
Split the origin code in `dcap-retrieve-pckid` into `lib.rs` and `main.rs`

So that this code could be easy reuse elsewhere.

### DIFF
--- a/intel-sgx/dcap-retrieve-pckid/src/lib.rs
+++ b/intel-sgx/dcap-retrieve-pckid/src/lib.rs
@@ -5,11 +5,12 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 #![deny(warnings)]
-use std::fmt;
 use std::convert::TryInto;
+use std::fmt;
 
 use aesm_client::AesmClient;
 use dcap_ql::quote::{Qe3CertDataPpid, Quote, Quote3SignatureEcdsaP256, QuoteHeader};
+use failure::Error as FailureError;
 use sgx_isa::Targetinfo;
 #[cfg(windows)]
 use sgxs_loaders::enclaveapi::Sgx as IsgxDevice;
@@ -53,11 +54,11 @@ impl ToString for PckId {
     }
 }
 
-pub fn retrieve_pckid_str() -> Result<PckId, &'static str> {
+pub fn retrieve_pckid_str() -> Result<PckId, FailureError> {
     const SGX_QL_ALG_ECDSA_P256: u32 = 2;
 
     let mut device = IsgxDevice::new()
-        .map_err(|_| "Error opening SGX device")?
+        .map_err(|err| FailureError::from(err).context("Error opening SGX device"))?
         .einittoken_provider(AesmClient::new())
         .build();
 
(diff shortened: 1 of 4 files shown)


intel-sgx/dcap-retrieve-pckid/src/lib.rs
```diff
@@ -0,0 +1,84 @@
...
+    let quote = Quote::parse(res.quote()).map_err(|_| "Error parsing quote")?;
+    let QuoteHeader::V3 { user_data, .. } = quote.header();
+    let sig = quote
+        .signature::<Quote3SignatureEcdsaP256>()
+        .map_err(|_| "Error parsing requested signature type")?;
+    let cd_ppid = sig
+        .certification_data::<Qe3CertDataPpid>()
+        .map_err(|_| "Certification data is already available for the current platform")?;
+
+    Ok(format!(
```
arai-fortanix, line 76: Rather than returning a string, I think it makes more sense for this function to return these values as either a tuple or a struct. This will make it easier for the calling code to consume these values if they want to do something with them other than just printing them.

arai-fortanix:
This crate doesn't change very often, so it probably makes sense to publish a new version as soon as this change lands (and is tested). It's probably fine to just go ahead and bump the version in Cargo.toml instead of doing that in a separate change. We're unlikely to be waiting for more changes before publishing the new version.

intel-sgx/dcap-retrieve-pckid/src/lib.rs
arai-fortanix: But otherwise, this seems fine.

Taowyoo: Make sense, let me update it

Taowyoo: Done

Taowyoo:
> This crate doesn't change very often, so it probably makes sense to publish a new version as soon as this change lands (and is tested). It's probably fine to just go ahead and bump the version in Cargo.toml instead of doing that in a separate change. We're unlikely to be waiting for more changes before publishing the new version.

Hi @arai-fortanix , I bump the version number in the toml file of the crate.
Does this repo has any automatic action/robot(bors) to push new version of crate to crates.io?
Or I need to do that manually?

arai-fortanix:
I think the crate is owned by Fortanix, so someone at Fortanix will have to publish the new crate after the pull request is merged.

intel-sgx/dcap-retrieve-pckid/src/lib.rs
```diff
@@ -34,7 +35,25 @@ impl<'a> fmt::Display for PrintHex<'a> {
     }
 }
 
-pub fn retrieve_pckid_str() -> Result<String, &'static str> {
+pub struct PckCertId {
```
raoulstrackx, line 38: Could you rename this `PckId`? (It identifies a platform certification key, not a certificate, see https://download.01.org/intel-sgx/latest/dcap-latest/linux/docs/Intel_SGX_ECDSA_QuoteLibReference_DCAP_API.pdf table 9)

intel-sgx/dcap-retrieve-pckid/src/lib.rs
```diff
@@ -34,7 +35,25 @@ impl<'a> fmt::Display for PrintHex<'a> {
...
+    pub cd_ppid: Qe3CertDataPpid<'static>,
```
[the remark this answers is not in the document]
Taowyoo: Got it. I will return the tuple of `Qe3CertDataPpid` and `qe_id` directly, here.

intel-sgx/dcap-retrieve-pckid/src/lib.rs
Taowyoo: Got it.
Also related to comment above. I think the struct `PckId` is still needed, since it need a struct to wrap the ppid and qe_id instead of using a tuple here.

Taowyoo: Renamed

intel-sgx/dcap-retrieve-pckid/src/lib.rs
```diff
@@ -0,0 +1,99 @@
...
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        for b in self.0 {
+            write!(f, "{:02x}", b)?;
+        }
+        Ok(())
+    }
+}
+
+pub struct PckId {
+    pub cd_ppid: Qe3CertDataPpid<'static>,
```
arai-fortanix, line 39: This may be me not understanding lifetimes, but do we really want to force the `Qe3CertDataPpid` field to have `'static` lifetime? Would it not be better to do something like this instead:
```
pub struct PckId<'a> {
    pub cd_ppid: Qe3CertDataPpid<'a>,
    pub qe_id: [u8; 16],
}
```

Taowyoo: Since, the `cd_ppid` is local variable created inside the function, so it needs to be a owned object to return.
After calling following function, the reference inside it is converted to `static`.

https://github.com/fortanix/rust-sgx/blob/f10c25b72081addc4843c2d779ef9660ff7cf60f/intel-sgx/dcap-retrieve-pckid/src/lib.rs#L97

arai-fortanix (approved):
This looks good to me.

Taowyoo:
bors r=[arai-fortanix]

Issue #458, opened 2023-05-23
[BUG]: `rust-sgx` CI currently fails


raoulstrackx:
**Describe the bug:**
The `rust-sgx` CI [fails](https://app.travis-ci.com/github/fortanix/rust-sgx/builds/263242421) with:
```
 thread 'main' panicked at '"x509_san_other_name_union_(unnamed_at_vendor/include/mbedtls/x509_crt_h_115_5)" is not a valid Ident', /home/travis/.cargo/registry/src/github.com-1ecc6299db9ec823/proc-macro2-1.0.32/src/fallback.rs:702:9
```

**Related Issue(s):**
Any related issues or pull requests.

**Labels:**
Any specific labels you think should be applied to this issue.

**Additional context:**
Learnings from fixing the issue:
- Travis Ubuntu "patch" upgrade (20.04.5 -> 20.04.6) can cause a major upgrade of certain libraries, such as clang version 7 -> clang version 16. After a lot of tinkering and local tests, we were able to identify this as the root cause
- TravisCI seems to install its binaries (such as clang) not as an Apt package but directly in `/usr/local`. One should be aware that there's no simple package to uninstall this way, and the very dirty solution of `rm -rf /usr/local/clang-16` might be needed.
- clang can work perfectly... even when `clang --version` fails
- We've discovered an underlying issue with our build that caused breakage from simply the *presence* of a newer clang, even if the desired clang-11 is both installed and specified as an environment and CLI variables. An issue for that has been filled here: https://github.com/fortanix/rust-sgx/issues/465

PR #463, opened 2023-06-05
Fix CI breakage

vn971:
Fixes https://github.com/fortanix/rust-sgx/issues/458

Learnings from fixing the issue:
- Travis Ubuntu "patch" upgrade (20.04.5 -> 20.04.6) can cause a major upgrade of certain libraries, such as clang version 7 -> clang version 16. After a lot of tinkering and local tests, we were able to identify this as the root cause
- TravisCI seems to install its binaries (such as clang) not as an Apt package but directly in `/usr/local`. One should be aware that there's no simple package to uninstall this way, and the very dirty solution of `rm -rf /usr/local/clang-16` might be needed.
- clang can work perfectly... even when `clang --version` fails
- We've discovered an underlying issue with our build that caused breakage from simply the *presence* of a newer clang, even if the desired clang-11 is both installed and specified as an environment and CLI variables. An issue for that has been filled here: https://github.com/fortanix/rust-sgx/issues/465

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,6 +32,14 @@ matrix:
         - CFLAGS_x86_64_fortanix_unknown_sgx="-isystem/usr/include/x86_64-linux-gnu -mlvi-hardening -mllvm -x86-experimental-lvi-inline-asm-hardening"
         - CC_x86_64_fortanix_unknown_sgx=clang-11
       before_script:
+        # Travis-CI supplies clang version 16 by itself, in /usr/local
+        # However, our current build fails if llvm-16 is also available in the system. This is likely a bug,
+        # created and reported here: https://github.com/fortanix/rust-sgx/issues/465
+        # Once this bug is solved, please feel free to remove this workaround.
+        #
+        # Note that this clang-16 is some Travis-CI magic and not a standard Debian/Ubuntu Apt package.
+        # We're thus forced to force-remove the underlying directory in /usr/local
+        - sudo rm -rf /usr/local/*clang*
         - rustup target add x86_64-fortanix-unknown-sgx x86_64-unknown-linux-musl
         - rustup toolchain add nightly
         - rustup target add x86_64-fortanix-unknown-sgx --toolchain nightly


.travis.yml
```diff
@@ -20,20 +20,29 @@ matrix:
...
         - CC_x86_64_fortanix_unknown_sgx=clang-11
       before_script:
+        # Travis-CI supplies clang version 16 by itself, in /usr/local
+        # However, our current build fails if llvm-16 is also available in the system. This is likely a bug,
+        # created and reported here: https://github.com/fortanix/rust-sgx/issues/465
+        # Once this bug is solved, please feel free to remove this workaround.
+        #
+        # Note that this clang-16 is some Travis-CI magic and not a standard Debian/Ubuntu Apt package.
+        # We're thus forced to force-remove the underlying directory in /usr/local
+        - ls -aFl /usr/local | grep clang
```
raoulstrackx, line 42: Debug instructions?

.travis.yml
```diff
@@ -20,20 +20,29 @@ matrix:
...
+        - sudo rm -rf /usr/local/*clang*
         - rustup target add x86_64-fortanix-unknown-sgx x86_64-unknown-linux-musl
-        - rustup toolchain add nightly
+        - rustup toolchain add --profile minimal nightly
```
raoulstrackx, line 45: This doesn't seem required.

README.md
```diff
@@ -1,6 +1,6 @@
 # Fortanix Rust Enclave Development Platform
 
-[![Runtime Encryption Slack](https://img.shields.io/badge/Slack-%23rust--sgx-blue.svg?logo=slack&logoColor=whitesmoke)](https://fortanix.com/runtime-encryption-slack) [![Build Status](https://travis-ci.com/fortanix/rust-sgx.svg?branch=master)](https://travis-ci.com/fortanix/rust-sgx)
+[![Runtime Encryption Slack](https://img.shields.io/badge/Slack-%23rust--sgx-blue.svg?logo=slack&logoColor=whitesmoke)](https://fortanix.com/runtime-encryption-slack) [![Build Status](https://travis-ci.com/fortanix/rust-sgx.svg?branch=master)](https://app.travis-ci.com/github/fortanix/rust-sgx/builds)
```
raoulstrackx, line 3: Can we keep this for #466?

.travis.yml
vn971: Yes, it's unrelated. Should I move it to a separate PR? I think it's just nicer to always specify minimal dependencies/profiles/downloads of any kind for CI

raoulstrackx: Oh can you also revert this change? I'm not against it, but it should be a separate PR then. PRs are preferably the minimal set of changes to fix an issue / add a new feature. I was considering a similar comment about your changes to the `package` key, but you can stretch it a bit and say you're doing changes to the installed packages anyway.

vn971: If you mean the `packages` section, then I've just ordered them alphabetically. (Ordered things are just nicer.)

My personal preferences would be to be practical and merge as-is (especially since we don't have release trains and are asked to rebase if `master` changes). But that's my personal preference, I can easily go with what you suggest

Should I split or revert?

vn971: @jethrogb @raoulstrackx --^

raoulstrackx: Personally I would also only rearrange the packages when I'm adding one.
We don't have an agreement on whether its still acceptable, so let's be puristic and only keep the minimal change required: `sudo rm -rf /usr/local/*clang*` and its comment.

vn971: Sure! To maybe explain a bit more on my personal take on this, I prefer the "development boy scout rule":
> The boy scout rule is about making small, continuous improvements to your code. If you are fixing a bug, and find the code around it is a mess, fix a bit of that mess. You don’t have to make it perfect, you don’t have to fix it all, but improve it just a little bit.

https://backendtea.com/post/boyscout-rule/

So personally, I include the "area around the main task" as a possible scope of changes, but I exclude "unrelated" changes, e.g. something in "another area".

This view also has downsides, it's still a trade off. For example, if you'd have to revert said branch, you'd also revert the "cleaning up" part around. On the other hand, strict separation leads to no "cleaning up" at all, because most people wouldn't really submit a separate PR that fixes some typos or adds some links to the documentation of a method that was read and used but not changed in the PR.

That all being said, I'll remove everything but the minimal code changes to address the issue now

vn971: Done :checkered_flag:

PR #485, opened 2023-06-22
Support for specifying unmapped area

raoulstrackx:
SGXv2 supports dynamically adding memory. This initial PR adds support to the `ftxsgx-elf2sgxs` tool to reserve unmapped memory within the enclave.

### DIFF
--- a/intel-sgx/fortanix-sgx-tools/src/bin/ftxsgx-elf2sgxs.rs
+++ b/intel-sgx/fortanix-sgx-tools/src/bin/ftxsgx-elf2sgxs.rs
@@ -40,6 +40,8 @@ struct Symbols<'a> {
     sgx_entry: &'a DynSymEntry,
     HEAP_BASE: &'a DynSymEntry,
     HEAP_SIZE: &'a DynSymEntry,
+    UNMAPPED_BASE: &'a DynSymEntry,
+    UNMAPPED_SIZE: &'a DynSymEntry,
     RELA: &'a DynSymEntry,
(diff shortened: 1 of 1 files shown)


intel-sgx/fortanix-sgx-tools/src/bin/ftxsgx-elf2sgxs.rs
```diff
@@ -797,6 +810,16 @@ fn main() {
...
         )
+        .arg(
+            Arg::with_name("unmapped-memory-size")
+                .short("U")
+                .long("unmapped-memory-size")
+                .value_name("BYTES")
+                .validator(u64::validate_arg)
+                .required(false)
+                .default_value("0")
+                .help("Specify the (minimum) size of unmapped memory (useable on SGXv2 capable processors only)"),
```
vn971, line 821: I wonder, should we clarify why a user would be interested in this option? E.g.

> Specifying an unmapped memory size allows dynamically allocating additional heap space on SGXv2 capable processors

I understand on the other hand that maybe we don't want to advertise it if the whole feature set is not ready yet; not sure.
What do you think?

Looks good to me either way :+1:

raoulstrackx: Maybe we should even call it dynamic memory or something.

raoulstrackx: Good point about not adding it to master yet. Will mark the PR as draft again

PR #486, opened 2023-06-23
rs-libc: prepare for publish new version of `rs-libc`

Taowyoo:
- Update README
- Bump version number


arai-fortanix (approved):
This change by itself seems fine. I just want to check that the other changes included in this version bump are compatible, so it's okay to bump the version number to 0.2.4 and not go to 0.3.0.

Taowyoo:
Checked with `git diff rs-libc_v0.2.3..master -- rs-libc` 
The only functional changes are:
- add dependency of `zeroize`
- add function `explicit_bzero`

PR #495, opened 2023-07-12
Support for attaching identity to Nitro Debug Enclaves.

jjfiv:
- Define new OID for Nitro PCR0
 - Wrap nsm_driver in a cleaner API.
 - Also make em-app::error::Error impl std::error::Error


arai-fortanix (approved):

Taowyoo (dismissed):
LGTM

fortanix-vme/vme-pkix/src/oid.rs
```diff
@@ -4,10 +4,13 @@
...
-lazy_static!{
+lazy_static! {
+    /// This OID identifies an extension containing a Nitro attestation report.
     pub static ref ATTESTATION_NITRO: ObjectIdentifier = vec![1, 3, 6, 1, 4, 1, 49690, 2, 2, 6].into();
-}
\ No newline at end of file
+    /// This OID identifies an extension containing a the PCR0 of a Nitro enclave;
+    /// used to insecurely provide identification to debug attestations that
+    /// otherwise contain all zeroes in their PCR0 measurement.
+    pub static ref ATTESTATION_NITRO_PCR0: ObjectIdentifier = vec![1, 3, 6, 1, 4, 1, 49690, 2, 2, 6, 1].into();
```
vn971, line 15: Should we maybe add a reference to the documentation here, explaining where these numbers come from?

fortanix-vme/vme-pkix/src/oid.rs
```diff
@@ -4,10 +4,13 @@
...
-
 use pkix::types::ObjectIdentifier;
 
-lazy_static!{
+lazy_static! {
+    /// This OID identifies an extension containing a Nitro attestation report.
     pub static ref ATTESTATION_NITRO: ObjectIdentifier = vec![1, 3, 6, 1, 4, 1, 49690, 2, 2, 6].into();
-}
\ No newline at end of file
+    /// This OID identifies an extension containing a the PCR0 of a Nitro enclave;
```
vn971, line 12: small typo `containing a the`

raoulstrackx (commented):
I'm wondering if this is the best approach to differentiate debug enclaves. Can't we directly change the PCR0 value in the attestation document and re-sign it with our own custom debug (or even random) key?

em-app/src/platform/nitro.rs
```diff
@@ -29,38 +70,38 @@ pub(crate) fn get_remote_attestation_parameters(
...
     });
-    
+
     let csr_pem = get_csr_common_name(signer, &common_name, attributes, &extensions)?;
 
     Ok((None, None, csr_pem))
 }
 
+fn pcr0_from_env() -> Result<Vec<u8>> {
+    let value = std::env::var("ENCLAVEOS_PCR0").map_err(|_| Error::AttestationCertInternal("Missing 'ENCLAVEOS_PCR0' inside debug enclave.".into()))?;
```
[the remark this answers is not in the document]
jjfiv: An environment variable was chosen here because debug enclaves aren’t secure in the first place (RAM can be decrypted). By putting PCR0 we have an identity that can be used to determine policy in insecure accounts that allow debug enclaves.

I see your point about generating build-time ids instead, which we could pursue, but that would make deterministic builds a bit harder for ease of use and I’m not sure we want to make that tradeoff for people.

fortanix-vme/vme-pkix/src/oid.rs
jjfiv: OIDs are registered; this is a prefix owned by Fortanix; a slight modification of the one we’ve been using for Nitro attestations for a while. This should be added to published documentation, but there’s no formula that derived these; they’re just a sort of registration ID.

vn971: Can we add this very info to the documentation for `ATTESTATION_NITRO_PCR0`? Just state the fact that it's Fortanix's registration OID. I think this might help developers reading the code

em-app/src/platform/nitro.rs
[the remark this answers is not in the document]
jjfiv: Also, by only including it in a separate extension, clients (including us) can choose to never trust enclaves including this separate PCR0 extension.

jjfiv:
Rebased on master and addressed reviewer comments.

raoulstrackx:
I'm still not thrilled about using an environment variable. Especially not because we'll need to support the same mechanism in EDP for all platforms similar to AWS Nitro Enclaves. Unfortunately I don't see another option if deterministic builds are required and the PCR0 value must be identical to non-debug versions of the same enclave. Why is the latter a strong requirement?

PR #496, opened 2023-07-14
Initialize enclave args with the first argument

### LINKED ISSUE #386: EnclaveBuilder::build should call initialized_args_mut to ensure the first arg is set correctly.
See https://github.com/fortanix/rust-sgx/pull/280#discussion_r797209197 for context

mkaynov:
When calling the enclave main first argument must be always initialized

Resolves #386

### DIFF
--- a/intel-sgx/enclave-runner/src/loader.rs
+++ b/intel-sgx/enclave-runner/src/loader.rs
@@ -332,6 +332,7 @@ impl<'a> EnclaveBuilder<'a> {
     }
 
     pub fn build<T: Load>(mut self, loader: &mut T) -> Result<Command, Error> {
+        self.initialized_args_mut();
         let args = self.cmd_args.take().unwrap_or_default();
         let c = self.usercall_ext.take();
         self.load(loader)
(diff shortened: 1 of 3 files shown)


intel-sgx/enclave-runner/src/loader.rs
```diff
@@ -332,6 +332,7 @@ impl<'a> EnclaveBuilder<'a> {
     }
 
     pub fn build<T: Load>(mut self, loader: &mut T) -> Result<Command, Error> {
+        self.initialized_args_mut();
```
vn971, line 335: Looking at the the source code around, my thinking would be that the `build_library` function needs arguments being set as well, to avoid running a `main` function without arguments at all. Opinions @raoulstrackx @mkaynov ?

Alternatively, we could move the logic of a fallback signle argument to the `load` function. This way, the failsafe code would be at one place, instead of across these individual functions.

vn971 (approved):
Seems good. We could also move "adding the zero argument" logic to the `load` function for later execution. That could be neater, because the current logic of artificially pushing a *constant* value in the vector is a bit counter-intuitive, and it's not a big surprise that an "edge" case was missed and caused a bug. By always having it in `load`, we'd simplify the logical flow a little bit.

But the current solution also totally works.

LGTM.

raoulstrackx:
Looks good to me! Could you rebase it?

PR #498, opened 2023-07-17
Fix typo in `dcap-ql` dependency

### LINKED ISSUE #497: Published dcap-ql 0.3.6 is broken
It seems that the currently latest published version of `dcap-ql` (0.3.6) is broken due to it incorrectly specifying features for the `mbedtls` crate:
https://github.com/fortanix/rust-sgx/blob/cdaabde0c1e6ee1a442b79ae524f08a6a49197fd/intel-sgx/dcap-ql/Cargo.toml#L48

It should be `features` instead of `feature`. This was introduced in commit 4e8b46d1c6ab822b4fc6f698f8cd3f9bf7b61e3a.

Since the configuration disables default features of `mbedtls` this results in a compile error saying that neither the `std` nor the `no_std_deps` feature is set.

I would suggest yanking the broken dcap-ql 0.3.6 release.

raoulstrackx:
A typo in the specification of the `mbedtls` dependency of `dcap-ql` may lead to a compilation error.

Fixes #497
cc: @Taowyoo @jethrogb @vn971 @mkaynov

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -661,7 +661,7 @@ dependencies = [
 
 [[package]]
 name = "dcap-ql"
-version = "0.3.6"
+version = "0.3.7"
 dependencies = [
  "byteorder 1.3.4",
  "dcap-ql-sys",
--- a/intel-sgx/dcap-ql/Cargo.toml
+++ b/intel-sgx/dcap-ql/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "dcap-ql"
-version = "0.3.6"
+version = "0.3.7"
 authors = ["Fortanix, Inc."]
 license = "MPL-2.0"
 description = """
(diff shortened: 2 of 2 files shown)


mkaynov:
LGTM

kostko:
Would be cool if there was a build check as part of CI :)

raoulstrackx:
> Would be cool if there was a build check as part of CI :)

Yes I totally agree. Unfortunately, the way features get enabled in a workspace, makes it difficult. Did you have a test in mind that reproduced the compilation error?

Taowyoo (approved):

Taowyoo:
The `dcap-ql` already be included in CI.
 And the default `std` I added for `mbedtls` in `dcap-ql` is for avoiding the compiling error when compile the `dcap-ql` with only `verify` and with `mbedtls` specify with `default-features = false` in a crate:
```
[package]
name = "dcap-ql-fail"
version = "0.1.0"
edition = "2021"

# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html

[dependencies]
dcap-ql = { version = "0.3.6", default-features = false, features = ["verify"] }
mbedtls = { version = "0.9.0", default-features = false}
```
Because `mbedtls` needs at least one of `std` or `no_std_deps` features is defined.

Taowyoo:
Created an issue for the build check problem above : https://github.com/fortanix/rust-sgx/issues/499

PR #512, opened 2023-08-22
Added version of certificate request functions that accept Subject as parameter

dlegalov-fortanix:
In light of recent requirements to be able to request certificates with no Common Name (as this is deprecated), this change creates versions of several functions that accept a `Name`  as a parameter instead of a common name string. In this case, the user will be able to provide an empty `Name`, where the internal value is an empty `Vec`. In order to provide backwards compatibility, the existing functions that use common name have not been changed.

### DIFF
--- a/em-app/src/csr.rs
+++ b/em-app/src/csr.rs
@@ -4,20 +4,21 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
+use crate::Error;
+use crate::Result;
 use mbedtls::hash;
 use mbedtls::pk::{Pk, Type as PkType};
-use pkix::DerWrite;
+pub use mbedtls::rng::Rdrand as FtxRng;
+use pkix;
 use pkix::bit_vec::BitVec;
 use pkix::pem::{der_to_pem, PEM_CERTIFICATE_REQUEST};
 use pkix::pkcs10::{CertificationRequest, CertificationRequestInfo};
-use pkix::types::{Attribute, Extension, ObjectIdentifier, TaggedDerValue, RsaPkcs15, EcdsaX962, Sha256, DerSequence, Name};
+use pkix::types::{
+    Attribute, DerSequence, EcdsaX962, Extension, Name, ObjectIdentifier, RsaPkcs15, Sha256,
+    TaggedDerValue,
+};
+use pkix::DerWrite;
 use yasna::tags::TAG_UTF8STRING;
-use pkix;
-
-use crate::Error;
-
-pub use mbedtls::rng::Rdrand as FtxRng;
-use crate::Result;
 
 #[derive(Clone, Copy, Debug)]
 pub enum SignatureAlgorithm {
@@ -68,7 +69,7 @@ where
                 }.write(writer)
             })
         };
-        
+
         Ok(der_to_pem(&csr, PEM_CERTIFICATE_REQUEST))
     }
 }
@@ -77,11 +78,11 @@ impl ExternalKey for Pk {
     fn get_public_key_der(&mut self) -> Result<Vec<u8>> {
         Ok(self.write_public_der_vec().map_err(|e| Error::ExternalKey(Box::new(e)))?)
     }
-    
+
     fn sign_sha256(&mut self, input: &[u8]) -> Result<(Vec<u8>, SignatureAlgorithm)> {
         let mut hash = [0u8; 32];
         hash::Md::hash(hash::Type::Sha256, &input, &mut hash).map_err(|e| Error::ExternalKey(Box::new(e)))?;
-        
+
         let mut sig = vec![0u8; (self.len()+7)/8];
         self.sign(hash::Type::Sha256, &hash, &mut sig, &mut FtxRng).map_err(|e| Error::ExternalKey(Box::new(e)))?;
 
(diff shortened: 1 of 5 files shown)


em-app/src/platform/mod.rs
```diff
@@ -24,4 +28,13 @@ pub(crate) mod nitro;
           any(target_env="gnu",
               target_env="musl",
               target_env = "fortanixvme")))]
-pub(crate) use nitro::get_remote_attestation_parameters;
+pub(crate) use nitro::{get_remote_attestation_parameters, get_remote_attestation_parameters_subject};
+
+pub(crate) fn get_extensions_from_alt_names(alt_names: Option<Vec<Cow<str>>>) -> Option<Vec<(ObjectIdentifier, bool, Vec<u8>)>> {
```
dlegalov-fortanix, line 33: This was common code used in the `platform` module, so I opted to make it a function. I do not know if it would be preferable to put it somewhere like a `utils.rs`, but here it is in the module declaration file.

em-app/src/lib.rs
```diff
@@ -45,7 +46,28 @@ pub fn get_fortanix_em_certificate(url: &str, common_name: &str, signer: &mut dy
     get_certificate(url, common_name, signer, None, None)
 }
 
-pub fn get_certificate(url: &str, 
+pub fn get_certificate_subject(url: &str,
```
ssavvides, line 49: We should try to keep our code well-formatted as much as possible. To achive that we can use:

```bash
cd em-app

cargo fmt
# or
rustfmt src/lib.rs
```

`cargo fmt` will format the entire crate
`rustfmt src/lib.rs` will format the entire file.

You should _not_ push these changes since there might be many formatting changes in code you did not change/add, but you can use the result of these commands as a reference to format your changes.

For instance, when I format this file I see that the `get_certificate_subject` is formatted as so:

```rust
pub fn get_certificate_subject(
    url: &str,
    subject: &Name,
    signer: &mut dyn CsrSigner,
    alt_names: Option<Vec<Cow<str>>>,
    config_id: Option<&str>,
) -> Result<FortanixEmCertificate> {
    let pub_key = signer.get_public_key_der()?;
    let user_data = get_user_data(&pub_key, config_id)?;

    let (attestation_certificate_der, node_certificate_der, csr_pem) =
        platform::get_remote_attestation_parameters_subject(
            signer, url, subject, &user_data, alt_names,
        )?;

    let certificate_response = request_issue_certificate(url, csr_pem)?;

    Ok(FortanixEmCertificate {
        attestation_certificate_der,
        node_certificate_der,
        certificate_response,
    })
}
```

em-app/src/platform/sgx.rs
```diff
@@ -42,14 +42,10 @@ pub(crate) fn get_remote_attestation_parameters(
...
-    let extensions = alt_names.and_then(|names| {
-        Some(vec![(pkix::oid::subjectAltName.clone(), false, pkix::yasna::construct_der(|w| DnsAltNames { names }.write(w)).into())])
-    });
-
+    let extensions = get_extensions_from_alt_names(alt_names);
     let mut sgx_name = SgxName::from_report(&report, true);
-    sgx_name.append(vec![(pkix::oid::commonName.clone(), common_name.to_string().into())]);
-    let subject = sgx_name.to_name();
-    
+    let mut subject = subject.clone(); // Copy the subject, since it must be modified
```
ssavvides, line 47: consider passing a mut reference of `subject`

dlegalov-fortanix: The nitro code does not need mutablity, so in wanting to keep things as restrained as possible, I tried to avoid using `mut`. The signature of this function does already have `signer: &mut dyn CsrSigner,`, so there is a mutable parameter. Another option is to use a pass by value/copy. The advantage of a mutable parameter would be the user having the modified Subject available with all values, but I do not know if this adds much.

em-app/src/csr.rs
```diff
@@ -4,20 +4,21 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
+use crate::Error;
```
ssavvides, line 7: nit:

We haven't really set a rustfmt config for this repo (as far as I know), but for our internal repos we use the `StdExternalCrate` style for our imports (https://rust-lang.github.io/rustfmt/?version=v1.6.0&search=#group_imports)

em-app/src/csr.rs
```diff
@@ -90,24 +91,24 @@ impl ExternalKey for Pk {
...
 }
 
-pub fn get_csr_common_name(signer: &mut dyn CsrSigner,
-                           common_name: &str,
-                           attributes: Vec<(ObjectIdentifier, Vec<Vec<u8>>)>,
-                           extensions: &Option<Vec<(ObjectIdentifier, bool, Vec<u8>)>>
-) -> Result<String> {
-    let subject = vec![(pkix::oid::commonName.clone(), TaggedDerValue::from_tag_and_bytes(TAG_UTF8STRING, common_name.as_bytes().to_vec()))].into();
-    get_csr(signer, &subject, attributes, extensions)
+pub(crate) fn common_name_to_subject(common_name: &str) -> Name {
```
ssavvides, line 99: I think this function can be moved to lib.rs

ssavvides (approved):

em-app/src/csr.rs
dlegalov-fortanix: Moved to `lib.rs`

em-app/src/csr.rs
dlegalov-fortanix: Thank you for the explanation, I will keep that in mind for styling!

dlegalov-fortanix:
Rebased off of `raoul/gh-516-fix_musl_ci_issue` (https://github.com/fortanix/rust-sgx/pull/517) pending its merge to master.

raoulstrackx:
@dlegalov-fortanix #517 is merged now. You probably want to rebase again

dlegalov-fortanix:
> @dlegalov-fortanix #517 is merged now. You probably want to rebase again

Thank you, rebased and awaiting CI.

arai-fortanix:
bors r=ssavvides

PR #544, opened 2023-10-31
Adds a function to compute hash over the contents of application config

nshyrei:
This PR provides a functionality to compute a hash over the contents of application config.

### DIFF
--- a/em-app/src/lib.rs
+++ b/em-app/src/lib.rs
@@ -26,6 +26,8 @@ pub mod error;
 pub use error::*;
 use yasna::models::TaggedDerValue;
 use yasna::tags::TAG_UTF8STRING;
+use sdkms::api_model::Blob;
+use std::collections::{BTreeMap, BTreeSet};
 
 type Result<T> = std::result::Result<T, Error>;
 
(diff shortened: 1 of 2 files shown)


arai-fortanix (commented):
Why is this function useful? Won't clients receive the json string from CCM and just want to hash that directly, instead of decoding it and then re-encoding it to JSON?

em-app/src/lib.rs
nshyrei, line 1: @arai-fortanix This is a copy of function with the same body from CCM backend. The code copy was done after a suggestion from John Foley on how to share the hash computation logic between CCM and Salmiac. Do you think that it's better to just use the same MD5 function in Salmiac and not expose the whole function here?

arai-fortanix: I'm trying to understand the intended use for this function. I think it's reasonable to have a convenience function for hashing the config JSON, since maybe we want to change how it's hashed in the future. But it seems to me that this shared function should operate on the JSON string and not on some data-structure-driven definition of the configuration. The client is going to get a config JSON string and may or may not be interested in using the same internal representation of the decoded config that CCM uses.

nshyrei: Changed the function to accept only string

em-app/src/lib.rs
```diff
@@ -138,6 +139,14 @@ pub fn request_issue_certificate(url: &str, csr_pem: String) -> Result<models::I
     client.issue_certificate(request).map_err(|e| Error::NodeAgentClient(Box::new(e)))
 }
 
+/// Computes application configuration hash from raw string
+pub fn compute_app_config_hash(app_config: &str) -> Result<Blob> {
+    let mut digest = vec![0; 32];
+    hash::Md::hash(hash::Type::Sha256, app_config.as_bytes(), &mut digest).map_err(|e| Error::TargetReportHash(Box::new(e)))?;
```
aditijannu, line 145: Should we also pass hash type to the function in addition to the app_config string? Do we always use sha256 hashes only?

aditijannu (approved):

em-app/src/lib.rs
nshyrei: changed

arai-fortanix (approved):

raoulstrackx:
bors r=arai-fortanix,aditijannu

PR #546, opened 2023-11-20
Ignore the suffix '- <unknown>' in stack-trace-resolve script

mzohreva:


raoulstrackx (commented):
I think the changes are correct, but the example in `examples/backtrace_panic` still doesn't work for me (even after updating the path in `test.sh`). Is it expected to work?

PR #566, opened 2024-03-15
Added a stress test for memory allocation (snmalloc)

NirjharRoyiitk:
This is a very basic stress test for memory allocator

Description of the test:

This thread creates 64 threads of 3 types:
1. Large threads allocates buffers of size 1GB to 4GB and scans them page wise
2. Medium threads allocates buffers of size 8MB to 128MB and scans them with an interval of 1MB
3. Small threads allocates buffers of size 1KB to 512KB and scans them with an interval of 1KB

After scanning the buffers every thread frees the buffer. 
In the end it calculates the time taken to do all the operations (list from step 1 to step 3). 
The test periodically prints the average operation time of each thread


examples/mem-alloc-test/src/main.rs
```diff
@@ -0,0 +1,152 @@
...
+    loop {
+        let mut page_size = 4096;
+        let to_kb = 1024;
+        let to_mb = to_kb * 1024;
+        let to_gb = to_mb * 1024;
+
+        let mut ran_num = 0;
+        let mut size = 0;
+        let mut scan_interval = page_size;
+        let memtypestr;
```
arai-fortanix, line 59: The rust style for this would be to have each arm of the `match` produce the values you need (as a tuple, since this is producing multiple values). So something like:

```
let (scan_interval, size, memtypestr) = match memtype {
    SizeType::LARGE_MEM => {
         (Get_Random_Num(1, 4) * to_gb, String::from("LARGE"))
    },
    SizeType::MEDIUM_MEM => {
    }
   ...
};
```

Basically in Rust, things like `match` and `if` can evaluate to a value, so you can do things like `let x = if ...` or `let y = match ...`. This form is generally preferred over assigning a variable in each arm of the `match` or `if`. Rust also has native support for tuples.

Also, I think there's a better way of getting a string from the enumeration names than having explicit code here that does it. In this particular case, it looks like we only care about the strings when we're printing the output. In that case, we can just use the enumeration value everyewhere. You can add `#[derive(Debug)]` on the line before your `enum SizeType` definition. This tells the Rust compiler to automatically generate code for a debug representation of the type. In the case of enums, this is just the name of the Enum variant. When you print the results, if you print using `"{:?}"` instead of just `"{}", it will print using the debug representation. So something like ```fun myfun(var: SizeType) { println!("{:?}", var); }``` will just do what you want, without having to write explicit conversions to string for your enumeration.

arai-fortanix: If you do need the string representation at run-time for other reasons, then the Rust-y way to do it would be to add a `to_str()` method to the enumeration, so the conversion is tied to the enumeration type. Rust doesn't have classes and objects, but you can add functions to enums and structs that operate somewhat like methods on classes in object-oriented language.

examples/mem-alloc-test/src/main.rs
```diff
@@ -0,0 +1,152 @@
...
+
+fn Get_Random_Num(start: usize, end: usize) -> usize {
+    let mut rng = rand::thread_rng();
+    let random_number: usize = rng.gen_range(start..=end);
+    return random_number;
+}
+
+fn worker_thread(tid: i32, shared_mutex_clone: Arc<Mutex<i32>>,
+                 shared_mutex_dur_clone: Arc<Mutex<u128>>, memtype: SizeType) {
+    loop {
```
arai-fortanix, line 50: So this loop continues forever? As a future improvement, it may make sense to run for either a specified number of iterations or a specified amount of time. Also as a future improvement, it can take a long time for all of the threads to start up. So you might want to consider having all of the threads wait on something before they start doing work, and only get signalled to start once all of the threads have been created.

arai-fortanix (approved):

examples/mem-alloc-test/src/main.rs
```diff
@@ -0,0 +1,275 @@
...
+    rng.gen_range(start..=end)
+}
+
+fn wait_per_thread(pair_clone: Arc<(Mutex<bool>, Condvar)>) {
+    let (lock, cvar) = &*pair_clone;
+    let mut started = lock.lock().unwrap();
+    while !*started {
+        started = cvar.wait(started).unwrap();
+    }
+    drop(started);
```
arai-fortanix, line 99: I'm not sure if an explicit `drop()` is needed here. Usually locks are scoped, so they'll be automatically dropped when they leave scope, but maybe there's a reason why the drop is needed with the condition variable.

examples/mem-alloc-test/src/main.rs
```diff
@@ -0,0 +1,275 @@
...
+}
+
+fn wakeup_all_child_threads(pair_clone: Arc<(Mutex<bool>, Condvar)>) {
+    let (lock, cvar) = &*pair_clone;
+    let mut started = lock.lock().unwrap();
+    *started = true;
+    drop(started);
```
arai-fortanix, line 106: I don't think it's wrong to drop the lock before notifying the condition variable, but I think it's safe to do the notify with the lock still held and then just let the lock be dropped when the function returns.

examples/mem-alloc-test/src/main.rs
```diff
@@ -0,0 +1,275 @@
...
+    let shared_mutex_clone_2 = Arc::clone(&shared_mutex_clone);
+    calculate_and_print_stat(shared_mutex_clone_2, tid, &memsize, avg_thread_latency);
+
+
+}
+
+fn spawn_threads(thread_count: i32) {
+    let mut handles = vec![];
+    let shared_variable = Arc::new(Mutex::new(Counters{
+            avg_duration_large_thread:0.0,
```
arai-fortanix, line 195: Minor: I think the preferred formatting for these is like:
```
avg_duration_large_thread: 0.0,
```
(with no space before the `:` but one space after the `:`. I think this is one of things that `cargo fmt` will format for you.

examples/mem-alloc-test/src/main.rs
```diff
@@ -0,0 +1,275 @@
...
+    data.avg_duration_medium_thread = data.avg_duration_medium_thread / num_medium_threads as f64;
+    data.avg_duration_small_thread = data.avg_duration_small_thread / num_small_threads as f64;
+    data.global_average =
+        (data.avg_duration_small_thread + data.avg_duration_medium_thread + data.avg_duration_large_thread) /
+        (num_large_threads + num_medium_threads + num_small_threads ) as f64;
+    println!("{},{},{},{},{}", thread_count, data.avg_duration_small_thread,
+             data.avg_duration_medium_thread,
+             data.avg_duration_large_thread,
+             data.global_average);
+
```
arai-fortanix, line 250: Trivial: A couple of unnecessary blank lines here.

examples/mem-alloc-test/src/main.rs
```diff
@@ -0,0 +1,275 @@
...
+}
+
+/* If there are n processors available, we will record the numbers with,2n threads,
+ * then n threads, then n/2 and so on.
+ */
+fn start_tests()
+{
+    println!("NUM_THREADS,LATENCY_SMALL_THREADS,LATENCY_MEDIUM_THREADS,LATENCY_LARGE_THREADS,GLOBAL_AVERAGE");
+    let mut num_processors = get_num_processors();
+    num_processors = 4;
```
arai-fortanix, line 265: `num_processors` is being hard-coded to 4, even though we called `get_num_processors()` on the previous line. Was this intentional?

arai-fortanix (approved):
New changes look good to me.

examples/mem-alloc-test/src/main.rs
```diff
@@ -0,0 +1,275 @@
...
+    println!("{},{},{},{},{}", thread_count, data.avg_duration_small_thread,
+             data.avg_duration_medium_thread,
+             data.avg_duration_large_thread,
+             data.global_average);
+
+
+}
+
+fn get_num_processors() -> usize {
+    num_cpus::get()
```
raoulstrackx, line 255: This won't work in SGX as it will attempt to open files from within the enclave or issue syscalls. You'll need to create a service in the enclave runner if you want to do something like that. See [this example](https://github.com/mzohreva/edp-examples/tree/main/http-server-async)

examples/mem-alloc-test/Cargo.toml
```diff
@@ -0,0 +1,9 @@
+[package]
```
raoulstrackx, line 1: This crate isn't tested in CI. Can you add it?

examples/mem-alloc-test/src/main.rs
NirjharRoyiitk: No it isn't. I will remove this.

examples/mem-alloc-test/Cargo.toml
NirjharRoyiitk: How do I add to CI?

arai-fortanix (approved):

examples/mem-alloc-test/src/main.rs
```diff
@@ -0,0 +1,293 @@
...
+        }
+    };
+}
+
+fn get_random_num(start: usize, end: usize) -> usize {
+    let mut rng = rand::thread_rng();
+    rng.gen_range(start..=end)
+}
+
+fn wait_per_thread(pair_clone: Arc<(Mutex<bool>, Condvar)>) {
```
raoulstrackx, line 89: You can simply this by using a [Barrier](https://doc.rust-lang.org/std/sync/struct.Barrier.html)

examples/mem-alloc-test/src/main.rs
```diff
@@ -0,0 +1,293 @@
...
+            MemSize::Medium => {
+                /* buffer size will be from 8MB to 128 */
+                (
+                    SCAN_INTERVAL_MEDIUM_THREAD,
+                    TO_MB * get_random_num(MEDIUM_THREAD_MEM_START, MEDIUM_THREAD_MEM_END),
+                    LIMIT_MEDIUM_THREAD,
+                )
+            }
+            MemSize::Small => {
+                /* buffer size will be from 1KB to 512KB */
```
raoulstrackx, line 157: For a heap allocator 1KB is still big. Can this start from 1B instead? Corner cases may only pop up with such tiny chunks

examples/mem-alloc-test/src/main.rs
```diff
@@ -0,0 +1,293 @@
...
+        drop(large_vector);
+
+        /* calculate the metrics */
+        let end_time = Instant::now();
+        let duration = end_time.duration_since(start_time);
+        tot_time_ns += duration.as_nanos();
+
+        count = count + 1;
+        if (count >= limit) {
+            break;
```
raoulstrackx, line 185: If you want to do something like this, you can write the loop as:
```
for _it in 0..count {
   ...
}
```

examples/mem-alloc-test/src/main.rs
```diff
@@ -0,0 +1,293 @@
...
+     * special service to make it work properly.
+     */
+    NUM_CPUS
+}
+
+/* If there are n processors available, we will record the numbers with,2n threads,
+ * then n threads, then n/2 and so on.
+ */
+fn start_tests() {
+    println!("NUM_THREADS,LATENCY_SMALL_THREADS,LATENCY_MEDIUM_THREADS,LATENCY_LARGE_THREADS,GLOBAL_AVERAGE");
```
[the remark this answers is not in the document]
NirjharRoyiitk: I am trying to generate a csv format file  and this is just the heading

PR #569, opened 2024-03-25
Adding a test that will test the correctness of snmalloc

NirjharRoyiitk:
This is a test that tests the correctness of our new allocator snmalloc. This test should run continuosly and should not crash. 

The pseudocode is taken from https://fortanix.atlassian.net/wiki/spaces/PLAT/pages/3155820642/Memory+Allocators

We should modify the configurable parameters https://github.com/fortanix/rust-sgx/blob/memory-correctness-test/examples/mem-correctness-test/src/main.rs#L43 and also modify the ```heap-size```  parameter in ```rust-sgx/examples/mem-correctness-test/Cargo.toml``` file then simply do the following

``` cd rust-sgx/examples/mem-correctness-test/
    cargo run --target x86_64-fortanix-unknown-sgx
```

### DIFF
--- a/examples/mem-correctness-test/Cargo.toml
+++ b/examples/mem-correctness-test/Cargo.toml
@@ -1,0 +1,17 @@
+[package]
+name = "mem-correctness-test"
+version = "0.1.0"
+edition = "2021"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+crossbeam = "0.8.0"
+rand = "0.8.4"
+num_cpus = "1.14.0"
+sha2 = "0.10"
+
+[package.metadata.fortanix-sgx]
+# heap size (in bytes), the default heap size is 0x2000000.
+heap-size=0x20000000
+debug=false
--- a/examples/mem-correctness-test/src/main.rs
+++ b/examples/mem-correctness-test/src/main.rs
@@ -1,0 +1,286 @@
+/* This test based on the following pseudo code and tests correctness of our
+ * new allocator snmalloc
+
+   for 0..num_threads {
+    loop {
+        let mut regions: Vec<(Box<[u8], u8>)>;
+        let mut mem_used = 0u64;
+        match rnd % 4 {
+           0 => // Check area
+           1..2 => // Alloc random area if less than x% of the heap is used (not 100% to
+                   // account for fragmentation, ...) and check area.
+           3 => // Free random area
+        }
+    }
+}
+ * So this basically runs for a long time and should never crash
+ */
+
+use core::arch::asm;
+use rand::Rng;
+use sha2::{Digest, Sha256};
+use std::alloc::{alloc, dealloc, Layout};
+use std::ptr;
+use std::slice;
+use std::sync::atomic::{AtomicUsize, Ordering};
+use std::sync::Barrier;
+use std::sync::{Arc};
+use std::thread;
+
+#[cfg(target_env = "sgx")]
+extern "C" {
+    static HEAP_BASE: u64;
+    static HEAP_SIZE: usize;
+}
+
+const PAGE_SIZE: usize = 4096;
+const TO_KB: usize = 1024;
+const TO_MB: usize = TO_KB * 1024;
+const TO_GB: usize = TO_MB * 1024;
+const ALIGN: usize = PAGE_SIZE;
+
+const NUM_OPERATION_CHOICES: usize = 4;
+static HEAP_ALLOCATED: AtomicUsize = AtomicUsize::new(0);
+
+/* Set of configurable parameters. These will adjusted as necessary while
+ * recording the performance numbers. Since this test will be in CI, I have
+ * kept the parameters to be less memory intensive.
+ */
+const NUM_THREADS: usize = 2;
+/* PER_THREAD_PER_BUFFER_MAX_SIZE is max size of a buffer that a thread can allocate
+ * on each call to add_new_buffer() function. Higher the value, more will be
+ * the total memory consumption and more time taken by the test.
+ */
+const PER_THREAD_PER_BUFFER_MAX_SIZE: usize = 4 * TO_KB;
+/* MAX_BUFFER_CHECKS_PER_THREAD_ITERATION is the maximum number of buffers to
+ * check on each call to select_and_check_random_buffer_contents()
+ */
+const MAX_BUFFER_CHECKS_PER_THREAD_ITERATION: usize = 4;
+/* MAX_INDEX_CHECKS_PER_BUFFER is the number of indices/locations to check
+ * per buffer.
+ */
+const MIN_ALLOWED_FREE_HEAP_PERCENTAGE: f64 = 10.0;
+/* MAX_ITERATIONS_PER_THREAD is the number of operations per thread (1 operation per
+ * per iteration)
+ */
+const MAX_ITERATIONS_PER_THREAD: usize = 4;
+
+#[cfg(target_env = "sgx")]
+#[inline(always)]
+pub fn image_base() -> u64 {
+    let base: u64;
+    unsafe {
+        asm!(
+            "lea IMAGE_BASE(%rip), {}",
+            lateout(reg) base,
+            options(att_syntax, nostack, preserves_flags, nomem, pure),
+        )
+    };
+    base
+}
+
+#[cfg(target_env = "sgx")]
+#[inline(always)]
+pub(crate) unsafe fn rel_ptr_mut<T>(offset: u64) -> *mut T {
+    (image_base() + offset) as *mut T
+}
+
+/* Returns the base memory address of the heap */
+#[cfg(target_env = "sgx")]
+pub(crate) fn heap_base() -> *const u8 {
+    unsafe { rel_ptr_mut(HEAP_BASE) }
+
+}
+
+/* Returns the size of the heap */
+pub(crate) fn heap_size() -> usize {
+    #[cfg(target_env = "sgx")]
+    unsafe { HEAP_SIZE }
+    #[cfg(not(target_env = "sgx"))]
+    usize::MAX
+}
+
+fn update_occupied_heap_size_on_delete(size: usize) {
+    HEAP_ALLOCATED.fetch_sub(size, Ordering::SeqCst);
+}
+
+fn update_occupied_heap_size_on_addition(size: usize) {
+    HEAP_ALLOCATED.fetch_add(size, Ordering::SeqCst);
+}
+
+fn get_occupied_heap_size() -> usize {
+    HEAP_ALLOCATED.load(Ordering::SeqCst)
+}
+
+fn get_free_heap_size_in_bytes() -> usize {
+    heap_size() - get_occupied_heap_size()
+}
+
+fn get_free_heap_percentage() -> f64 {
+    (get_free_heap_size_in_bytes() as f64 / heap_size() as f64) * 100.0
+}
+
+fn get_random_num(start: usize, end: usize) -> usize {
+    let mut rng = rand::thread_rng();
+    rng.gen_range(start..=end)
+}
+
+fn wait_per_thread(barrier_clone: Arc<Barrier>) {
+    barrier_clone.wait();
+}
+
+fn traverse_and_check_buffer(buf: &(*mut u8, usize, String)) -> bool {
+    if buf.2 != compute_sha256_hex(buf.0, buf.1) {
+        return false;
+    }
+    return true;
+}
+
+fn select_and_check_random_buffer_contents(array_of_vectors: &Vec<(*mut u8, usize, String)>) {
+    /* This function selects a random number of buffers and then calls
+     * traverse_and_check_buffer which checks random contents of the set of
+     * randomly selected buffers
+     */
+    let num_active_vectors = array_of_vectors.len();
+    if num_active_vectors > 0 {
+        let random_buffer_check_count = get_random_num(1, MAX_BUFFER_CHECKS_PER_THREAD_ITERATION);
+        for _i in 1..=random_buffer_check_count {
+            let random_buffer_index_to_check = get_random_num(0, array_of_vectors.len() - 1);
+            assert!(traverse_and_check_buffer(
+                &(array_of_vectors[random_buffer_index_to_check]),
+            ));
+        }
+    }
+}
+
+fn delete_random_buffer(array_of_vectors: &mut Vec<(*mut u8, usize, String)>) {
+    let num_active_vectors = array_of_vectors.len();
+    if num_active_vectors > 0 {
+        let random_index_to_delete = get_random_num(0, array_of_vectors.len() - 1);
+        let len = array_of_vectors[random_index_to_delete].1;
+
+        unsafe {
+            dealloc(
+                array_of_vectors[random_index_to_delete].0,
+                Layout::from_size_align(len, ALIGN).unwrap(),
+            );
+        }
+        array_of_vectors.remove(random_index_to_delete);
+        update_occupied_heap_size_on_delete(len);
+    }
+}
+
+fn add_new_buffer(array_of_vectors: &mut Vec<(*mut u8, usize, String)>) -> bool {
+    /* This function assumes that the percentage of free heap space is more
+     * than MIN_ALLOWED_FREE_HEAP_PERCENTAGE
+     */
+
+    let random_size = get_random_num(1, PER_THREAD_PER_BUFFER_MAX_SIZE);
+
+    // Create a layout based on the size and alignment
+    let layout = Layout::from_size_align(random_size, ALIGN).unwrap();
+
+    // Allocate memory using the global allocator
+    let ptr = unsafe { alloc(layout) };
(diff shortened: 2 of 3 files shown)


examples/mem-correctness-test/Cargo.toml
```diff
@@ -0,0 +1,15 @@
...
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+rand = "0.8.4"
+num_cpus = "1.14.0"
+
+[package.metadata.fortanix-sgx]
+# heap size (in bytes), the default heap size is 0x2000000.
+heap-size=0x800000000
+debug=false
```
arai-fortanix, line 15: Is there a reason we're setting `debug=false` here?

examples/mem-correctness-test/src/main.rs
```diff
@@ -0,0 +1,267 @@
...
+    static HEAP_SIZE: usize;
+}
+
+const PAGE_SIZE: usize = 4096;
+const TO_KB: usize = 1024;
+const TO_MB: usize = TO_KB * 1024;
+const TO_GB: usize = TO_MB * 1024;
+
+const NUM_OPERATION_CHOICES: usize = 4;
+const MAGIC_NUM_TO_CHECK: u32 = 0xdeadbeef;
```
arai-fortanix, line 40: I have a slight preference for not using `0xdeadbeef` as a magic number. This is a popular choice for magic number, so if we're using 3rd party libraries, there's a slightly higher chance of this particular magic number being used by something else, compared to picking some different magic number. Also, 32 bits is pretty small if it's important that we don't encounter the magic number on accident. 64 bits would be better, if that's possible.

Both of these comments are very minor.

examples/mem-correctness-test/src/main.rs
```diff
@@ -0,0 +1,267 @@
...
+fn traverse_and_check_buffer(buf: &Vec<u32>) {
+    let mut ptr = 0;
+    let num_indices_checks = get_random_num(1, MAX_INDEX_CHECKS_PER_BUFFER);
+    for i in 1..=num_indices_checks {
+        /* Check for random indices and number of such indices is  num_indices_checks
+         * Please note that depending on the number random number generator, we
+         * can check for the same index multiple times. We could have checked
+         * for all the indices but that would be too time consuming
+         */
+        if (buf[get_random_num(0, buf.len() - 1)] != MAGIC_NUM_TO_CHECK) {
```
arai-fortanix, line 133: Maybe as a future improvement, it may be a good idea to fill the buffer with pseudo-random data that can be easily checked, rather than by repeating the same bytes over and over. I've worked on memory allocators in the past that had bugs that manifested as data being copied to the wrong location in memory. These bugs were not caught by a test similar to this one that repeated the same magic number repeatedly, because all of the magic numbers were the same. When we changed the test to use pseudo-random (but calculatable) sequences, the test was able to catch the bug.

examples/mem-correctness-test/src/main.rs
```diff
@@ -0,0 +1,267 @@
...
+            panic!("Corruption detected in traverse_and_check_buffer");
+        }
+    }
+}
+
+fn select_and_check_random_buffer_contents(array_of_vectors: &Vec<Vec<u32>>) {
+    /* This buffer selects a random number of buffers and then calls
```
arai-fortanix, line 140: Should the comment say "This function selects a random number..." instead of "This buffer selects..."?

examples/mem-correctness-test/src/main.rs
```diff
@@ -0,0 +1,267 @@
...
+     * traverse_and_check_buffer which checks random contents of the set of
+     * randomly selected buffers
+     */
+    let num_active_vectors = array_of_vectors.len();
+    if (num_active_vectors > 0) {
+        let random_buffer_check_count =
+            get_random_num(1, 100) % MAX_BUFFER_CHECKS_PER_THREAD_ITERATION + 1;
```
arai-fortanix, line 147: Is there a reason to do this calculation this way instead of just calling `get_random_num(1, MAX_BUFFER_CHECKS_PER_THREAD_ITERATION)`?

examples/mem-correctness-test/src/main.rs
```diff
@@ -0,0 +1,267 @@
...
+        for i in 1..=random_buffer_check_count {
+            let random_buffer_index_to_check = get_random_num(1, 100) % array_of_vectors.len();
```
arai-fortanix, line 149: Use `get_random_num(0, array_of_vectors.len() - 1)`?

examples/mem-correctness-test/src/main.rs
```diff
@@ -0,0 +1,267 @@
...
+    }
+}
+
+fn delete_random_buffer(array_of_vectors: &mut Vec<Vec<u32>>) {
+    let num_active_vectors = array_of_vectors.len();
+    if (num_active_vectors > 0) {
+        let random_index_to_delete = get_random_num(1, 100) % array_of_vectors.len();
+        let len = &array_of_vectors[random_index_to_delete].len() * mem::size_of::<i32>();
+        drop(&array_of_vectors[random_index_to_delete]);
+        array_of_vectors.remove(random_index_to_delete);
```
arai-fortanix, line 161: Does removing the vec from `array_of_vectors` not cause the vec to be dropped?

examples/mem-correctness-test/src/main.rs
```diff
@@ -0,0 +1,267 @@
...
+
+/* Set of configurable parameters. These will adjusted as necessary while
+ * recording the performance numbers
+ */
+const NUM_CPUS: usize = 64;
+const PER_THREAD_PER_BUFFER_MAX_SIZE: usize = (4 * TO_MB);
+const MAX_BUFF_NUM_PER_THREAD: usize = (32);
+const MAX_BUFFER_CHECKS_PER_THREAD_ITERATION: usize = 8;
+const MAX_INDEX_CHECKS_PER_BUFFER: usize = 1 << 16;
+const MAX_ALLOWED_FREE_HEAP_PERCENTAGE: f64 = 80.0;
```
arai-fortanix, line 51: Minor: Maybe this should be named `MAX_ALLOWED_USED_HEAP_PERCENTAGE`?

arai-fortanix (dismissed):
I don't know how Rust vecs interact with the underlying heap allocator, but as a future improvement I think it probably makes sense to exercise `resize()` on existing vecs to grow and shrink them, in addition to allocating and freeing vecs.

It might also be nice to have command-line arguments or environment variables or some other mechanism for controlling the number of threads and number of iterations without having to recompile the code.

examples/mem-correctness-test/src/main.rs
```diff
@@ -0,0 +1,267 @@
...
+    static HEAP_BASE: u64;
+    static HEAP_SIZE: usize;
+}
+
+const PAGE_SIZE: usize = 4096;
+const TO_KB: usize = 1024;
+const TO_MB: usize = TO_KB * 1024;
+const TO_GB: usize = TO_MB * 1024;
+
+const NUM_OPERATION_CHOICES: usize = 4;
```
BobbyAtFortanix, line 39: why don't you create an array of tester functions: fn traverse_and_check_buffer(buf: &Vec<u32>)
and in the worker thread we just go by the number of elements in the array. 
this way we can:
- expand the test variety by adding more operations
- change the test type weight/ratio by adding more functions of the same type
- remove test types, for debug purposes or when they become obsolete, or for whatever other reason
- other fun stuff

examples/mem-correctness-test/src/main.rs
```diff
@@ -0,0 +1,271 @@
...
+static heap_allocated: AtomicUsize = AtomicUsize::new(0);
+
+/* Set of configurable parameters. These will adjusted as necessary while
+ * recording the performance numbers
+ */
+const NUM_CPUS: usize = 64;
```
aditijannu, line 44: Would be nice to have an explanation or a one line comment about each of these parameters.

examples/mem-correctness-test/src/main.rs
```diff
@@ -0,0 +1,272 @@
...
+
+fn wakeup_all_child_threads(pair_clone: Arc<(Mutex<bool>, Condvar)>) {
+    let (lock, cvar) = &*pair_clone;
+    let mut started = lock.lock().unwrap();
+    *started = true;
+    cvar.notify_all();
+    drop(started);
+}
+
+fn traverse_and_check_buffer(buf: &Vec<usize>, MAGIC_NUM_TO_CHECK: usize) {
```
aditijannu, line 123: Change MAGIC_NUM_TO_CHECK to snake case as it is a dynamic value

examples/mem-correctness-test/src/main.rs
aditijannu, line 1: You can compile this program and address the warnings generated by the rust compiler, I see a lot of unused variables or improvements that can be easily addressed:
```
warning: unused import: `num_cpus`
  --> src/main.rs:20:5
   |
20 | use num_cpus;
   |     ^^^^^^^^
   |
   = note: `#[warn(unused_imports)]` on by default

warning: unused import: `std::time::Instant`
  --> src/main.rs:26:5
   |
26 | use std::time::Instant;
   |     ^^^^^^^^^^^^^^^^^^

warning: unnecessary parentheses around assigned value
  --> src/main.rs:45:47
   |
45 | const PER_THREAD_PER_BUFFER_MAX_SIZE: usize = (4 * TO_KB);
   |                                               ^         ^
   |
   = note: `#[warn(unused_parens)]` on by default
help: remove these parentheses
   |
45 - const PER_THREAD_PER_BUFFER_MAX_SIZE: usize = (4 * TO_KB);
45 + const PER_THREAD_PER_BUFFER_MAX_SIZE: usize = 4 * TO_KB;
   |

```

arai-fortanix: You can also turn on a setting that makes warning fail the build.

examples/mem-correctness-test/src/main.rs
```diff
@@ -0,0 +1,272 @@
...
+     * special service to make it work properly.
+     */
+    NUM_CPUS
+}
+
+/* If there are n processors available, we will record the numbers with,2n threads,
+ * then n threads, then n/2 and so on.
+ */
+fn start_tests() {
+    let MAGIC_NUM_TO_CHECK = get_random_num(1, usize::MAX);
```
arai-fortanix, line 263: My intention with better randomization of the fill data was that every element of every vec should get a distinct value. I think that's fine as a future enhancement.

examples/mem-correctness-test/src/main.rs
arai-fortanix: I like this idea (as a future enhancement), but I would add that the usual way to have different weights for different functions is not to duplicate their entries in the table but rather to have an explicit weight associated with each table entry, and then randomly choose an activity taking the weights into account.

arai-fortanix (approved):

examples/mem-correctness-test/src/main.rs
```diff
@@ -0,0 +1,279 @@
...
+    HEAP_ALLOCATED.load(Ordering::SeqCst)
+}
+
+fn get_free_heap_size_in_bytes() -> usize {
+    heap_size() - get_occupied_heap_size()
+}
+
+fn get_free_heap_percentage() -> f64 {
+    (get_free_heap_size_in_bytes() as f64 / heap_size() as f64) * 100.0
+}
```
raoulstrackx, line 111: This and the functions above could be logically grouped in `Heap` struct. That'd also simplify their names. You could do the same for allocating, freeing and checking allocated chunks.

examples/mem-correctness-test/src/main.rs
```diff
@@ -0,0 +1,279 @@
...
+fn wait_per_thread(pair_clone: Arc<(Mutex<bool>, Condvar)>) {
+    let (lock, cvar) = &*pair_clone;
+    let mut started = lock.lock().unwrap();
+    while !*started {
+        started = cvar.wait(started).unwrap();
+    }
+    drop(started);
+}
+
+fn wakeup_all_child_threads(pair_clone: Arc<(Mutex<bool>, Condvar)>) {
```
raoulstrackx, line 127: You can use a `Barrier` here

examples/mem-correctness-test/src/main.rs
```diff
@@ -0,0 +1,279 @@
...
+    let (lock, cvar) = &*pair_clone;
+    let mut started = lock.lock().unwrap();
+    *started = true;
+    cvar.notify_all();
+    drop(started);
+}
+
+fn traverse_and_check_buffer(buf: &Vec<usize>, magic_num_to_check: usize) {
+    let num_indices_checks = get_random_num(1, MAX_INDEX_CHECKS_PER_BUFFER);
```
raoulstrackx, line 136: I think you have the same issue here as in #566: You allocate a `Vec` assuming each element is a byte, which is not the case

examples/mem-correctness-test/src/main.rs
```diff
@@ -0,0 +1,279 @@
...
+    for _i in 1..=num_indices_checks {
+        /* Check for random indices and number of such indices is  num_indices_checks
+         * Please note that depending on the number random number generator, we
+         * can check for the same index multiple times. We could have checked
+         * for all the indices but that would be too time consuming
+         */
+        if buf[get_random_num(0, buf.len() - 1)] != magic_num_to_check {
```
raoulstrackx, line 143: I liked Dan's comment on magic numbers. You could even calculate a checksum instead

examples/mem-correctness-test/src/main.rs
```diff
@@ -0,0 +1,279 @@
...
+    /* Start all the threads */
+    wakeup_all_child_threads(pair);
+
+    /* Wait for all threads to finish */
+    for handle in handles {
+        handle.join().unwrap();
+    }
+}
+
+fn get_num_processors() -> usize {
```
raoulstrackx, line 258: This is confusing as `get_num_processors` doesn't return the number of processors on the machine

examples/mem-correctness-test/src/main.rs
```diff
@@ -0,0 +1,279 @@
...
+        update_occupied_heap_size_on_delete(len);
+    }
+}
+
+fn add_new_buffer(array_of_vectors: &mut Vec<Vec<usize>>, magic_num_to_check: usize) {
+    /* This function assumes that the percentage of free heap space is more
+     * than MIN_ALLOWED_FREE_HEAP_PERCENTAGE
+     */
+
+    let random_size = get_random_num(mem::size_of::<usize>() * 2, PER_THREAD_PER_BUFFER_MAX_SIZE)
```
raoulstrackx, line 185: Can the minimum be 1 byte?

examples/mem-correctness-test/src/main.rs
```diff
@@ -0,0 +1,279 @@
...
+        / mem::size_of::<usize>();
+    let mut tmp: Vec<usize> = Vec::with_capacity(random_size);
+    tmp.resize(random_size, magic_num_to_check);
```
raoulstrackx, line 188: Can you interact with the memory allocator directly? `Vec` may internally do some optimizations. For example, allocating more than requested. That'd mess up the stress test for the allocator

examples/mem-correctness-test/src/main.rs
NirjharRoyiitk: As an optimization I will fill up this function with so that it returns number of processors. Currently it returns a hardcoded value

examples/mem-correctness-test/src/main.rs
NirjharRoyiitk: I initially tried with 1 byte but since the vector is of of u64, I kept the minimum size to be 16 bytes

examples/mem-correctness-test/src/main.rs
NirjharRoyiitk: I haven't tried interacting with the allocator directly. Do you want me to try this in this PR? @raoulstrackx

raoulstrackx: Yes, please

examples/mem-correctness-test/src/main.rs
raoulstrackx: Yes but if it's hardcoded, there's no need to have this function

examples/mem-correctness-test/src/main.rs
```diff
@@ -0,0 +1,279 @@
...
+use rand::Rng;
+use std::mem;
+use std::sync::atomic::{AtomicUsize, Ordering};
+use std::sync::{Arc, Condvar, Mutex};
+use std::thread;
+
+extern "C" {
+    static HEAP_BASE: u64;
+    static HEAP_SIZE: usize;
+}
```
raoulstrackx, line 29: Pretty much all this code could also be executed on almost all targets the Rust compiler supports, except some statements/functions like these. You could restrict these to only the SGX target with:
```
#[cfg(target_env = "sgx")]
extern "C" {
    static HEAP_BASE: u64;
    static HEAP_SIZE: usize;
}
```
That way you can develop and test under Linux, and if that works, try it within an enclave.

arai-fortanix (approved):

PR #570, opened 2024-04-03
Replacing the allocation technique in memory allocation stress test

NirjharRoyiitk:
I am replacing the Vector new functions with raw allocator functions to allocate and deallocate memory


examples/mem-alloc-test/src/main.rs
```diff
@@ -54,15 +54,15 @@ const MAX_INDEX_CHECKS_PER_BUFFER: usize = 32;
 
 fn calculate_and_print_stat(
     shared_mutex_clone: Arc<Mutex<Counters>>,
-    tid: i32,
+    _tid: i32,
```
raoulstrackx, line 57: If `tid` isn't used, it can simply be removed.

examples/mem-alloc-test/src/main.rs
```diff
@@ -166,22 +167,29 @@ fn worker_thread(
         let start_time = Instant::now();
 
         /* Create an array of x GB where x is a random number between 1 to 4 */
-        let mut large_vector = Vec::with_capacity(size);
-        large_vector.resize(size, 0);
+
+        // Create a layout based on the size and alignment
+        let layout = Layout::from_size_align(size, ALIGN).unwrap();
```
arai-fortanix, line 172: Always aligning to the page size is stressful in some ways, but not very stressful in other ways, as we'll never end up with two allocations on the same page. It's probably better to pick the alignment randomly as well (with the restriction that the alignment must be a power of 2), and then check that the returned pointer has the correct alignment.

examples/mem-alloc-test/src/main.rs
```diff
@@ -103,22 +101,22 @@ fn wakeup_all_child_threads(pair_clone: Arc<(Mutex<bool>, Condvar)>) {
     drop(started);
 }
 
-fn traverse_buffer(buf: &mut Vec<i32>, scan_interval: usize) {
-    let mut ptr = 0;
+fn traverse_buffer(buf: *mut u8, size: usize, _scan_interval: usize) {
```
arai-fortanix, line 104: Is it intentional that scan_interval is no longer used? Should the parameter just be removed instead?

examples/mem-alloc-test/src/main.rs
```diff
@@ -166,22 +164,30 @@ fn worker_thread(
         let start_time = Instant::now();
 
         /* Create an array of x GB where x is a random number between 1 to 4 */
-        let mut large_vector = Vec::with_capacity(size);
-        large_vector.resize(size, 0);
+
+        // Create a layout based on the size and alignment
+        let align = get_random_num(2, PAGE_SIZE).next_power_of_two();
```
arai-fortanix, line 169: This works, but has a slight bias toward larger alignments. If you want the different alignments to have the same probability, you could do something like generating a number X between 0 and 12 inclusive, and then using the alignment (1 << X). Not a big deal.

examples/mem-alloc-test/src/main.rs
```diff
@@ -166,22 +164,30 @@ fn worker_thread(
...
+        let layout = Layout::from_size_align(size, align).unwrap();
+
+        // Allocate memory using the global allocator
+        let ptr = unsafe { alloc(layout) };
+        assert!(!ptr.is_null());
```
arai-fortanix, line 174: It would also be nice to check that the allocated memory is correctly aligned.

arai-fortanix (approved):

PR #577, opened 2024-04-16
bump version of ipc-queue for new release

Taowyoo:
- ipc-queue crates contains breaking changes, so need to bump version before publishing.


arai-fortanix (approved):
This seems fine, but when are we going to start making things 1.0 versions?

PR #585, opened 2024-04-23
chore: stop grouping dependabot version update

Taowyoo:
Stop grouping Dependabot version updates and Dependabot security updates.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -5,10 +5,6 @@ updates:
   schedule:
     interval: weekly
   open-pull-requests-limit: 10
-  groups:
-    crates-io:
-      patterns:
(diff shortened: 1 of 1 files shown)


arai-fortanix (approved):
Is the open-pull-requests-limit of 10 likely to be a problem now that we're not grouping updates?

Taowyoo:
@arai-fortanix 
I turn off grouping vision updates because some versions updates cause CI error. And dependabot won't create single version update when `groups` config exist.

With single version update, we could approve update PRs that pass the CI, and fix the failed update PR one by one or ask dependabot to ignore some dependency update if we do not want to update them.

I think 10 should be enough if anyone is working on these PRs.

PR #601, opened 2024-05-02
Glue code for using snmalloc in EDP

jethrogb:


.github/workflows/build.yml
```diff
@@ -48,7 +50,7 @@ jobs:
         rustup update
 
     - name: Cargo test --all --exclude sgxs-loaders
-      run: cargo test --verbose --locked --all --exclude sgxs-loaders --exclude async-usercalls && [ "$(echo $(nm -D target/debug/sgx-detect|grep __vdso_sgx_enter_enclave))" = "w __vdso_sgx_enter_enclave" ]
+      run: true || cargo test --verbose --locked --all --exclude sgxs-loaders --exclude async-usercalls --exclude snmalloc-edp && [ "$(echo $(nm -D target/debug/sgx-detect|grep __vdso_sgx_enter_enclave))" = "w __vdso_sgx_enter_enclave" ]
```
jethrogb, line 53: There appears to be an unrelated CI issue. Once fixed, please remove `true ||` here.
```
error: struct `Unverified` is never constructed
   --> intel-sgx/ias/src/api.rs:196:19
    |
196 | pub(crate) struct Unverified {}
    |                   ^^^^^^^^^^```

File: snmalloc-edp/src/rust-sgx-snmalloc-shim.cpp [hunk 2]
```diff
@@ -0,0 +1,187 @@
...
+#include "../snmalloc/src/snmalloc/snmalloc_core.h"
+
+using namespace snmalloc;
+
+using Globals = FixedRangeConfig<PALEdpSgx>;
+using Alloc = LocalAllocator<Globals>;
+
+/// Do global initialization for snmalloc. Should be called exactly once prior
+/// to any other snmalloc function calls.
+// TODO: this function shouldn't need the addresses passed in, these can be
```
jethrogb, line 126: TODO

snmalloc-edp/tests/global_alloc.rs
```diff
@@ -0,0 +1,111 @@
...
+        sn_rust_realloc(ptr, layout.align(), layout.size(), new_size)
+    }
+}
+
+// SAFETY: this should only be called once per thread, and the global
+// allocator shouldn't be used outside of this function
+unsafe fn with_thread_allocator<F: FnOnce() -> R, R>(f: F) -> R {
+    unsafe {
+        let layout = alloc::Layout::from_size_align(sn_alloc_size, sn_alloc_align).unwrap();
+        // TODO: bootstrap the thread-local allocator allocation a different way
```
jethrogb, line 47: TODO

snmalloc-edp/src/rust-sgx-snmalloc-shim.cpp
```diff
@@ -0,0 +1,187 @@
...
+#define SNMALLOC_MEMORY_PROVIDER PALEdpSgx
+#define OPEN_ENCLAVE
+// needed for openenclave header:
+#define OE_OK 0
+
+#include "../snmalloc/src/snmalloc/pal/pal_noalloc.h"
+
+namespace snmalloc {
+void register_clean_up() {
+    // TODO: not sure what this is supposed to do
```
jethrogb, line 71: TODO

snmalloc-edp/src/rust-sgx-snmalloc-shim.cpp
```diff
@@ -0,0 +1,187 @@
...
+    static constexpr uint64_t pal_features = EdpBasePAL::pal_features | Entropy;
+
+    template <bool page_aligned = false>
+    static void zero(void *p, size_t size) noexcept {
+        memset(p, 0, size);
+    }
+
+    static inline uint64_t get_entropy64() {
+        long long unsigned int result = 0;
+        while (_rdrand64_step(&result) != 1)
```
jethrogb, line 101: This should probably be a bounded loop that aborts after some number of tries

snmalloc-edp/src/rust-sgx-snmalloc-shim.cpp
```diff
@@ -0,0 +1,187 @@
...
+/*** snmalloc SGX PAL definition ***/
+/***********************************/
+
+#define SNMALLOC_PROVIDE_OWN_CONFIG
+#define SNMALLOC_SGX
+#define SNMALLOC_USE_SMALL_CHUNKS
+#define SNMALLOC_MEMORY_PROVIDER PALEdpSgx
+#define OPEN_ENCLAVE
+// needed for openenclave header:
+#define OE_OK 0
```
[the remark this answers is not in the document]
jethrogb: This is needed if OPEN_ENCLAVE is defined.

snmalloc-edp/src/rust-sgx-snmalloc-shim.cpp
```diff
@@ -0,0 +1,187 @@
...
+
+/***********************************/
+/*** snmalloc SGX PAL definition ***/
+/***********************************/
+
+#define SNMALLOC_PROVIDE_OWN_CONFIG
+#define SNMALLOC_SGX
+#define SNMALLOC_USE_SMALL_CHUNKS
+#define SNMALLOC_MEMORY_PROVIDER PALEdpSgx
+#define OPEN_ENCLAVE
```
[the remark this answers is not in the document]
jethrogb: Perhaps not, the impact in `snmalloc/ds_core/mitigations.h` should be investigated.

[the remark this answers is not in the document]
aditijannu: Looks like these mitigations are applied only to checked builds i.e. when `SNMALLOC_CHECK_CLIENT` is set. So it is disabled by default.

snmalloc-edp/src/rust-sgx-snmalloc-shim.cpp
```diff
@@ -0,0 +1,190 @@
...
+        memset(p, 0, size);
+    }
+
+    static inline uint64_t get_entropy64() {
+        long long unsigned int retry_count = 0;
+        long long unsigned int result = 0;
+        while (_rdrand64_step(&result) != 1 && retry_count < RAND_NUM_GEN_MAX_RETRIES) {
+            retry_count++;
+        }
+        return result;
```
jethrogb, line 111: This is not correct, this will return something even if the last call to rdrand returned an error.

NirjharRoyiitk: @jethrogb I was thinking, maybe I can have an assertion at the end of the loop, something like the following:

long long unsigned int retry_count = 0;
long long unsigned int result = 0;
int status;
while ( retry_count < RAND_NUM_GEN_MAX_RETRIES && status = _rdrand64_step(&result) != 1) {
    retry_count++;
}
assert(retry_count < RAND_NUM_GEN_MAX_RETRIES);
return result;

How does the above look?

jethrogb:
```c
for (size_t retry_count = 0; retry_count < RAND_NUM_GEN_MAX_RETRIES; retry_count++) {
long long unsigned int result;
if (_rdrand64_step(&result) == 1) {
    return result;
}
}
EdpErrorHandler::error("no entropy available");
```

snmalloc-edp/tests/global_alloc.rs
nshyrei: Is this todo still relevant? Because with @raoulstrackx we have figured that the way to use `sn_thread_init(allocator);`
and `sn_thread_cleanup(allocator);` is to put them around https://github.com/rust-lang/rust/blob/master/library/std/src/sys/pal/sgx/thread.rs#L43

snmalloc-edp/src/rust-sgx-snmalloc-shim.cpp
nshyrei: I haven't found what `register_clean_up` is supposed to do exactly. The search in snmalloc repository gives me this https://github.com/search?q=repo%3Amicrosoft%2Fsnmalloc%20register_clean_up&type=code, where the first match calls
 ```inline void register_clean_up()
  {
    error("Critical Error: This should never be called.");
  }```
cc @raoulstrackx

snmalloc-edp/tests/global_alloc.rs
jethrogb: > put them around https://github.com/rust-lang/rust/blob/master/library/std/src/sys/pal/sgx/thread.rs#L43.

That is not right. That code is only run for additional threads besides the main thread. It should be somewhere in entry.

> Is this todo still relevant?

Yes you still need to allocate memory for the thread allocator.

.github/workflows/build.yml
```diff
@@ -96,6 +98,24 @@ jobs:
     - name: Build em-app, get-certificate for x86_64-fortanix-unknown-sgx
       run: cargo build --verbose --locked -p em-app -p get-certificate --target=x86_64-fortanix-unknown-sgx
 
+    - name: Build snmalloc-edp
+      run: |
+        git submodule update --init --recursive
+        detect_cxx_include_path() {
+            for path in $(clang++-12 -print-search-dirs|sed -n 's/^libraries:\s*=//p'|tr : ' '); do
```
jethrogb, line 105: Should use `CXX_x86_64_fortanix_unknown_sgx`

PR #609, opened 2024-05-24
[Check code with CI] aj/update-sgx-alloc

Taowyoo:
Glue code for snmalloc allocator in SGX enclaves.


Taowyoo:
The code change here should be moved to https://github.com/fortanix/rust/pull/16 
CC: @NirjharRoyiitk @aditijannu 
Could you help doing this ?

snmalloc-edp/src/rust-sgx-snmalloc-shim.cpp
```diff
@@ -0,0 +1,181 @@
...
+#define SNMALLOC_PROVIDE_OWN_CONFIG
+#define SNMALLOC_SGX
+#define SNMALLOC_USE_SMALL_CHUNKS
+#define SNMALLOC_MEMORY_PROVIDER PALEdpSgx
+
+#include "../snmalloc/src/snmalloc/pal/pal_noalloc.h"
+
+namespace snmalloc {
+void register_clean_up() {
+    // TODO: not sure what this is supposed to do
```
arai-fortanix, line 68: We probably need to resolve this before upstreamining.

snmalloc-edp/src/rust-sgx-snmalloc-shim.cpp
```diff
@@ -0,0 +1,181 @@
...
+
+/***************************************************/
+/*** Imported symbols needed by snmalloc SGX PAL ***/
+/***************************************************/
+
+// from entry.S
+extern "C" size_t get_tcs_addr();
+
+// from Rust std
+extern "C" void __rust_print_err(const char *m, size_t s);
```
arai-fortanix, line 37: The existing snmalloc code looks like i't pretty consistent about writing pointer arguments like `const char* m` and not `const char *m`, even in extern"C" blocks, so we should probably follow that.

snmalloc-edp/tests/global_alloc.rs
```diff
@@ -0,0 +1,108 @@
...
+
+        THREAD_ALLOC.set(ptr::null_mut());
+        sn_thread_cleanup(allocator.as_mut_ptr());
+
+        r
+    }
+}
+
+#[test]
+fn test() {
```
arai-fortanix, line 60: I know that this isn't the only testing we have for the allocator/snmalloc, but I think we should have at least a little more testing here. At a minimum, I think we should exercise all of the alloc/alloc_zeroed/dealloc/realloc functions.

snmalloc-edp/src/rust-sgx-snmalloc-shim.cpp
arai-fortanix: I'm not 100% sure on this, but the only place I see `register_clean_up()` being called is if the allocator supports lazy initialization. I'm assuming our allocator doesn't support lazy initialization, so I think we probably don't need our `register_clean_up()` to do anything.

.github/workflows/build.yml
```diff
@@ -16,7 +16,9 @@ env:
   CARGO_INCREMENTAL: 0
   CARGO_NET_RETRY: 10
   CFLAGS_x86_64_fortanix_unknown_sgx: "-isystem/usr/include/x86_64-linux-gnu -mlvi-hardening -mllvm -x86-experimental-lvi-inline-asm-hardening"
+  # CXXFLAGS is set below
```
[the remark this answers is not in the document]
jethrogb: Because it needs to call `detect_cxx_include_path`

snmalloc-edp/tests/global_alloc.rs
aditijannu: This test also seems to be failing with a signal 11 (SIGSEGV), I need to investigate what the issue is. But when snmalloc-edp is plugged into the rust compiler's standard library, I see all the tests assembly and run-make tests (tests included in CI) succeed.

Taowyoo:
Close this because  snmalloc testing is done

PR #621, opened 2024-07-02
Removed unused error variants

nshyrei:
Removes error variants that are no longer in use

### DIFF
--- a/intel-sgx/dcap_artifact_retrieval/src/lib.rs
+++ b/intel-sgx/dcap_artifact_retrieval/src/lib.rs
@@ -11,16 +11,11 @@ use std::str::Utf8Error;
 #[cfg(feature = "reqwest")]
 use reqwest::blocking::{Client as ReqwestClient};
 use pcs::Error as OAError;
-use pkix::ASN1Error;
 use quick_error::quick_error;
 
 quick_error! {
     #[derive(Debug)]
     pub enum Error {
-        PckIdParseError(msg: &'static str) {
-            description("Error during parsing PCKID file")
-            display("Error parsing PCKID file: {}", msg)
-        }
         ReadResponseError(msg: Cow<'static, str>) {
             from()
             display("{}", msg)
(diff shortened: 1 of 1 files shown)


nshyrei:
@Raoul Strackx The OfflineAttestationError, BadRequest and PCSError are still used in PckCertApi (client/dcap_artifact_retrieval/src/provisioning_client/azure.rs:87 ). Is this functionality still needed?

nshyrei:
@raoulstrackx Please review

raoulstrackx:
The `OfflineAttestationError` error should also be removed. It doesn't convey any meaningful info. Worse, it doesn't even make sense if you get this error while doing online attestation. You'll need to come up with a better approach/name.
`BadRequest` errors should also be removed as it can also be expressed as a `PCSError` with a status code, so that also doesn't make sense.

nshyrei:
@raoulstrackx updated, please review.

intel-sgx/dcap_artifact_retrieval/src/lib.rs
```diff
@@ -52,48 +42,21 @@ quick_error! {
...
         NoQeID {
             description("QEID is required, but not provided")
             display("No QE ID was provided")
         }
-        NoAPIKey {
-            description("PCS key is required, but not provided")
-            display("No api_key was provided")
-        }
-        OfflineAttestationError(err: OAError) {
+        PCSProvisioningError(err: pcs::Error) {
```
nshyrei, line 57: @raoulstrackx This is the most generic name I came up with that I believe describes the logic of the 2 places this error variant is used in:
1) ```let tcbinfo = prov_client.tcbinfo(&fmspc)?;
        let tcbinfo_file = tcbinfo.store(output_dir).map_err(|e| Error::PCSProvisioningError(e))?;
```
2) fn parse_response(&self, response_body: String, response_headers: Vec<(String, String)>, _api_version: PcsVersion) -> Result<Self::Output, Error> {
        let ca_chain = parse_issuer_header(&response_headers, "SGX-PCK-Certificate-Issuer-Chain")?;
        PckCerts::parse(&response_body, ca_chain).map_err(|e| Error::PCSProvisioningError(e))
    }
```

raoulstrackx: The first one is only used in the `main.rs`. So you can easily resolve that in another way.
The second is about errors in parsing the result you get from the PCS when fetching `PckCerts`. So `PckCertsParseError` or `BodyParseError` would make sense

nshyrei:
@raoulstrackx updated, please review

raoulstrackx (approved):

PR #622, opened 2024-07-04
Add feature flag for reducing randomness in memalloc test

aditijannu:


examples/mem-alloc-test/src/main.rs
```diff
@@ -218,7 +239,21 @@ fn worker_thread(
         /* Create an array of x GB where x is a random number between 1 to 4 */
 
         // Create a layout based on the size and alignment
-        let align = get_random_num(2, PAGE_SIZE).next_power_of_two();
+        let align: usize;
+        #[cfg(feature = "reduce_randomness")]
+        {
+            align = PAGE_SIZE;
+            size = match memsize {
```
NirjharRoyiitk, line 246: Minor: Can we have the size conditionally defined in the ```match``` case before this? Since both the size and the layout is being conditionally defined, I was thinking if we can have the size in the condition block this 


```
/* Create a random size depending on the memory type */
        let (scan_interval, mut size, limit) = match memsize {
            MemSize::Large => {
                (
                    SCAN_INTERVAL_LARGE_THREAD,
                    #[cfg(feature = "reduce_randomness")] {crate::TO_GB * 1} 
                    #[cfg(not(feature = "reduce_randomness"))] {TO_GB * get_random_num(LARGE_THREAD_MEM_START, LARGE_THREAD_MEM_END)},
                    LIMIT_LARGE_THREAD,
                )
            }
            MemSize::Medium => {
                (
                    SCAN_INTERVAL_MEDIUM_THREAD,
                    // similarly for medium and small threads
                    TO_MB * get_random_num(MEDIUM_THREAD_MEM_START, MEDIUM_THREAD_MEM_END),
                    LIMIT_MEDIUM_THREAD,
                )
            }
            MemSize::Small => {
                (
                    SCAN_INTERVAL_SMALL_THREAD,
                    TO_KB * get_random_num(SMALL_THREAD_MEM_START, SMALL_THREAD_MEM_END),
                    LIMIT_SMALL_THREAD,
                )
            }
        };

```

NirjharRoyiitk (approved):

PR #624, opened 2024-07-05
Add allow dead_code attribute to structs 

aditijannu:
Source files of `ipc-queue` have `#![deny(warnings)]` attribute and some structs are not always used which causes the compiler to throw warnings like this:
```
error: struct `AsyncSender` is never constructed
   --> ipc-queue/src/lib.rs:149:12
    |
149 | pub struct AsyncSender<T: 'static, S> {
    |            ^^^^^^^^^^^
```
To avoid these warnings to turn into errors, add the allow[dead_code] attribute to these structs.


ipc-queue/src/lib.rs
```diff
@@ -146,11 +146,13 @@ pub trait AsyncSynchronizer: Clone {
     fn notify(&self, event: QueueEvent);
 }
 
+#[allow(dead_code)]
 pub struct AsyncSender<T: 'static, S> {
     inner: Fifo<T>,
     synchronizer: S,
 }
 
+#[allow(dead_code)]
```
raoulstrackx, line 155: Simply allowing dead code here, doesn't tell the user why this would be dead code. In this case it's only dead code when the target is not sgx. (The tests pass for the SGX target). I would annotate it as such. The same applies for the other errors as well I think.

PR #625, opened 2024-07-12
Allow specifying number of worker threads used to run an enclave

s-arash:
This change allows users of this library to specify the number of worker threads. This lets users limit the number of cores used by an enclave. 

This is a backwards-compatible change.

### DIFF
--- a/intel-sgx/enclave-runner/src/command.rs
+++ b/intel-sgx/enclave-runner/src/command.rs
@@ -24,6 +24,7 @@ pub struct Command {
     forward_panics: bool,
     force_time_usercalls: bool,
     cmd_args: Vec<Vec<u8>>,
+    num_worker_threads: usize,
 }
 
 impl MappingInfo for Command {
@@ -47,6 +48,7 @@ impl Command {
         forward_panics: bool,
         force_time_usercalls: bool,
         cmd_args: Vec<Vec<u8>>,
+        num_worker_threads: usize,
     ) -> Command {
         let main = tcss.remove(0);
         Command {
@@ -58,6 +60,7 @@ impl Command {
             forward_panics,
             force_time_usercalls,
             cmd_args,
+            num_worker_threads,
         }
     }
 
@@ -66,6 +69,6 @@ impl Command {
     }
 
     pub fn run(self) -> Result<(), Error> {
-        EnclaveState::main_entry(self.main, self.threads, self.usercall_ext, self.forward_panics, self.force_time_usercalls, self.cmd_args)
+        EnclaveState::main_entry(self.main, self.threads, self.usercall_ext, self.forward_panics, self.force_time_usercalls, self.cmd_args, self.num_worker_threads)
     }
 }
--- a/intel-sgx/enclave-runner/src/loader.rs
+++ b/intel-sgx/enclave-runner/src/loader.rs
@@ -6,6 +6,7 @@
 
 use std::fs::File;
 use std::io::{Error as IoError, ErrorKind, Read, Result as IoResult};
+use std::ops::RangeInclusive;
 use std::os::raw::c_void;
 use std::path::Path;
 use std::{arch, str};
@@ -70,6 +71,7 @@ pub struct EnclaveBuilder<'a> {
     forward_panics: bool,
     force_time_usercalls: bool,
     cmd_args: Option<Vec<Vec<u8>>>,
+    num_worker_threads: Option<usize>,
 }
 
 #[derive(Debug, ThisError)]
@@ -148,6 +150,7 @@ impl<'a> EnclaveBuilder<'a> {
             forward_panics: false,
             force_time_usercalls: true, // By default, keep the old behavior of always doing a usercall on an insecure_time call
             cmd_args: None,
+            num_worker_threads: None,
         };
 
         let _ = ret.coresident_signature();
@@ -310,11 +313,19 @@ impl<'a> EnclaveBuilder<'a> {
     /// Adding command arguments and then calling [`build_library`] will cause
     /// a panic.
     ///
-    /// [`Command`]: struct.Command.html
     /// [`build_library`]: struct.EnclaveBuilder.html#method.build_library
     pub fn arg<S: AsRef<[u8]>>(&mut self, arg: S) -> &mut Self {
         let arg = arg.as_ref().to_owned();
         self.initialized_args_mut().push(arg);
+        self
+    }
+
+    /// Sets the number of worker threads used to run the enclave.
+    ///
+    /// **NOTE:** This is only applicable to [`Command`] enclaves.
+    /// Setting this and then calling [`build_library`](Self::build_library) will cause a panic.
+    pub fn num_worker_threads(&mut self, num_worker_threads: usize) -> &mut Self {
+        self.num_worker_threads = Some(num_worker_threads);
         self
     }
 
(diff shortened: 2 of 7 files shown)


intel-sgx/enclave-runner/src/command.rs
```diff
@@ -63,6 +63,13 @@ impl Command {
     }
 
     pub fn run(self) -> Result<(), Error> {
-        EnclaveState::main_entry(self.main, self.threads, self.usercall_ext, self.forward_panics, self.cmd_args)
+        EnclaveState::main_entry(self.main, self.threads, self.usercall_ext, self.forward_panics, self.cmd_args, num_cpus::get())
+    }
+
+    /// Run the enclave with the provided number of worker threads.
+    /// # Panics
```
arai-fortanix, line 70: Why is it a panic if the number of worker threads is 0 and not an error?

arai-fortanix (dismissed):

aditijannu:
I had been working the CI fixes here which you will face in your CI job too - https://github.com/fortanix/rust-sgx/pull/624
It is not complete yet and I need to fix one of the nitro crates used in rust-sgx.

intel-sgx/enclave-runner/src/command.rs
s-arash: I can make it an error. But my thinking is that `worker_threads == 0` is a user error/precondition failure that the user must fix and not an error that comes out of running the enclave.

Another option would be to use `NonZeroUsize`, but it's very rarely used in Rust libraries.

arai-fortanix: Libraries should try really hard to avoid panics. Panics in libraries should really only be for fatal conditions that the library has no way to recover from, like internal data structure inconsistencies that can't be reconciled somehow. An incorrect number of worker threads is absolutely not fatal in this way. Either the API should make it impossible to pass 0 via `NonZeroUsize`, or the call should return an error. The caller could decide it wants to panic if it gets this error, or it might try to make the call again with a different number of threads.

I think having an error for an invalid number of threads makes sense, as we probably want to have some upper bound on the number of threads as well. Does it really make sense for the application to request for example trillions of threads?

intel-sgx/enclave-runner/src/command.rs
```diff
@@ -63,6 +63,13 @@ impl Command {
...
+    /// Panics if `worker_threads == 0`
+    pub fn run_with_worker_threads(self, worker_threads: usize) -> Result<(), Error> {
```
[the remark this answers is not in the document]
s-arash: I switched to using the builder pattern, as suggested by @Goirad.

intel-sgx/enclave-runner/src/command.rs
s-arash: I'm now returning an error instead of panicking when `num_worker_threads` is set to 0.

s-arash:
> I had been working the CI fixes here which you will face in your CI job too - #624 It is not complete yet and I need to fix one of the nitro crates used in rust-sgx.

Thanks @aditijannu. I'll rebase this PR once your PR lands.

intel-sgx/enclave-runner/src/loader.rs
```diff
@@ -298,14 +300,22 @@ impl<'a> EnclaveBuilder<'a> {
...
     ///
-    /// [`Command`]: struct.Command.html
     /// [`build_library`]: struct.EnclaveBuilder.html#method.build_library
     pub fn arg<S: AsRef<[u8]>>(&mut self, arg: S) -> &mut Self {
         let arg = arg.as_ref().to_owned();
         self.initialized_args_mut().push(arg);
         self
     }
 
+    /// Sets the number of wroker threads used to run the enclave.
```
arai-fortanix, line 310: Typo: wroker

intel-sgx/enclave-runner/src/loader.rs
```diff
@@ -332,21 +342,26 @@ impl<'a> EnclaveBuilder<'a> {
     }
 
     pub fn build<T: Load>(mut self, loader: &mut T) -> Result<Command, anyhow::Error> {
+        let num_cpus = num_cpus::get();
+        let num_worker_threads = self.num_worker_threads.unwrap_or(num_cpus);
+        anyhow::ensure!(num_worker_threads > 0, "`num_worker_threads` cannot be zero");
+        anyhow::ensure!(num_worker_threads <= num_cpus, "`num_worker_threads` cannot exceed number of logical cores (`num_cpus::get()`)");
```
arai-fortanix, line 348: I'm not sure this is a useful check, in the presence of CPU hotplug. The system could have more CPUs added to it later, ot it could have CPUs removed from it at run time. So we could disallow a number of worker threads that could later become legal, or we might allow a number of threads now that becomes illegal later.

intel-sgx/enclave-runner/src/loader.rs
```diff
@@ -332,21 +342,26 @@ impl<'a> EnclaveBuilder<'a> {
...
         self.initialized_args_mut();
         let args = self.cmd_args.take().unwrap_or_default();
         let c = self.usercall_ext.take();
         self.load(loader)
-            .map(|(t, a, s, fp)| Command::internal_new(t, a, s, c, fp, args))
+            .map(|(t, a, s, fp)| Command::internal_new(t, a, s, c, fp, args, num_worker_threads))
     }
 
-    /// Panics if you have previously called [`arg`] or [`args`].
+    /// Panics if you have previously called [`arg`], [`args`], or [`num_worker_threads`].
```
arai-fortanix, line 357: With so many parameters that aren't shared between Command and Library enclaves, does it maybe make sense to have separate builders for Command and Library enclaves? That way, each type of builder could only have the parameters that make sense for the type of enclave it builds.

arai-fortanix (dismissed):

intel-sgx/enclave-runner/src/loader.rs
s-arash: Hmm, I can remove the upper bound check in that case, or have a fixed upper bound (maybe something like 512?). Which one would you prefer?

intel-sgx/enclave-runner/src/loader.rs
s-arash: I agree. But that would be a breaking change. I think it's better to release a sem-ver compatible version with this change, and leave this kind of refactoring to a future sem-ver breaking version.

intel-sgx/enclave-runner/src/loader.rs
s-arash: Fixed, thanks!

intel-sgx/enclave-runner/src/loader.rs
```diff
@@ -332,21 +342,25 @@ impl<'a> EnclaveBuilder<'a> {
     }
 
     pub fn build<T: Load>(mut self, loader: &mut T) -> Result<Command, anyhow::Error> {
+        let num_worker_threads = self.num_worker_threads.unwrap_or_else(num_cpus::get);
+        anyhow::ensure!(num_worker_threads > 0, "`num_worker_threads` cannot be zero");
+        anyhow::ensure!(num_worker_threads <= 512, "`num_worker_threads` cannot exceed 512");
```
Goirad, line 347: Is this needed at all? It might be unlikely but I don't think we should _prevent_ people from spawning more threads than there are cores. It's just inefficient, but not wrong.

s-arash: Right. That's a good point. But maybe we should enforce a reasonable maximum? It doesn't have to be 512, but maybe we should somehow prevent specifying something like `1 << 60` as `num_worker_threads`.

intel-sgx/enclave-runner/src/loader.rs
arai-fortanix: I think a fixed upper bound is fine, but 512 is too low. There are already systems with something like 320 logical cores, so we're not that far away from systems with 512. Something like 64K could be reasonable.

s-arash: Will make it 64K!

s-arash:
@raoulstrackx @aditijannu @Taowyoo @mzohreva @Goirad Could you take a look at this PR? We need to merge this soon.

intel-sgx/enclave-runner/src/loader.rs
```diff
@@ -346,21 +357,30 @@ impl<'a> EnclaveBuilder<'a> {
...
 
     pub fn build<T: Load>(mut self, loader: &mut T) -> Result<Command, anyhow::Error> {
+        if let Some(num_worker_threads) = self.num_worker_threads {
+            const NUM_WORKER_THREADS_RANGE: RangeInclusive<usize> = 1..=65536;
+            anyhow::ensure!(
+                NUM_WORKER_THREADS_RANGE.contains(&num_worker_threads),
+                "`num_worker_threads` must be in range {NUM_WORKER_THREADS_RANGE:?}"
+            );
+        }
+        let num_worker_threads = self.num_worker_threads.unwrap_or_else(num_cpus::get);
```
Taowyoo, line 367: Will there be any problems when the specified worker thread number is bigger than the number of cpus

Taowyoo (commented):
Lgtm,, left one question regarding available range of input thread number

intel-sgx/enclave-runner/src/loader.rs
s-arash: It won't be an issue. It just may not be optimal.

Taowyoo (approved):

PR #631, opened 2024-08-13
- pin version

nshyrei:


Taowyoo:
this pin is no longer needed

PR #633, opened 2024-08-14
Fix CI

raoulstrackx:


Cargo.toml
```diff
@@ -48,6 +48,8 @@ exclude = [
 resolver = "2"
 
 [patch.crates-io]
+aws-nitro-enclaves-cose_0_1 = { git = "https://github.com/fortanix/aws-nitro-enclaves-cose.git", branch = "raoul/0.1.1", package = "aws-nitro-enclaves-cose" }
```
aditijannu, line 51: Can we have a comment here on why there are two versions of the same dependency?

aditijannu (dismissed):

Cargo.toml
raoulstrackx: One patches `aws-nitro-enclaves-cose` version 0.1, the other 0.5. Both apply a similar fix. I was hoping this was clear from the name. What kind of comment would you like?

fortanix-vme/aws-nitro-enclaves/nitro-attestation-verify/Cargo.toml
```diff
@@ -10,7 +10,7 @@ chrono = "0.4"
 serde_cbor = "0.11"
 # Required until PR36 is accepted
```
aditijannu, line 11: These two comments can be cleaned up now since we are not using the branch here anymore.

nshyrei (dismissed):

fortanix-vme/aws-nitro-enclaves/nitro-attestation-verify/Cargo.toml
raoulstrackx: I missed that. Good catch! This patch for v0.5 isn't even needed.

aditijannu (approved):

PR #637, opened 2024-08-27
yasna & dependencies update

mridul-manohar:
The currently used version of yasna is ver0.3, and it is advised to upgrade to latest yasna ver0.5.2.
This PR upgrades yasna version along with required dependencies.


Taowyoo:
Regarding CI failure, I guess there is some `libc` version change breaks `vsock` compilation.

PR #648, opened 2024-10-31
Insecure time through `rdtscp`

raoulstrackx:
SGXv2 platforms support calling the `rdtscp` instruction inside an enclave. This PR implements logic to keep track of the time inside of an enclave. Note that the value `rdtscp` returns, can be manipulated by an attacker; even with these changes `insecure_time` is (almost) as insecure as before. There's only an option to force the time to be monotonic.

This PR is blocked on internal API review


intel-sgx/dcap-artifact-retrieval/src/main.rs
```diff
@@ -71,25 +71,6 @@ pub fn download_dcap_artifacts(
...
-    let pckcrl = prov_client
-        .pckcrl()
-        .and_then(|crl| crl.write_to_file(output_dir).map_err(|e| e.into()))?;
-    let qe_identity = prov_client
-        .qe_identity()
-        .and_then(|qe_id| qe_id.write_to_file(output_dir).map_err(|e| e.into()))?;
-    if verbose {
-        println!("==[ generic ]==");
-        println!("   pckcrl:      {}", pckcrl);
-        println!("   QE identity: {}", qe_identity);
```
Taowyoo, line 92: Could I assume they  are deleted because only pck certs are necessary for dcap attestation?

insecure-time/src/lib.rs
```diff
@@ -0,0 +1,853 @@
...
+
+    fn set(&self, t: Ticks) {
+        let t = t.0.load(Ordering::Relaxed);
+        self.0.store(t, Ordering::Relaxed);
+    }
+
+    pub fn from_duration(duration: Duration, freq: &Freq) -> Self {
+        let freq = freq.as_u64();
+        let ticks_secs = duration.as_secs() * freq;
+        let ticks_nsecs = duration.subsec_nanos() as u64 * freq / NANOS_PER_SEC as u64;
```
zugzwang, line 136: consider using [subsec_millis](https://doc.rust-lang.org/std/time/struct.Duration.html#method.subsec_millis)

insecure-time/src/lib.rs
```diff
@@ -0,0 +1,853 @@
...
+    }
+}
+
+#[derive(PartialEq, Eq, Debug)]
+pub enum ClockSource {
+    TSC,
+}
+
+impl Freq {
+    pub fn new(crystal_hz: u32, numerator: u32, denominator: u32) -> Self {
```
zugzwang, line 223: since we're dividing and computing remainders by `Freq` in a lot of places, maybe we should ensure this is never 0

zugzwang: Nevermind, I just saw the `assert!(freq != 0);` in line 228 below.

intel-sgx/enclave-runner/src/usercalls/mod.rs
```diff
@@ -1633,11 +1650,13 @@ impl<'tcs> IOHandlerInput<'tcs> {
     }
 
     #[inline(always)]
-    fn insecure_time(&mut self) -> u64 {
+    fn insecure_time(&mut self) -> (u64, *const InsecureTimeInfo) {
         let time = time::SystemTime::now()
             .duration_since(time::UNIX_EPOCH)
             .unwrap();
-        (time.subsec_nanos() as u64) + time.as_secs() * 1_000_000_000
+        let t = (time.subsec_nanos() as u64) + time.as_secs() * 1_000_000_000;
```
zugzwang, line 1657: Use a constant as before `NANOS_PER_SECOND`

intel-sgx/fortanix-sgx-abi/src/lib.rs
```diff
@@ -566,13 +567,21 @@ impl Usercalls {
...
+}
+
 /// # Miscellaneous
 impl Usercalls {
     /// This returns the number of nanoseconds since midnight UTC on January 1,
     /// 1970\. The enclave must not rely on the accuracy of this time for
     /// security purposes, such as checking credential expiry or preventing
     /// rollback.
-    pub fn insecure_time() -> u64 { unimplemented!() }
+    pub fn insecure_time() -> (u64, *const InsecureTimeInfo) { unimplemented!() }
```
zugzwang, line 584: PR guard, in case this needs to be addressed in a subsequent commit

insecure-time/src/lib.rs
```diff
@@ -0,0 +1,853 @@
...
+#[cfg(feature = "std")]
+impl NativeTime for SystemTime {
+    fn minimum() -> Self {
+        SystemTime::UNIX_EPOCH
+    }
+
+    fn abs_diff(&self, other: &Self) -> Duration {
+        match self.duration_since(*other) {
+            Ok(duration) => duration,
+            Err(e) => e.duration()
```
zugzwang, line 45: Just learned about `duration_since`. Since it is odd, I think it merits a comment explaining that `Err(e)` contains the duration when `other` is earlier than self.

insecure-time/src/lib.rs
```diff
@@ -0,0 +1,853 @@
...
+        let mut ecx = 0;
+        let mut edx = 0;
+
+        cpuid(0x12, &mut eax, &mut ebx, &mut ecx, &mut edx);
+        eax & (0x1 << 1) != 0x0
+
+    }
+}
+
+pub struct Freq(AtomicU64);
```
zugzwang, line 179: From the code I understand this is an integer number representing the amount of ticks per second (add a comment).

Also is this a reasonable assumption? E.g. each second has an (exact) integer amount of ticks? Put otherwise, are ticks aligned to seconds?

raoulstrackx: Keeping track of the time is tricky. Some platforms do report their frequency, but do that incorrectly, others report it not at all (for older platforms, probably), or incompletely. The expectation is that processors that do support SGXv2 (and can thus call `rdtsc` in the enclave) do report their frequency, but running the main application it was observed that this reported frequency isn't correct (and deviates more than a fraction of a tick). The logic in this crates was made to be able to handle this.
I've added comments on the `Tsc` struct.

insecure-time/src/lib.rs
raoulstrackx: Added a comment

intel-sgx/fortanix-sgx-abi/src/lib.rs
raoulstrackx: PR change was approved

intel-sgx/dcap-artifact-retrieval/src/main.rs
raoulstrackx: Yes that shouldn't have been added to the commit.

PR #657, opened 2024-11-15
fix outdated info in README.md

Taowyoo:
- Replace the old Travis CI badge with GitHub Action's one.
- Correct the path to SGX code in https://github.com/rust-lang/rust.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # Fortanix Rust Enclave Development Platform
 
-[![Runtime Encryption Slack](https://img.shields.io/badge/Slack-%23rust--sgx-blue.svg?logo=slack&logoColor=whitesmoke)](https://fortanix.com/runtime-encryption-slack) [![Build Status](https://travis-ci.com/fortanix/rust-sgx.svg?branch=master)](https://app.travis-ci.com/github/fortanix/rust-sgx/builds)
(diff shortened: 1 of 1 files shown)


nshyrei (approved):

Taowyoo:
Hi @nshyrei  could you help re-approve this PR.
I update CI to not ignore changes on docs.
This is a simple fix for now because To make CI and status check to work correctly with path ignore is complex.

.github/workflows/build.yml
```diff
@@ -2,9 +2,6 @@ name: CI
 
 on:
   pull_request:
-    paths-ignore:
-      - '*.md'
-      - 'LICENSE'
```
raoulstrackx, line 7: Why is this removed?

Taowyoo: Without this, no CI will run then it breaks the check requirement on the PR.

README.md
```diff
@@ -24,7 +24,7 @@ cargo run --target x86_64-fortanix-unknown-sgx
 [Documentation](https://edp.fortanix.com/docs/)
 
 Looking for the in-enclave source code? Check out the [rust-lang/rust
-repository](https://github.com/rust-lang/rust/tree/master/library/std/src/sys/sgx).
+repository](https://github.com/rust-lang/rust/tree/master/library/std/src/sys/pal/sgx).
```
raoulstrackx: The upstream Rust compiler changed its code structure a bit. There are other SGX specific changes that are not under that directory. I'd change just point to `https://github.com/rust-lang/rust/tree/master/library/std`

Taowyoo: Since I already started discussion in the PR https://github.com/fortanix/rust-sgx/pull/853
I am going to change this PR to only fix the CI.
Then suggest contributer to do these change instead.

PR #658, opened 2024-11-19
Update to mbedtls v0.13.0

mridul-manohar:
rust-mbedtls [PR #372](https://github.com/fortanix/rust-mbedtls/pull/372) adds combined errors for mbedtls. this enhancement is available from mbedtls v0.13.0

### DIFF
--- a/fortanix-vme/aws-nitro-enclaves/eif-tools/Cargo.toml
+++ b/fortanix-vme/aws-nitro-enclaves/eif-tools/Cargo.toml
@@ -22,7 +22,7 @@ env_logger = "0.9"
 log = "0.4"
 nitro-cli = { git = "https://github.com/fortanix/aws-nitro-enclaves-cli.git", branch = "main" }
 once_cell = "1.9.0"
-serde = { version = "1.0", features = ["derive"] }
+serde = { version = "1.0.214", features = ["derive"] }
 sha2 = "0.9.5"
 tempdir = "0.3"
 thiserror = "1.0"
--- a/fortanix-vme/aws-nitro-enclaves/nitro-attestation-verify/Cargo.toml
+++ b/fortanix-vme/aws-nitro-enclaves/nitro-attestation-verify/Cargo.toml
@@ -9,9 +9,9 @@ publish = false
 chrono = "0.4"
 serde_cbor = "0.11"
 aws-nitro-enclaves-cose = { version = "0.5.0", default-features = false }
-mbedtls = { version = "0.12", features = ["rdrand", "std", "time", "ssl"], default-features = false, optional = true }
+mbedtls = { version = "0.13.0", features = ["rdrand", "std", "time", "ssl"], default-features = false, optional = true }
 num-bigint = "0.4"
-serde = { version = "1.0", features = ["derive"] }
+serde = { version = "1.0.214", features = ["derive"] }
 serde_bytes = "0.11"
 pkix = ">=0.1.2, <0.3.0"
 yasna = { version = "0.4", features = ["num-bigint"] }
--- a/fortanix-vme/aws-nitro-enclaves/nitro-attestation-verify/src/mbedtls.rs
+++ b/fortanix-vme/aws-nitro-enclaves/nitro-attestation-verify/src/mbedtls.rs
@@ -6,6 +6,7 @@
 use aws_nitro_enclaves_cose::crypto::{Hash, MessageDigest, SignatureAlgorithm, SigningPublicKey};
 use aws_nitro_enclaves_cose::error::CoseError;
 use mbedtls::alloc::Box as MbedtlsBox;
+use mbedtls::error::{codes, Error as ErrMbed};
 use mbedtls::hash::{self, Md};
 use mbedtls::pk::EcGroupId;
 use mbedtls::x509::Certificate;
@@ -97,7 +98,7 @@ impl SigningPublicKey for WrappedCert {
         // We'll throw error if signature verify does not work
         match pk.verify(*md, &digest, &sig) {
             Ok(_) => Ok(true),
-            Err(mbedtls::Error::EcpVerifyFailed) => Ok(false),
+            Err(ErrMbed::HighLevel(codes::EcpVerifyFailed)) => Ok(false),
             Err(e) => Err(CoseError::SignatureError(Box::new(e))),
         }
     }
--- a/intel-sgx/pcs/Cargo.toml
+++ b/intel-sgx/pcs/Cargo.toml
@@ -13,7 +13,7 @@ sgx-isa = { path = "../sgx-isa", default-features = true }
 pkix = "0.2.0"
 yasna = { version = "0.3", features = ["num-bigint", "bit-vec"] }
 rustc-serialize = "0.3"
-serde = { version = "1.0.7", features = ["derive"] }
+serde = { version = "1.0.214", features = ["derive"] }
 sgx_pkix = "0.2"
 serde_json = { version = "1.0", features = ["raw_value"] }
 percent-encoding = "2.1.0"
@@ -23,7 +23,7 @@ failure = "0.1.1"
 anyhow = { version = "1", optional = true }
 quick-error = "1.2.3"
 num = "0.2"
-mbedtls = { version = "0.12.3", features = ["std", "time"], default-features = false, optional = true }
+mbedtls = { version = "0.13.0", features = ["std", "time"], default-features = false, optional = true }
 
 [dev-dependencies]
 hex = "0.4.2"
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -122,7 +122,7 @@ version = "0.1.0"
 source = "git+https://github.com/fortanix/aws-nitro-enclaves-cose.git?branch=v0.1.0-patched#fe2558390c13cbd0e82f9c1610e74c4a4ddfd7e1"
 dependencies = [
  "openssl",
- "serde",
+ "serde 1.0.215",
  "serde_bytes",
  "serde_cbor",
  "serde_repr",
@@ -135,7 +135,7 @@ version = "0.5.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "5ce1d9954a5cb2841ad8ab206a050cd07ed34200ea6aafb7fa73a33771aaf48c"
 dependencies = [
- "serde",
+ "serde 1.0.215",
  "serde_bytes",
  "serde_cbor",
  "serde_repr",
@@ -151,7 +151,7 @@ dependencies = [
  "libc",
  "log 0.4.21",
  "nix 0.20.2",
- "serde",
+ "serde 1.0.215",
  "serde_bytes",
  "serde_cbor",
 ]
@@ -246,7 +246,7 @@ dependencies = [
  "lazycell",
  "log 0.4.21",
  "peeking_take_while",
- "proc-macro2 1.0.78",
+ "proc-macro2 1.0.89",
  "quote 1.0.35",
  "regex",
  "rustc-hash",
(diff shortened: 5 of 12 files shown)


fortanix-vme/aws-nitro-enclaves/eif-tools/Cargo.toml
```diff
@@ -22,7 +22,7 @@ env_logger = "0.9"
 log = "0.4"
 nitro-cli = { git = "https://github.com/fortanix/aws-nitro-enclaves-cli.git", branch = "main" }
 once_cell = "1.9.0"
-serde = { version = "1.0", features = ["derive"] }
+serde = { version = "1.0.214", features = ["derive"] }
```
Taowyoo, line 25: Is this necessary? Is there any dependency needs this?
Otherwise, you could just update the lock file and leave all these changes to `serde` in Cargo.toml untouched.

Taowyoo:
@mridul-manohar From CI, some errors used in code need to be updated. Please check.

fortanix-vme/aws-nitro-enclaves/eif-tools/Cargo.toml
Pagten: I agree. Don't update the version requirements in the Cargo.toml file if that's not necessary.

Taowyoo (dismissed):

Pagten:
For all crates that didn't need any rust code modifications when upgrading to mbedtls 0.13, we should consider specifying the dependency as
```
mbedtls = { version = ">=0.12.0, <0.14.0", ... }
```
This more accurately presents the actual dependency constraints of these crates and it may ease the process for upgrading them, since consumers won't be forced to simultaneously upgrade to the new mbedtls version.

fortanix-vme/aws-nitro-enclaves/eif-tools/Cargo.toml
mridul-manohar: reverted the serde updates.

mridul-manohar:
> For all crates that didn't need any rust code modifications when upgrading to mbedtls 0.13, we should consider specifying the dependency as
> 
> ```
> mbedtls = { version = ">=0.12.0, <0.14.0", ... }
> ```
> 
> This more accurately presents the actual dependency constraints of these crates and it may ease the process for upgrading them, since consumers won't be forced to simultaneously upgrade to the new mbedtls version.



> For all crates that didn't need any rust code modifications when upgrading to mbedtls 0.13, we should consider specifying the dependency as
> 
> ```
> mbedtls = { version = ">=0.12.0, <0.14.0", ... }
> ```
> 
> This more accurately presents the actual dependency constraints of these crates and it may ease the process for upgrading them, since consumers won't be forced to simultaneously upgrade to the new mbedtls version.

Addressed this.

fortanix-vme/aws-nitro-enclaves/nitro-attestation-verify/Cargo.toml
```diff
@@ -9,7 +9,7 @@ publish = false
 chrono = "0.4"
 serde_cbor = "0.11"
 aws-nitro-enclaves-cose = { version = "0.5.0", default-features = false }
-mbedtls = { version = "0.12", features = ["rdrand", "std", "time", "ssl"], default-features = false, optional = true }
+mbedtls = { version = "0.13.1", features = ["rdrand", "std", "time", "ssl"], default-features = false, optional = true }
```
Taowyoo, line 12:
```suggestion
mbedtls = { version = ">=0.12.0, <0.14.0", features = ["rdrand", "std", "time", "ssl"], default-features = false, optional = true }
```

intel-sgx/pcs/Cargo.toml
```diff
@@ -35,11 +35,12 @@ failure = "0.1.1"
 anyhow = { version = "1", optional = true }
 quick-error = "1.2.3"
 num = "0.2"
-mbedtls = { version = "0.12.3", features = ["std", "time"], default-features = false, optional = true }
+mbedtls = { version = "0.13.2", features = ["std", "time"], default-features = false, optional = true }
```
Taowyoo, line 38:
```suggestion
mbedtls = { version = ">=0.12.0, <0.14.0", features = ["std", "time"], default-features = false, optional = true }
```

Taowyoo (commented):
Since changing error type in most cases is breaking change.
Could you check and bump the version of updated crates ?

fortanix-vme/aws-nitro-enclaves/nitro-attestation-verify/src/lib.rs
```diff
@@ -328,8 +328,8 @@ mod tests {
         //   Not After : Sep  9 13:19:20 2021 GMT
         static ref PROPER_TOKEN : Vec<u8> = include_bytes!("../data/request_proper.bin").to_vec();
         static ref PROPER_VALIDITY: (DateTime<Utc>, DateTime<Utc>) = (Utc.with_ymd_and_hms(2021, 9, 9, 10, 19, 19).unwrap(),  Utc.with_ymd_and_hms(2021, 9, 9, 13, 19, 21).unwrap());
-        static ref NOT_VALID_YET_ERR: NitroError = NitroError::CertificateVerifyFailure("Certificate verify failure: X509CertVerifyFailed, The certificate validity starts in the future\n".to_string());
-        static ref EXPIRED_ERR: NitroError = NitroError::CertificateVerifyFailure("Certificate verify failure: X509CertVerifyFailed, The certificate validity has expired\n".to_string());
+        static ref NOT_VALID_YET_ERR: NitroError = NitroError::CertificateVerifyFailure("Certificate verify failure: HighLevel(X509CertVerifyFailed), The certificate validity starts in the future\n".to_string());
+        static ref EXPIRED_ERR: NitroError = NitroError::CertificateVerifyFailure("Certificate verify failure: HighLevel(X509CertVerifyFailed), The certificate validity has expired\n".to_string());
```
raoulstrackx, line 332: Does this change still work on mbedtls 0.12?

fortanix-vme/aws-nitro-enclaves/nitro-attestation-verify/src/mbedtls.rs
```diff
@@ -97,7 +98,7 @@ impl SigningPublicKey for WrappedCert {
         // We'll throw error if signature verify does not work
         match pk.verify(*md, &digest, &sig) {
             Ok(_) => Ok(true),
-            Err(mbedtls::Error::EcpVerifyFailed) => Ok(false),
+            Err(ErrMbed::HighLevel(codes::EcpVerifyFailed)) => Ok(false),
```
raoulstrackx, line 101: Does this change still work on mbedtls 0.12?

intel-sgx/pcs/src/pckcrt.rs
```diff
@@ -487,13 +487,13 @@ impl PckCert<Verified> {
     }
 
     #[cfg(feature = "verify")]
-    pub fn pck(&self) -> Result<MbedtlsBox<Certificate>, MbedError> {
-        let cert = CString::new(self.cert.as_bytes()).map_err(|_| MbedError::X509InvalidFormat)?;
-        Certificate::from_pem(cert.as_bytes_with_nul()).map_err(|_| MbedError::X509InvalidFormat)
+    pub fn pck(&self) -> Result<MbedtlsBox<Certificate>, ErrMbed> {
+        let cert = CString::new(self.cert.as_bytes()).map_err(|_| ErrMbed::HighLevel(codes::X509InvalidFormat))?;
+        Certificate::from_pem(cert.as_bytes_with_nul()).map_err(|_| ErrMbed::HighLevel(codes::X509InvalidFormat))
```
raoulstrackx, line 492: Does this change still work on mbedtls 0.12?

intel-sgx/pcs/Cargo.toml
```diff
@@ -35,11 +35,12 @@ failure = "0.1.1"
 anyhow = { version = "1", optional = true }
 quick-error = "1.2.3"
 num = "0.2"
-mbedtls = { version = "0.12.3", features = ["std", "time"], default-features = false, optional = true }
+mbedtls = { version = ">=0.12.0, <0.14.0", features = ["std", "time"], default-features = false, optional = true }
 
 [dev-dependencies]
 hex = "0.4.2"
 tempdir = "0.3.7"
+mbedtls = { version = ">=0.12.0, <0.14.0" }
```
raoulstrackx, line 43: Why is this needed?

raoulstrackx (dismissed):
There are many code changes that are (likely) not compatible with mbedtls 0.12. You may be able to call `.into()` to ensure the code keeps working for both versions.

intel-sgx/pcs/Cargo.toml
mridul-manohar: Had added this for mbedtls imports in `mod tests { .. }`
removed this now since it was not needed.

intel-sgx/pcs/Cargo.toml
mridul-manohar: Changed as suggested.

fortanix-vme/aws-nitro-enclaves/nitro-attestation-verify/src/lib.rs
mridul-manohar: This requires v0.13.1 (therefore not backcompt)
and we are pointing to exact `version = "0.13.1"`

fortanix-vme/aws-nitro-enclaves/nitro-attestation-verify/src/mbedtls.rs
mridul-manohar: This requires v0.13.1 (therefore not backcompt)
and we are pointing to exact `version = "0.13.1"` now

intel-sgx/pcs/src/pckcrt.rs
mridul-manohar: This requires v0.13.1 (therefore not backcompt)
and we are pointing to exact `version = "0.13.1"`

mridul-manohar:
> There are many code changes that are (likely) not compatible with mbedtls 0.12. You may be able to call `.into()` to ensure the code keeps working for both versions.

@raoulstrackx thanks for pointing this out. After discussing with @Taowyoo we have decided to point to specific 
`version = "0.13.1"`  for the crates which required code changes for combined error types. This ensures that we don't need to ensure compatibility ( using `.into()` ) with ver0.12

raoulstrackx:
> > There are many code changes that are (likely) not compatible with mbedtls 0.12. You may be able to call `.into()` to ensure the code keeps working for both versions.
> 
> @raoulstrackx thanks for pointing this out. After discussing with @Taowyoo we have decided to point to specific `version = "0.13.1"` for the crates which required code changes for combined error types. This ensures that we don't need to ensure compatibility ( using `.into()` ) with ver0.12

What was the reasoning to force upgrades to mbedtls 0.13.1? It'd be at least a breaking change for `pcs` (I haven't checked the other crates). I'm also a little bit worried about the impact on the fortanixvme target. Just adding `.into()` to fix the errors [should work](https://github.com/fortanix/rust-mbedtls/blob/fb34177568510219b7019b076841f25ad8e1c689/mbedtls/src/error.rs#L153), and would only require a patch number bump.

Taowyoo:
HI @raoulstrackx 
@mridul’s PR is going to upgrade mbedtls in rust-sgx repo to 0.13 
Any crates that directly use mbedtls Error type need to upgrade the dependency requirement on mbedtls  to 0.13  , include:
- nitro-attestation-verify
- pcs

Crates that do not explicitly use mbedtls Error type could use range version ">=0.12.0, <0.14.0" to keep compatibility for 0.12 version

Taowyoo:
I think the main concern from @Raoul Strackx is that bumping mbedtls version in vme crates might cause compile issue in roche.

So I suggest @mridul to have a PR on roche side with changes in rust-sgx  to ensure changes in rust-sgx is fine with roche CI.

Taowyoo (approved):

PR #659, opened 2024-11-21
Insecure time computation

raoulstrackx:
SGXv2 platforms support calling the rdtscp instruction inside an enclave. This PR implements logic to keep track of the time inside of an enclave. Note that the value rdtscp returns, can be manipulated by an attacker; even with these changes insecure_time is (almost) as insecure as before. There's only an option to force the time to be monotonic.

This PR is one of the parts of #648 that was split up.


intel-sgx/insecure-time/src/lib.rs
```diff
@@ -0,0 +1,861 @@
...
+    fn set(&self, t: Ticks) {
+        let t = t.0.load(Ordering::Relaxed);
+        self.0.store(t, Ordering::Relaxed);
+    }
+
+    pub fn from_duration(duration: Duration, freq: &Freq) -> Self {
+        let freq = freq.as_u64();
+        let ticks_secs = duration.as_secs() * freq;
+        let ticks_nsecs = duration.subsec_nanos() as u64 * freq / NANOS_PER_SEC as u64;
+        Ticks::new(ticks_secs + ticks_nsecs)
```
Taowyoo, line 139: I think we need to handle potential overflow here, because `freq` is type of u64.

```suggestion
        let ticks_secs = duration.as_secs().saturating_mul(freq);
        let ticks_nsecs = (duration.subsec_nanos() as u64).saturating_mul(freq) / NANOS_PER_SEC as u64;
        Ticks::new(ticks_secs.saturating_add(ticks_nsecs))
```

zugzwang: Good point, although I'm not sure that `saturating` is what we want. If the `u64::MAX` limit is ever reached (which should already be the result of a bug obviously), then `Ticks` would remain static, e.g., not ticking anymore.

intel-sgx/insecure-time/src/lib.rs
```diff
@@ -0,0 +1,861 @@
...
+        let mut ecx = 0;
+        let mut edx = 0;
+
+        cpuid(0x12, &mut eax, &mut ebx, &mut ecx, &mut edx);
+        eax & (0x1 << 1) != 0x0
+
+    }
+}
+
+/// Frequency in ticks per second
```
zugzwang, line 181: From the type, it is clear that the ticks and seconds are aligned (e.g., there is always an integer amount of ticks per second). Is this always the case? If so, can you add a comment? If not, do we need to handle carry ticks or is this effect negligible?

intel-sgx/insecure-time/src/lib.rs
raoulstrackx: Letting `Ticks` stop ticking would result in incorrect computations. I think we can return an error here, and fallback to doing normal usercalls again. I'll update

intel-sgx/insecure-time/src/lib.rs
raoulstrackx: (Some) hardware does report the frequency of the crystal clock. Interestingly, using that frequency directly results in a small (non negligible) clock drift. To account for that, there's a way to "learn" the correct frequency by requesting the correct wallclock time through a usercall, and reading the value of the TSC (see `TscMode::Learn`). There's also an option to ensure that the clock is re-synced within a specific time interval. That way clock drift is kept within limits.
See also the internal "Insecure time without leaving the enclave" page.

intel-sgx/insecure-time/src/lib.rs
raoulstrackx: I've created a separate commit to address this comment

intel-sgx/insecure-time/src/lib.rs
```diff
@@ -0,0 +1,861 @@
...
+    #[derive(Copy, Clone, PartialOrd, PartialEq)]
+    // Time in nanoseconds since UNIX_EPOCH
+    struct RandTime(u64);
+
+    #[cfg(feature = "std")]
+    impl Add<Duration> for RandTime {
+        type Output = RandTime;
+
+        fn add(self, other: Duration) -> Self::Output {
+            let t = self.0 + other.as_secs() * super::NANOS_PER_SEC as u64 + other.subsec_nanos() as u64;
```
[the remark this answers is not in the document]
raoulstrackx: Resolved in a separate commit

intel-sgx/insecure-time/src/lib.rs
zugzwang: Thanks

intel-sgx/insecure-time/src/lib.rs
```diff
@@ -0,0 +1,887 @@
...
+
+    pub fn as_duration_ex(&self, freq: &Freq) -> Duration {
+        let freq = freq.as_u64();
+        let ticks = self.0.load(Ordering::Relaxed);
+
+        let time_secs = ticks / freq;
+        let time_nsecs = (ticks % freq * NANOS_PER_SEC) / freq;
+        let time_nsecs: u32 = time_nsecs.try_into().expect("must be smaller than 1sec");
+
+        Duration::new(time_secs, time_nsecs)
```
zugzwang, line 170: Actually, you could only compute `time_nsecs` and use [`Duration::from_nanos`](https://doc.rust-lang.org/std/time/struct.Duration.html#method.from_nanos)

raoulstrackx: I assume you're thinking of:
```
pub fn as_duration_ex(&self, freq: &Freq) -> Duration {
    let freq = freq.as_u64();
    let ticks = self.0.load(Ordering::Relaxed);
    Duration::from_nanos(ticks * NANOS_PER_SEC / freq)
}
```
I'd rather not do that. Doing `ticks * NANOS_PER_SEC` could result in an overflow.

zugzwang (approved):
Thanks for the PR!

PR #660, opened 2024-11-22
Support letting enclaves do time keeping

raoulstrackx:
SGXv2 platforms support calling the rdtscp instruction inside an enclave. This PR piggy backs on #659 that implemented logic to keep track of time. This PR starts using this logic in various tools and crates (e.g., async-usercalls, enclave-runner, ...).
Note that the enclave-runner _does not_ change the returned value of `insecure_time` usercalls at this point; it will always return a null pointer as the second value in the returned tuple. These changes can thus be used on all existing enclaves, no new Rust compiler is required. Another PR will be filed to change the default behavior of enclave runner so enclaves by default take advantage of their improved hardware capabilities.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1022,6 +1022,7 @@ dependencies = [
  "fnv",
  "fortanix-sgx-abi",
  "futures 0.3.17",
+ "insecure-time",
  "ipc-queue",
  "lazy_static",
  "libc",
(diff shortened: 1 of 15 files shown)


mzohreva (approved):

Taowyoo (dismissed):

Taowyoo (approved):

jethrogb:
Please add doc comments for the changes in fortanix-sgx-abi

PR #665, opened 2024-12-06
Add `Fmspc` type in `pcs` crate

mzohreva:


intel-sgx/pcs/src/tcb_info.rs
Taowyoo, line 1: Do you also want to update the type of `fmspc` field in `TcbData`

Taowyoo: And also the `fmspc` field in `SGXPCKCertificateExtension` ?
And missing functions for `PckCert<Verified>` ?
```rust
   pub fn ppid(&self) -> Result<Vec<u8>, ASN1Error> {
        let extension = self.sgx_extension()?;
        Ok(extension.ppid)
    }

    pub fn fmspc(&self) -> Result<Fmspc, ASN1Error> {
        let extension = self.sgx_extension()?;
        Ok(extension.fmspc)
    }
```

mzohreva: > Do you also want to update the type of `fmspc` field in `TcbData`

Let's just keep this as is. There might have been a good reason to change that field's type to `String`.

mzohreva: > And also the `fmspc` field in `SGXPCKCertificateExtension` ?
It's already there with type `Vec<u8>`. I think that's good enough.

> And missing functions for `PckCert<Verified>` ?
No, since `fn sgx_extension` is already public and the `ppid` and `fmspc` fields in `SGXPCKCertificateExtension` are also public, these two functions are not necessary

Taowyoo (approved):

PR #669, opened 2024-12-12
dcap-artifact-retrieval: add CLI option for specifying PCS API version and format code

mzohreva:
This is needed for PROD-7165.
I've also moved the logic in `main.rs` to a library module which would be helpful in RTE-121.
Most of the changes are formatting changes. You can see the top commit separately for actual changes.


intel-sgx/dcap-artifact-retrieval/src/cli.rs
```diff
@@ -0,0 +1,223 @@
...
+                }
+            };
+            download_dcap_artifacts(&*client, pckid_file, output_dir, 0 < verboseness)
+        }
+        _ => unreachable!("validated"),
+    };
+
+    match result {
+        Ok(()) => {}
+        Err(Error::PCSError(StatusCode::NotFound, _)) => {
```
Taowyoo, line 214: Nit: I suggest to also print error.

Taowyoo (dismissed):

intel-sgx/dcap-artifact-retrieval/src/cli.rs
mzohreva: > Nit: I suggest to also print error.

I don't think there is more information in the message parameter after checking the code.

Taowyoo (approved):

PR #677, opened 2025-01-06
dcap-ql: Fix length validation logic for quote signatures

### LINKED ISSUE #676: dcap-ql: should accept extra trailing data in quote
**Describe the bug:**

we've recently receive following error, which is generated in this line (https://github.com/fortanix/rust-sgx/blob/master/intel-sgx/dcap-ql/src/quote.rs#L240)
`Invalid signature length, expected 4164, got 4168`

After some investigation, following is the root cause

The quote function in dcap-ql which generate quote contains following two steps
1. get_quote_size, to get an ESTIMATED size of the quote, and it is the application's (which is our) responsibility to allocate a buffer that is AT LEAST this big, and pass to get_quote function. Since this is just estimation, it has some default value if PCCS return empty data / failed to return data at the moment
2. get_quote

The erroneous signature that trigger the error above is actually still valid, just that there are four trailing '\x00' at the end of the quote, which should be caused by a larger estimated quote size
get quote function looks like this, and the `let mut quote = vec![0; quote_size as _];` should be the root cause of those trailing '\x00'

```
pub fn quote(report: &Report) -> Result<Vec<u8>, Quote3Error> {
    unsafe {
        let mut quote_size = 0;
        err_code_to_result(get_quote_size(&mut quote_size))?;

        let mut quote = vec![0; quote_size as _];
        err_code_to_result(get_quote(&report, quote_size, quote.as_mut_ptr()))?;
        Ok(quote)
    }
}
```

And when parsing the signature in fortanix's library, it requires the input buffer (which is vec![0; quote_size as _] to match the exact content (which is represented by format [length_of_following_bytes, byte_1, byte_2, ...], e.g. [0x5, 0xab, 0xcd, 0xef, 0x01, 0x23])

And the fix is simple, instead of failing when the expected quote length and actual quote length mismatch, we just keep parsing, and discard any trailing data (so effectively accepting longer quote, but will still reject quote that is shorter than expected)

**To Reproduce:**
1. Append several random bytes after getting quote from `dcap_ql::quote`
2. Parse the quote using `dcap_ql` crate

**Expected behavior:**
Extra bytes are ignored, and parsing passed

**Reproducibility:**
 - Always

**Environment:**
- Ubuntu 22.04 LTS

**Possible Solution:**
Parse the quote as normal, but instead of failing directly after detecting extra trailing bytes, simply discard them

**Severity:**
 - [ ] Critical
 - [x] Major
 - [ ] Normal
 - [ ] Minor

mzohreva:
This should fix #676.

### DIFF
--- a/intel-sgx/dcap-ql/src/quote.rs
+++ b/intel-sgx/dcap-ql/src/quote.rs
@@ -159,7 +159,9 @@ impl<'a> TakePrefix for Cow<'a, str> {
 }
 
 pub trait Quote3Signature<'a>: Sized {
-    fn parse(type_: Quote3AttestationKeyType, data: Cow<'a, [u8]>) -> Result<Self>;
+    /// Parse the signature encoded in data and return the parsed value along
+    /// with any trailing data that was not part of the signature.
+    fn parse(type_: Quote3AttestationKeyType, data: Cow<'a, [u8]>) -> Result<(Self, Cow<'a, [u8]>)>;
 }
 
 pub trait Qe3CertData<'a>: Sized {
@@ -231,7 +233,7 @@ fn get_ecdsa_sig_der(sig: &[u8]) -> Result<Vec<u8>> {
 }
 
 impl<'a> Quote3Signature<'a> for Quote3SignatureEcdsaP256<'a> {
-    fn parse(type_: Quote3AttestationKeyType, mut data: Cow<'a, [u8]>) -> Result<Self> {
+    fn parse(type_: Quote3AttestationKeyType, mut data: Cow<'a, [u8]>) -> Result<(Self, Cow<'a, [u8]>)> {
         if type_ != Quote3AttestationKeyType::EcdsaP256 {
             bail!("Invalid attestation key type: {:?}", type_)
         }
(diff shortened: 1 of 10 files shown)


Taowyoo (dismissed):

jethrogb (changes requested):
Trailing input data should not be silently ignored.

intel-sgx/dcap-ql/src/quote.rs
```diff
@@ -236,14 +236,19 @@ impl<'a> Quote3Signature<'a> for Quote3SignatureEcdsaP256<'a> {
...
+                "Invalid signature length. Got {}, expected >= {}",
                 data.len(),
                 sig_len
             );
         }
+        // NOTE: data may contain trailing zeros due to `get_quote_size` and
+        // `get_quote` C APIs allowing larger than necessary buffer to be
+        // allocated to hold the quote.
+        data = data.take_prefix(cmp::min(data.len(), sig_len))?;
+
```
jason-liang-vault, line 251: Probably can have a warning log here if `data.len() != 0`?

mzohreva: I have a better idea, see the new commit.

intel-sgx/dcap-ql/src/quote.rs
```diff
@@ -538,11 +550,15 @@ impl<'a> Quote3SignatureVerify<'a> for Quote3SignatureEcdsaP256<'a> {
 
 impl<'a> Quote<'a> {
     #[cfg(feature = "verify")]
-    pub fn verify<T: Quote3SignatureVerify<'a>>(quote: &'a [u8], root_of_trust: T::TrustRoot) -> Result<Self> {
+    /// Parses the `Quote` encoded in `quote`, verifies the signature within the
+    /// quote, and returns the parsed quote along with any trailing data that
+    /// was not part of the signature. It's up to the caller to decide what to
+    /// do with the trailing bytes.
```
jethrogb, line 556: This voluntary behavior is not acceptable - the API should such that it's difficult to misuse.

PR #678, opened 2025-01-11
[DEVOPS-5452] Support choose native-tls or rustls-tls for dcap-artifact-retrieval

Taowyoo:
Original `native-tls` in `dcap-artifact-retrieval` wont work in distroless container where no CA is provided by system.
It will throw some error like this:
```
FetcherFailure("error sending request for url (...), caused by: client error (Connect), caused by: error:1416F086:SSL routines:tls_process_server_certificate:certificate verify failed:../ssl/statem/statem_clnt.c:1913: (unable to get local issuer certificate), caused by: error:1416F086:SSL routines:tls_process_server_certificate:certificate verify failed:../ssl/statem/statem_clnt.c:1913:")
```

The solution I proposed here is to adding support for choosing  `rustls` as TLS provider for `reqwest`, `rustls` will use [webpki-roots](https://crates.io/crates/webpki-roots) for default root CAs.
Also the original `native-tls`  becomes addictive feature but remain default.


Taowyoo:
closed, since PR from fork cannot read secret set in upstream. creating  another PR

PR #680, opened 2025-01-13
 [DEVOPS-5452] Support different tls for dcap-artifact-retrieval

Taowyoo:
## Background

Original `native-tls` in `dcap-artifact-retrieval` wont work in distroless container where no root CA is
provided by system.

## Change

The solution I proposed here is adding support for choosing  `rustls` as TLS provider for `reqwest`, `rustls` will use [webpki-roots](https://crates.io/crates/webpki-roots) for default root CAs.

- Add feature `rustls-tls`
- Add/update functions for creating a reqwest client.
- Add test cases in CI for dcap-artifact-retrieval using rustls.

### DIFF
--- a/intel-sgx/dcap-artifact-retrieval/src/lib.rs
+++ b/intel-sgx/dcap-artifact-retrieval/src/lib.rs
@@ -113,7 +113,8 @@ quick_error! {
 
 pub type Result<T> = std::result::Result<T, Error>;
 
-#[cfg(feature = "reqwest")]
+/// Create a reqwest client using native tls.
+#[cfg(all(feature = "reqwest", feature = "native-tls"))]
 pub fn reqwest_client() -> ReqwestClient {
     ReqwestClient::builder()
         .use_native_tls()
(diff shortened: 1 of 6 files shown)


intel-sgx/dcap-artifact-retrieval/src/lib.rs
```diff
@@ -113,15 +113,25 @@ quick_error! {
...
+    ReqwestClient::builder()
+        .use_rustls_tls()
+        .build()
+        .expect("Failed to build reqwest client")
+}
+
+#[cfg(all(feature = "reqwest", feature = "native-tls"))]
 #[doc(hidden)]
 pub fn reqwest_client_insecure_tls() -> ReqwestClient {
     ReqwestClient::builder()
```
mzohreva, line 137: Not having a rustls version of this means that tests depend on the native-tls feature but I don't think the cfg annotations on the tests reflect that

Taowyoo: Nice catch. Just fix this by simplify the PR to only add support of adding `rustls-tls`  instead of selecting one of two.

intel-sgx/dcap-artifact-retrieval/src/lib.rs
```diff
@@ -121,6 +122,15 @@ pub fn reqwest_client() -> ReqwestClient {
         .expect("Failed to build reqwest client")
 }
 
+/// Create a reqwest client using rustls tls.
+#[cfg(all(feature = "reqwest", feature = "rustls-tls"))]
```
mzohreva, line 126: I suggest the following API:
- A `pub fn reqwest_client_native_tls()`
- A `pub fn reqwest_client_rustls()`
- A new `pub fn reqwest_client()` that chooses to call one of the above based on enabled features, e.g. call `reqwest_client_rustls()` if `rustls-tls` is enabled, otherwise call the native-tls version

Taowyoo: Added, also apply same logic change on `fn reqwest_client()` to `fn reqwest_client_insecure_tls()`

mzohreva:
Please update the PR description (first comment) to reflect the new changes.

mzohreva (approved):

PR #683, opened 2025-01-21
[RTE-241] Always fetch early variant of DCAP artifacts

raoulstrackx:
By default we want to fetch the "early" variant of DCAP artifacts. See the [pcs](https://api.portal.trustedservices.intel.com/content/documentation.html#pcs-tcb-info-model-v3) API documentation.

### DIFF
--- a/intel-sgx/dcap-artifact-retrieval/src/provisioning_client/intel.rs
+++ b/intel-sgx/dcap-artifact-retrieval/src/provisioning_client/intel.rs
@@ -355,7 +355,7 @@ impl<'inp> ProvisioningServiceApi<'inp> for TcbInfoApi {
         let api_version = input.api_version as u8;
         let fmspc = input.fmspc.as_bytes().to_hex();
         let url = format!(
-            "{}/sgx/certification/v{}/tcb?fmspc={}",
+            "{}/sgx/certification/v{}/tcb?fmspc={}&update=early",
             INTEL_BASE_URL, api_version, fmspc,
         );
         Ok((url, Vec::new()))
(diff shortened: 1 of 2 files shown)


intel-sgx/dcap-artifact-retrieval/src/provisioning_client/intel.rs
```diff
@@ -430,7 +430,7 @@ impl<'inp> ProvisioningServiceApi<'inp> for QeIdApi {
     fn build_request(&self, input: &Self::Input) -> Result<(String, Vec<(String, String)>), Error> {
         let api_version = input.api_version as u8;
         let url = format!(
-            "{}/sgx/certification/v{}/qe/identity",
+            "{}/sgx/certification/v{}/qe/identity?update=early",
```
mzohreva, line 433: I think these need to be added in PCCS as well?

raoulstrackx: Good catch! I've updated the last commit.

mzohreva: Does this apply to Azure as well?

mzohreva (approved):

intel-sgx/dcap-artifact-retrieval/src/provisioning_client/intel.rs
raoulstrackx: The Azure client fetches the TcbInfo and QeIdentity data from Intel and thus also pull in the changes in this PR.

PR #685, opened 2025-01-21
[RTE-166] Verify tcb info dates

raoulstrackx:


intel-sgx/pcs/src/lib.rs
```diff
@@ -89,6 +90,9 @@ quick_error! {
         InvalidQe3Id(err: MbedError){
             display("Invalid QE3 ID: {}", err)
         }
+        Qe3IdNotValid(err: String){
```
jethrogb, line 93:
```suggestion
        Qe3NotValid(err: String){
```

intel-sgx/pcs/src/lib.rs
```diff
@@ -160,6 +164,16 @@ fn intel_signature_deserializer<'de, D: Deserializer<'de>>(deserializer: D) -> R
     crate::get_ecdsa_sig_der(signature).map_err(|_| de::Error::custom("Failed ECDSA signature conversion"))
 }
 
+fn iso8601timestamp<'de, D: Deserializer<'de>>(deserializer: D) -> Result<DateTime<Utc>, D::Error> {
+    // Same as %Y-%m-%dT%H:%M:%S%.f%:z, i.e. 0, 3, 6 or 9 fractional digits for seconds and colons in the time zone offset.
+    // This format also supports having a Z or UTC in place of %:z. They are equivalent to +00:00.
+    // https://docs.rs/chrono/0.4.39/chrono/format/strftime/index.html#fn7
+    const ISO8601_FORMAT: &'static str = "%+";
```
jethrogb, line 171: Not sure if having to explain this shorthand in a comment is better than just putting the full format string.

intel-sgx/pcs/src/tcb_info.rs
```diff
@@ -377,7 +391,10 @@ mod tests {
             TcbInfo::restore("./tests/data/", &Fmspc::try_from("00906ea10000").expect("static fmspc")).expect("validated");
         let root_certificate = include_bytes!("../tests/data/root_SGX_CA_der.cert");
         let root_certificates = [&root_certificate[..]];
-        assert!(info.verify(&root_certificates).is_ok());
+        match info.verify(&root_certificates) {
+            Err(Error::InvalidTcbInfo(msg)) => assert_eq!(msg, String::from("TCB Info expired on 2020-06-17 17:49:24 UTC")),
+            e => assert!(false, "wrong result: {:?}", e),
```
Taowyoo, line 396: I suggest adding a round trip test for serialization and deserialization

intel-sgx/pcs/src/qe_identity.rs
```diff
@@ -226,6 +231,15 @@ where
     serializer.serialize_str(&miscselect)
 }
 
+fn datetime_serializer<S>(timestamp: &DateTime<Utc>, serializer: S) -> ::std::result::Result<S::Ok, S::Error>
```
jethrogb, line 234: Can you put this close to `iso8601timestamp` and use the same format string constant in both? `#[serde(with = "module")]` may be a good way to do this.

intel-sgx/pcs/src/qe_identity.rs
```diff
@@ -352,13 +366,24 @@ impl QeIdentitySigned {
...
-        })
+        };
+
+        let now = Utc::now();
+        if now < identity.issue_date {
+            return Err(Error::Qe3IdNotValid(format!("QE3 only valid from {}", identity.issue_date)))
+        }
+        if identity.next_update < now {
+            return Err(Error::Qe3IdNotValid(format!("QE3 expired on {}", identity.next_update)))
+        }
```
jethrogb, line 377: This seems like a weird location for this check. Why don't you do this check prior to constructing the `Verified` struct, like the version check?

intel-sgx/pcs/src/tcb_info.rs
```diff
@@ -353,7 +356,17 @@ impl TcbInfo {
...
-        })
+        };
+
+        let now = Utc::now();
+        if now < data.issue_date {
+            return Err(Error::InvalidTcbInfo(format!("TCB Info only valid from {}", data.issue_date)))
+        }
+        if data.next_update < now {
+            return Err(Error::InvalidTcbInfo(format!("TCB Info expired on {}", data.next_update)))
+        }
```
jethrogb, line 367: Should be done prior to construction of the `Verified` struct.

jethrogb (changes requested):

intel-sgx/pcs/src/tcb_info.rs
```diff
@@ -135,8 +136,8 @@ fn sgx_platform() -> Platform {
 pub struct TcbData<V: VerificationType = Verified> {
     id: Platform,
     version: u16,
-    issue_date: String,
-    next_update: String,
+    issue_date: DateTime<Utc>,
+    next_update: DateTime<Utc>,
```
mzohreva, line 140: There is a foot gun here, if we add `derive(Serialize)` for this in the future and forget to annotate the fields with `#[serde(with = "crate::iso8601")]`. Why not add a wrapper type for serialization/deserialization logic?

raoulstrackx: Adding `derive(Serialize)` here may be an issue in itself as we wouldn't be able to verify the struct anymore when the serialized data doesn't match exactly with how Intel created it. The `TcbInfo` already derives serialize and avoids that issue by storing the `raw_tcb_info` as a `String`

raoulstrackx: Let me merge the PR now. We can sort this out in another PR if we reach another conclusion.

PR #686, opened 2025-01-23
Fix `pckcerts_with_fallback` logic to retrieve TCB info and get PCK certs for each TCB level

mzohreva:
This should address PROD-9810.

### DIFF
--- a/intel-sgx/dcap-artifact-retrieval/src/provisioning_client/mod.rs
+++ b/intel-sgx/dcap-artifact-retrieval/src/provisioning_client/mod.rs
@@ -5,6 +5,7 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/.
  */
 
+use std::collections::HashMap;
 use std::convert::{TryFrom, TryInto};
 use std::hash::{DefaultHasher, Hash, Hasher};
 use std::io::Read;
@@ -534,10 +535,13 @@ pub trait ProvisioningClient {
 
     fn qe_identity(&self) -> Result<QeIdentitySigned, Error>;
 
-    /// Retrieve PCK certificates using `pckcerts()` and fallback to `pckcert()`
-    /// if provisioning client does not support pckcerts. Note that in case of
-    /// fallback the returned PckCerts will only contain a single certificate
-    /// associated with the highest TCB level applied to the platform.
+    /// Retrieve PCK certificates using `pckcerts()` and fallback to the
+    /// following method if that's not supported:
+    /// - Call `pckcert()` to find the FMSPC.
+    /// - Using the FMSPC value, call `tcbinfo()` to get TCB info.
+    /// - For each TCB level in the result of previous call:
+    ///   - Call `pckcert()` to get the best available PCK cert for that TCB level.
+    /// Note that PCK certs for some TCB levels may be missing.
     fn pckcerts_with_fallback(&self, pck_id: &PckID) -> Result<PckCerts, Error> {
         match self.pckcerts(&pck_id.enc_ppid, pck_id.pce_id) {
             Ok(pck_certs) => return Ok(pck_certs),
@@ -545,6 +549,9 @@ pub trait ProvisioningClient {
             Err(e) => return Err(e),
         }
         // fallback:
+
+        // NOTE: at least with PCCS, any call to `pckcert()` will return the
+        // "best available" PCK cert for the specified TCB level.
         let pck_cert = self.pckcert(
             Some(&pck_id.enc_ppid),
             &pck_id.pce_id,
@@ -552,8 +559,27 @@ pub trait ProvisioningClient {
             pck_id.pce_isvsvn,
             Some(&pck_id.qe_id),
         )?;
-
-        pck_cert
+        let fmspc = pck_cert.sgx_extension()?.fmspc;
+        let tcb_info = self.tcbinfo(&fmspc)?;
+        let tcb_data = tcb_info.data()?;
+        let mut pcks = HashMap::new();
+        for (cpu_svn, pce_isvsvn) in tcb_data.iter_tcb_components() {
+            let p = match self.pckcert(
+                Some(&pck_id.enc_ppid),
+                &pck_id.pce_id,
+                &cpu_svn,
+                pce_isvsvn,
+                Some(&pck_id.qe_id),
+            ) {
+                Ok(cert) => cert,
+                Err(Error::PCSError(StatusCode::NotFound, _)) |
+                Err(Error::PCSError(StatusCode::NonStandard462, _)) => continue,
+                Err(other) => return Err(other)
+            };
+            pcks.insert(p.platform_tcb()?.cpusvn, p);
+        }
+        let pcks: Vec<_> = pcks.into_iter().map(|(_, v)| v).collect();
+        pcks
             .try_into()
             .map_err(|e| Error::PCSDecodeError(format!("{}", e).into()))
     }
--- a/intel-sgx/dcap-artifact-retrieval/src/provisioning_client/pccs.rs
+++ b/intel-sgx/dcap-artifact-retrieval/src/provisioning_client/pccs.rs
@@ -538,6 +538,42 @@ mod tests {
 
     #[test]
     #[ignore = "needs a running PCCS service"] // FIXME
+    pub fn test_pckcerts_with_fallback() {
+        for api_version in [PcsVersion::V3, PcsVersion::V4] {
+            let client = make_client(api_version);
+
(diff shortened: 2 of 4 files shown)


intel-sgx/dcap-artifact-retrieval/src/provisioning_client/pccs.rs
```diff
@@ -536,6 +536,42 @@ mod tests {
         }
     }
 
+    #[test]
+    #[ignore = "needs a running PCCS service"] // FIXME
+    pub fn test_pckcerts_with_fallback() {
```
Taowyoo, line 541: Do we have similar tests for fallback for Intel?

Taowyoo: Never mind, intel api impl does not need to fallback

intel-sgx/dcap-artifact-retrieval/src/provisioning_client/mod.rs
```diff
@@ -552,8 +554,22 @@ pub trait ProvisioningClient {
...
+            let pce_isvsvn = tcb_level.tcb.pce_svn();
+            pcks.push(self.pckcert(
+                Some(&pck_id.enc_ppid),
+                &pck_id.pce_id,
+                &cpu_svn,
+                pce_isvsvn,
+                Some(&pck_id.qe_id),
+            )?);
+        }
+        pcks
```
Taowyoo, line 572: How could we assume they share same CA chain?
I think we should mention this assumption into function comment.

Taowyoo (approved):

intel-sgx/dcap-artifact-retrieval/src/provisioning_client/mod.rs
mzohreva: The assumption is fine in practice. I've already documented it on the impl block which shows up in rendered docs:
```
impl<V> TryFrom<Vec<PckCert<V>>> for PckCerts
where
    V: VerificationType,

NOTE: This conversion is only correct if all PCK certs in the vec have the same CA chain.

type Error = ASN1Error

The type returned in the event of a conversion error.

fn try_from(pcks: Vec<PckCert<V>>) -> Result<PckCerts, ASN1Error>

Performs the conversion.
```

intel-sgx/dcap-artifact-retrieval/src/provisioning_client/mod.rs
```diff
@@ -534,10 +534,12 @@ pub trait ProvisioningClient {
...
-    /// if provisioning client does not support pckcerts. Note that in case of
-    /// fallback the returned PckCerts will only contain a single certificate
-    /// associated with the highest TCB level applied to the platform.
+    /// Retrieve PCK certificates using `pckcerts()` and fallback to the
+    /// following method if that's not supported:
+    /// - Call `pckcert()` to find the FMSPC.
+    /// - Using the FMSPC value, call `tcbinfo()` to get TCB info.
+    /// - For each TCB level in the result of previous call:
+    ///   - Call `pckcert()` to get the PCK cert associated with that TCB level.
     fn pckcerts_with_fallback(&self, pck_id: &PckID) -> Result<PckCerts, Error> {
```
raoulstrackx, line 543: Do we still need a special function for that? Can't we just implement the fallback when it's needed?

intel-sgx/dcap-artifact-retrieval/src/provisioning_client/mod.rs
```diff
@@ -552,8 +554,22 @@ pub trait ProvisioningClient {
...
-
-        pck_cert
+        let fmspc = pck_cert.sgx_extension()?.fmspc;
+        let tcb_info = self.tcbinfo(&fmspc)?;
+        let tcb_data = tcb_info.data()?;
+        let mut pcks = Vec::new();
+        for tcb_level in tcb_data.tcb_levels() {
+            let cpu_svn = tcb_level.tcb.cpu_svn();
+            let pce_isvsvn = tcb_level.tcb.pce_svn();
+            pcks.push(self.pckcert(
```
raoulstrackx, line 564: For multipackage platforms, there may not be a pckcert for every tcb level. An error in getting a single `pckcert` should not result in an error for the overall get all pckcerts.

intel-sgx/pcs/src/pckcrt.rs
```diff
@@ -743,6 +748,26 @@ where
...
+impl<V> TryFrom<Vec<PckCert<V>>> for PckCerts
+where
+    V: VerificationType,
+{
+    type Error = ASN1Error;
+
+    fn try_from(mut pcks: Vec<PckCert<V>>) -> Result<PckCerts, ASN1Error> {
+        let pck_data = pcks.iter().map(|pck| pck.as_pck_cert_body_item()).collect::<Result<Vec<_>, ASN1Error>>()?;
+        // NOTE: assuming that all PCK certs in the vec have the same CA chain,
+        // so we pick the ca_chain from the first one:
```
raoulstrackx, line 762: So what happens when an intermediate CA changes (e.g., one expired, or got revoked)? The PckCerts won't have a valid certificate chain anymore? Its not clear that Azure would register there old platform configurations again, just to get new a PckCert for each one. We may need to change how `PckCerts` stores this chain internally, if we fully want to emulate the API.

intel-sgx/pcs/src/tcb_info.rs
```diff
@@ -221,7 +221,7 @@ impl TcbData<Unverified> {
 }
 
 impl<V: VerificationType> TcbData<V> {
-    pub(crate) fn tcb_levels(&self) -> &Vec<TcbLevel> {
+    pub fn tcb_levels(&self) -> &Vec<TcbLevel> {
```
raoulstrackx, line 224: Can you return an iterator over the cpusvn and pcesvn for each level? The whole point of using type states here is to prevent us from accessing the TCB levels without checking whether the TcbInfo is valid in the first place. Making this function public, completely breaks that logic.

intel-sgx/dcap-artifact-retrieval/src/provisioning_client/mod.rs
mzohreva: This is that implementation, whether or not it's used is up to the caller. I think we have two places we need to call this, that's why I have implemented it here to avoid duplication.

intel-sgx/pcs/src/pckcrt.rs
mzohreva: I think this is an issue with the PCK Certs API in general since the issuer chain is returned as a separate header and not specific to each PCK cert, see `SGX-PCK-Certificate-Issuer-Chain` header in https://api.portal.trustedservices.intel.com/content/documentation.html#pcs-certificates-v4-response

intel-sgx/pcs/src/tcb_info.rs
mzohreva: The intention to keep these private should be documented.

intel-sgx/dcap-artifact-retrieval/src/provisioning_client/mod.rs
mzohreva: That's true with Intel PCS. However, with PCCS if there are any PCKs cached for the platform, PCCS will return a "best available" PCK cert. I don't know the behavior for Azure though. Is there a reference I could check?

intel-sgx/dcap-artifact-retrieval/src/provisioning_client/mod.rs
raoulstrackx: Oh you can't just always use the `pckcerts` function, because you need more info.

intel-sgx/dcap-artifact-retrieval/src/provisioning_client/mod.rs
raoulstrackx: And what happens when there isn't such an "best available" PCK cert? It can't just make up a `PckCert`, so I guess it must through an error, but that's not what you want here.

raoulstrackx: A "best available" PCK cert also implies that we need to check what the tcb level is for the pck you get back.

intel-sgx/pcs/src/pckcrt.rs
raoulstrackx: Yes, but Intel could easily sign these Pck certs with a different intermediate CA if they want to. It's an issue now because we need to emulate that API. But maybe this is something for another PR.

Taowyoo (approved):

PR #681, opened 2025-01-15
Bump `enclave_runner` version to `0.6.1`

s-arash:
According to `cargo semver-check` this is a semver-compatible change.

Disclosure: the following dependencies of `enclave-runner` have been updated (since v0.6.0 release of `enclave-runner`):
- `fortanix-sgx-abi`: `0.5.0` -> `0.6.0`
- `ipc_queue`: `0.2.0` -> `0.3.0`

PR #690, opened 2025-02-04
Bump `enclave-runner` to `v0.7.0` (also bump dependent crates `sgx-tools`, `report-test`, `dcap-provider`, and `dcap-retrieve-pkcid`)

s-arash:
Also:
- sgxs-tools: 0.8.6 -> 0.9.0
- report-test: 0.4.0 -> 0.5.0
- dcap-provider: 0.4.0 -> 0.5.0
- dcap-retrieve-pckid: 0.2.0 -> 0.3.0

These are all published crates that (transitively) depend on `enclave-runner`, so we should bump their versions as well, and publish new versions of them if/when needed.

See this comment: https://github.com/fortanix/rust-sgx/pull/681#issuecomment-2616354292 for why we decided to bump enclave runner to a semver-incompatible version.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -783,7 +783,7 @@ dependencies = [
 
 [[package]]
 name = "dcap-provider"
-version = "0.4.0"
+version = "0.5.0"
 dependencies = [
  "byteorder 1.3.4",
  "dcap-ql",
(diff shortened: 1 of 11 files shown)


Taowyoo (commented):
lgtm, but could you update your PR title to aligh with changes in PR.
Not only enclave-runner's version is bumped.

Taowyoo (approved):

Issue #565, opened 2024-03-08
[BUG]: "I/O ctl failed" in sgx-detect


ndokmai:
**Describe the bug:**
`sgx-detect` reports the following error:

> $ sgx-detect --verbose
> Detecting SGX, this may take a minute...
> ✔  SGX instruction set
>   ✔  CPU support
>   ✔  CPU configuration
>   ✔  Enclave attributes
>   ✔  Enclave Page Cache
>   SGX features
>     ✔  SGX2  ✔  EXINFO  ✘  ENCLV  ✘  OVERSUB  ✔  KSS
>     Total EPC size: 127.2GiB (no integrity protection)
> ✔  Flexible launch control
>   ✔  CPU support
>   ？ CPU configuration
>   ✔  Able to launch production mode enclave
> ✘  SGX system software
>   ✔  SGX kernel device (/dev/sgx_enclave)
>   ✔  libsgx_enclave_common
>   ✔  AESM service
>   ✘  Able to launch enclaves
>     ✘  Debug mode
>     ✔  Production mode
> 
> 🕮  SGX system software > Able to launch enclaves > Debug mode
> The enclave could not be launched.
> 
> debug: failed to load report enclave
> debug: cause: Failed to call ECREATE.
> debug: cause: I/O ctl failed.
> debug: cause: Input/output error (os error 5)

Running the test on the most recent commit of `sgxs-loaders` also produces similar results:

> $ git clone https://github.com/fortanix/rust-sgx.git
> $ cd rust-sgx/intel-sgx/sgxs-loaders/
> $ RUST_BACKTRACE=1 cargo test sgxdevice
> 

>      Running tests/loaders.rs (/home/nd549/rust-sgx/target/debug/deps/loaders-2b03394f05fc64eb)
> 
> running 1 test
> test sgxdevice ... FAILED
> 
> failures:
> 
> ---- sgxdevice stdout ----
> thread 'sgxdevice' panicked at intel-sgx/sgxs-loaders/tests/loaders.rs:30:5:
> error getting report: Create(Io(Os { code: 5, kind: Uncategorized, message: "Input/output error" }))
> 
>    0: failure::backtrace::internal::InternalBacktrace::new
>              at /home/nd549/.cargo/registry/src/index.crates.io-6f17d22bba15001f/failure-0.1.8/src/backtrace/internal.rs:46:44
>    1: failure::backtrace::Backtrace::new
[...]

PR #694, opened 2025-02-17
[RTE-370] Generate valid xfrm when required

raoulstrackx:
There are two issues with running SGX enclaves on Sapphire Rapids:

- When running debug enclaves, we used to use XCR0 unmodified for the xfrm attribute. But now advanced matrix extensions (AMX) may lead to larger SSA frames (3 pages) than expected. The elf2sgxs tool by default only selected 1 page for SSA frames. When generating a signature, we need to ensure features are turned off when they would lead to larger SSA frames.
- Turning features off, may conflict with one another. XTILECFG and XTILEDATA need to be either enabled both, or disabled both. This PR sets the xfrm attribute accordingly.

Address #565

### DIFF
--- a/intel-sgx/enclave-runner/src/loader.rs
+++ b/intel-sgx/enclave-runner/src/loader.rs
@@ -196,16 +196,9 @@ impl<'a> EnclaveBuilder<'a> {
                     return 0;
                 }
 
-                let (base, size) = {
-                    let cpuid = cpuid(0x0d, bit);
-                    let base = cpuid.map_or(0, |c| c.ebx);
-
-                    if base == 0 {
-                        (0, 0)
-                    } else {
-                        let size = cpuid.map_or(0, |c| c.eax);
-                        (base, size)
-                    }
+                let CpuidResult { ebx: base, eax: size, .. } = match cpuid(0x0d, bit) {
+                    None | Some(CpuidResult { ebx: 0, .. }) => return 0,
+                    Some(v) => v,
                 };
 
                 if max_ssaframesize_in_pages * 0x1000 < base + size {
--- a/intel-sgx/enclave-runner/src/loader.rs
+++ b/intel-sgx/enclave-runner/src/loader.rs
@@ -165,61 +165,67 @@ impl<'a> EnclaveBuilder<'a> {
         ret
     }
 
-    fn cpuid(eax: u32, ecx: u32) -> Option<CpuidResult> {
-        unsafe {
-            if eax <= x86_64::__get_cpuid_max(0).0 {
-                Some(x86_64::__cpuid_count(eax, ecx))
-            } else {
-                None
+    fn generate_xfrm(max_ssaframesize_in_pages: u32) -> u64 {
+        fn cpuid(eax: u32, ecx: u32) -> Option<CpuidResult> {
+            unsafe {
+                if eax <= x86_64::__get_cpuid_max(0).0 {
+                    Some(x86_64::__cpuid_count(eax, ecx))
+                } else {
+                    None
+                }
             }
         }
-    }
 
-    fn generate_xfrm(max_ssaframesize_in_pages: u32) -> u64 {
         fn xgetbv0() -> u64 {
             unsafe { arch::x86_64::_xgetbv(0) }
         }
 
-        debug_assert!(0 < max_ssaframesize_in_pages);
+        debug_assert_ne!(0, max_ssaframesize_in_pages);
         let xcr0 = xgetbv0();
 
-        // See allorithm of Intel dev manual Chpt 40.7.2.2
+        // See algorithm of Intel x86 dev manual Chpt 40.7.2.2
         let xfrm = (0..64)
             .map(|bit| {
                 let select = 0x1 << bit;
-                match bit {
-                    0 | 1 => select, // Bit 0 and 1 always need to be set
-                    _ => {
-                        if xcr0 & select == 0 {
-                            return 0;
-                        }
 
-                        let (base, size) = {
-                            let cpuid = Self::cpuid(0x0d, bit);
-                            let base = cpuid.map_or(0, |c| c.ebx);
-                            let size = cpuid.map_or(0, |c| c.eax);
-                            (base, size)
-                        };
-                        if max_ssaframesize_in_pages * 0x1000 <= base + size {
-                            return 0;
-                        }
-
-                        select
-                    }
+                if bit == 0 || bit == 1 {
+                    return select; // Bit 0 and 1 always need to be set
                 }
+
+                if xcr0 & select == 0 {
+                    return 0;
+                }
+
+                let (base, size) = {
+                    let cpuid = cpuid(0x0d, bit);
+                    let base = cpuid.map_or(0, |c| c.ebx);
+
+                    if base == 0 {
+                        (0, 0)
+                    } else {
+                        let size = cpuid.map_or(0, |c| c.eax);
+                        (base, size)
+                    }
+                };
+
+                if max_ssaframesize_in_pages * 0x1000 < base + size {
+                    return 0;
+                }
+
+                select
             })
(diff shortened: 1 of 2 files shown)


intel-sgx/enclave-runner/src/loader.rs
```diff
@@ -163,11 +165,67 @@ impl<'a> EnclaveBuilder<'a> {
...
+            })
+            .fold(0, |xfrm, b| xfrm | b);
+
+        // Intel x86 manual Vol 3 Chpt 13.3:
+        // "Executing the XSETBV instruction causes a general-protection fault (#GP) if ECX = 0
+        // and EAX[17] ≠ EAX[18] (TILECFG and TILEDATA must be enabled together). This implies
+        // that the value of XCR0[18:17] is always either 00b or 11b."
+        // The enclave entry code executes xrstor, and we may have just cleared bit 18, so we
+        // need to correct the invariant
+        if (xfrm & (0x1 << 17)) != (xfrm & (0x1 << 18)) {
```
mzohreva, line 219: The comparison logic seems wrong to me. Don't wee need to shift the result of `&` for each one? e.g.
```Rust
    if ((xfrm & (0x1 << 17)) >> 17) != ((xfrm & (0x1 << 18)) >> 18) { ... }
```

jethrogb (changes requested):

intel-sgx/enclave-runner/src/loader.rs
raoulstrackx: Good catch!

intel-sgx/enclave-runner/src/loader.rs
```diff
@@ -163,11 +165,67 @@ impl<'a> EnclaveBuilder<'a> {
...
+                match bit {
+                    0 | 1 => select, // Bit 0 and 1 always need to be set
+                    _ => {
+                        if xcr0 & select == 0 {
+                            return 0;
+                        }
+
+                        let (base, size) = {
+                            let cpuid = Self::cpuid(0x0d, bit);
+                            let base = cpuid.map_or(0, |c| c.ebx);
```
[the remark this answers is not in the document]
raoulstrackx: Good catch!

intel-sgx/enclave-runner/src/loader.rs
```diff
@@ -163,11 +165,67 @@ impl<'a> EnclaveBuilder<'a> {
...
+            xfrm & !(0x3 << 17)
+        } else {
+            xfrm
+        }
+    }
+
+    fn generate_dummy_signature(&mut self) -> Result<Sigstruct, anyhow::Error> {
+        let mut enclave = self.enclave.try_clone().unwrap();
+        let create_info = PageReader::new(&mut enclave)?;
         let mut enclave = self.enclave.try_clone().unwrap();
```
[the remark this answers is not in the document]
raoulstrackx: Yes, but that's not what happening here. The enclave from `self` is cloned twice before it's read. That avoids the situation you describe.

intel-sgx/enclave-runner/src/loader.rs
```diff
@@ -163,11 +165,67 @@ impl<'a> EnclaveBuilder<'a> {
...
+        // "Executing the XSETBV instruction causes a general-protection fault (#GP) if ECX = 0
+        // and EAX[17] ≠ EAX[18] (TILECFG and TILEDATA must be enabled together). This implies
+        // that the value of XCR0[18:17] is always either 00b or 11b."
+        // The enclave entry code executes xrstor, and we may have just cleared bit 18, so we
+        // need to correct the invariant
+        if (xfrm & (0x1 << 17)) != (xfrm & (0x1 << 18)) {
+            xfrm & !(0x3 << 17)
+        } else {
+            xfrm
+        }
```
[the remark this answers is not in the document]
raoulstrackx: Good catch!

intel-sgx/enclave-runner/src/loader.rs
```diff
@@ -163,11 +165,67 @@ impl<'a> EnclaveBuilder<'a> {
...
+                            return 0;
+                        }
+
+                        let (base, size) = {
+                            let cpuid = Self::cpuid(0x0d, bit);
+                            let base = cpuid.map_or(0, |c| c.ebx);
+                            let size = cpuid.map_or(0, |c| c.eax);
+                            (base, size)
+                        };
+                        if max_ssaframesize_in_pages * 0x1000 <= base + size {
```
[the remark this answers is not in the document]
raoulstrackx: I think you're right

intel-sgx/enclave-runner/src/loader.rs
```diff
@@ -163,11 +165,73 @@ impl<'a> EnclaveBuilder<'a> {
...
+                    let cpuid = cpuid(0x0d, bit);
+                    let base = cpuid.map_or(0, |c| c.ebx);
+
+                    if base == 0 {
+                        (0, 0)
+                    } else {
+                        let size = cpuid.map_or(0, |c| c.eax);
+                        (base, size)
+                    }
+                };
```
jethrogb, line 209: How about this
```suggestion
                let CpuidResult { ebx: base, eax: size, .. } = match cpuid(0x0d, 1) {
                    None | Some(CpuidResult { ebx: 0, .. }) => return 0,
                    Some(v) => v,
                };
```

intel-sgx/enclave-runner/src/loader.rs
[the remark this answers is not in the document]
jethrogb: Ah, indeed.

jethrogb (dismissed):

intel-sgx/enclave-runner/src/loader.rs
```diff
@@ -163,11 +165,66 @@ impl<'a> EnclaveBuilder<'a> {
...
+
         fn xgetbv0() -> u64 {
             unsafe { arch::x86_64::_xgetbv(0) }
         }
 
+        debug_assert_ne!(0, max_ssaframesize_in_pages);
+        let xcr0 = xgetbv0();
+
+        // See algorithm of Intel x86 dev manual Chpt 40.7.2.2
+        let xfrm = (0..64)
```
Taowyoo, line 187: HI @raoulstrackx , if you have time, could we have a short call about this part of code.
I am not quite understand how you map these loop to Intel's doc:
>  The following pseudo code illustrates how software can calculate this length using XFRM as the input parameter without modifying XCR0:
offset = 576;
size_last_x = 0;
For x=2 to 63
IF (XFRM[x] != 0) Then
tmp_offset = CPUID.(EAX=0DH, ECX= x):EBX[31:0];
IF (tmp_offset >= offset + size_last_x) Then
offset = tmp_offset;
size_last_x = CPUID.(EAX=0DH, ECX= x):EAX[31:0];
FI;
FI;
EndFor
return (offset + size_last_x); (* compute_xsave_size(XFRM), see “ECREATE—Create an SECS page in the Enclave
Page Cache”*)
Where the non-zero bits in XFRM are a subset of non-zero bit fields in XCR0.

intel-sgx/enclave-runner/src/loader.rs
```diff
@@ -163,11 +165,66 @@ impl<'a> EnclaveBuilder<'a> {
         ret
     }
 
-    fn generate_dummy_signature(&mut self) -> Result<Sigstruct, anyhow::Error> {
+    fn generate_xfrm(max_ssaframesize_in_pages: u32) -> u64 {
+        fn cpuid(eax: u32, ecx: u32) -> Option<CpuidResult> {
+            unsafe {
+                if eax <= x86_64::__get_cpuid_max(0).0 {
```
jethrogb, line 171: This function is invalid and shouldn't be used: https://github.com/rust-lang/stdarch/issues/1718

jethrogb: Nevermind, that applies to the second tuple element.

PR #696, opened 2025-02-17
[RTE-348] Support patched upstream driver

raoulstrackx:
Fortanix uses a slightly modified upstream Linux SGX driver that adds support for partially measured pages. This PR adds support for that driver to EDP.

### DIFF
--- a/intel-sgx/sgxs-loaders/src/isgx/mod.rs
+++ b/intel-sgx/sgxs-loaders/src/isgx/mod.rs
@@ -54,8 +54,10 @@ pub enum DriverFamily {
     ///
     /// # Compatibility notes
     ///
-    /// Currently, no implementations of this API family support
-    /// partially-measured pages or providing a launch token.
+    /// Currently, no upstream implementations of this API family support
+    /// partially-measured pages or providing a launch token. A Fortanix internal
+    /// SGX driver explicitly adds support for partially-measured pages for backwards
+    /// compatibility reasons.
     Augusta,
 }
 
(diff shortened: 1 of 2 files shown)


intel-sgx/sgxs-loaders/src/isgx/mod.rs
```diff
@@ -81,6 +83,8 @@ pub enum Error {
     Create(#[source] SgxIoctlError),
     #[error("Failed to call EADD.")]
     Add(#[source] SgxIoctlError),
+    #[error("Failed to call EXTEND.")]
```
jethrogb, line 86:
```suggestion
    #[error("Failed to call EEXTEND.")]
```

intel-sgx/sgxs-loaders/src/isgx/mod.rs
```diff
@@ -249,6 +248,19 @@ impl EnclaveLoad for InnerDevice {
...
 
+                if partially_measured {
+                    for chunk in 0..16 {
+                        if (0x1 << chunk) & chunks.0 != 0 {
+                            // Only supported with Fortanix' internal upstream driver patch
+                            let mut extend = ioctl::augusta::SgxEnclaveExtend {
+                                offset: eadd.offset as u64 + chunk as u64 * 256,
+                            };
+                            let fd = mapping.device.fd.as_raw_fd();
+                            ioctl_unsafe!(Extend, ioctl::augusta::extend(fd, &mut extend))?;
```
jethrogb, line 259: Check if ENOIOCTLCMD and return the same error we used to return.

jethrogb (changes requested):

Taowyoo:
May need to bump version in this PR or another one later

raoulstrackx:
Good catch! I've updated all versions where needed

raoulstrackx:
So can this PR be merged?

Taowyoo (approved):
LGTM, But since @jethrogb left a change request before. Could you check with him that current PR is ok for him?

jethrogb (approved):

PR #700, opened 2025-02-20
Correct version of crates

Taowyoo:
- Bump `ipc-queue` version
- Correct versions in other crates


ipc-queue/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "ipc-queue"
-version = "0.3.0"
+version = "0.3.1"
```
s-arash, line 3: Since the `fortanix-sgx-abi` dep has been updated from 0.5 to 0.6, shouldn't this be bumped to 0.4?

Taowyoo: Make sense, Wii fix it

Taowyoo, line 3:
```suggestion
version = "0.4.0"
```

s-arash (approved):

Issue #703, opened 2025-02-21
Update CI for dependabot change


Taowyoo:
Because of https://github.com/dependabot/dependabot-core/issues/3253#issuecomment-852541544 
Now dependabot will not have access to repo secrets.

We need to update our CI script to handle this case.

PR #704, opened 2025-02-21
Fix dependabot CI failures

Taowyoo:
## Background

Because of https://github.blog/changelog/2021-02-19-github-actions-workflows-triggered-by-dependabot-prs-will-run-with-read-only-permissions, dependabot cannot access secrets when action event type is `pull_request`.

## Solution

Update event type to `pull_request_target` so that dependabot is able to access secrets.

## Note

It's is safe, because any PR from forked repo will require an approval from maintainer before running CI.

Fix: #703

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,10 +1,7 @@
 name: CI
 
 on:
-  pull_request:
-    paths-ignore:
-      - '*.md'
-      - 'LICENSE'
+  pull_request_target:
     branches:
       - master
   # This CI will be triggered on any merge_group events


mzohreva (dismissed):

jethrogb:
Why is it ok to skip a test when updating dependencies?

Taowyoo:
Hi @jethrogb , I change the approach to fix the problem, Please check again.

Taowyoo:
I found this could be solved by setting https://github.com/fortanix/rust-sgx/settings/secrets/dependabot
Will close this PR once I verified CI is working.

PR #713, opened 2025-02-26
[RTE-381] Pin to `nightly-2025-02-22` compiler for some CI tests

raoulstrackx:
Compiling `async-usercalls` triggers an upstream compiler bug in `nightly-2025-02-23` or subsequent compilers. This PR pins some CI tests to the `nightly-2025-02-22` compiler until the bug is resolved upstream. A minimal PoC still needs to be created and a bug report still needs to be filed.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -51,12 +51,15 @@ jobs:
         rustup toolchain add nightly
         rustup target add x86_64-fortanix-unknown-sgx --toolchain nightly
         rustup update
+        # nightly-2025-02-22 and following are buggy which causes some crates not to compile
+        rustup install nightly-2025-02-22
+        rustup +nightly-2025-02-22 target add x86_64-fortanix-unknown-sgx
 
     - name: Cargo test --all --exclude sgxs-loaders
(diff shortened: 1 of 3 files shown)


s-arash:
Yeah, since the CI is unblocked for now, let's not pin the rust toolchain version in CI scripts.

raoulstrackx:
Oh I missed that PR. In general I think it's better to pin to a working compiler. In the macro that #708 changed, we really meant that it needs to be a type, not a token tree. Once the upstream compiler is fixed, we probably will forget to revert that PR. Also note that if a stable compiler is ever produced with this bug, other CI tests that use it, will start failing too (e.g., the `enclave-runner`). Well, we should get this fixed before it gets that far.

Taowyoo:
Based on my discussion with  @s-arash , this is not a compiler bug but regression.

Note: in our marco code, we mix use :ty and :tt for return types.

Anyway, we need to identify this and find/create issue in rust upstream.

raoulstrackx:
Filed issue [#138975](https://github.com/rust-lang/rust/issues/138975) upstream

Issue #725, opened 2025-03-24
[BUG]: Cargo install fails


javierron:
**Describe the bug:**
`cargo install fortanix-sgx-tools sgxs-tools` fails with errors:

```
error[E0658]: the `!` type is experimental
   --> /home/javier/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/enclave-runner-0.7.0/src/usercalls/abi.rs:311:1
    |
311 | invoke_with_usercalls!(define_usercalls);
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = note: see issue #35121 <https://github.com/rust-lang/rust/issues/35121> for more information
    = help: add `#![feature(never_type)]` to the crate attributes to enable
    = note: this compiler was built on 2025-03-23; consider upgrading it if it is out of date
    = note: this error originates in the macro `invoke_with_usercalls` (in Nightly builds, run with -Z macro-backtrace for more info)

error[E0277]: the trait bound `!: RegisterArgument` is not satisfied
   --> /home/javier/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/enclave-runner-0.7.0/src/usercalls/abi.rs:64:59
    |
64  |                     (handler, ReturnValue::into_registers(ret))
    |                               --------------------------- ^^^ the trait `RegisterArgument` is not implemented for `!`
    |                               |
    |                               required by a bound introduced by this call
...
311 | invoke_with_usercalls!(define_usercalls);
    | ---------------------------------------- in this macro invocation
    |
    = help: the following other types implement trait `RegisterArgument`:
              ()
              *const T
              *mut T
              Option<NonNull<T>>
              bool
              i16
              i32
              i64
            and 7 others
note: required for `std::result::Result<!, usercalls::EnclaveAbort<bool>>` to implement `ReturnValue`
   --> /home/javier/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/enclave-runner-0.7.0/src/usercalls/abi.rs:162:27
    |
162 | impl<T: RegisterArgument> ReturnValue for UsercallResult<T> {
    |         ----------------  ^^^^^^^^^^^     ^^^^^^^^^^^^^^^^^
    |         |
    |         unsatisfied trait bound introduced here
    = note: this error originates in the macro `define_usercalls` which comes from the expansion of the macro `invoke_with_usercalls` (in Nightly builds, run with -Z macro-backtrace for more info)

[...]

PR #726, opened 2025-03-25
Fix compile time error in enclave-runner

raoulstrackx:
Addresses #725


jethrogb:
Type inference failure sounds like a compiler regression

kostko:
Looks like a recent `deranged` crate update broke a lot of stuff everywhere.

kostko:
I believe the problematic deranged version has now been yanked.

tvsfx:
The only yanked version [I see](https://crates.io/crates/deranged/versions) is 4.1, whereas afaict we dependen on 3.x.x in the first place. That's what's confusing me.

kostko:
I believe it was the `time` crate from the `deranged` author that initially didn't want to yank the broken version.

PR #738, opened 2025-04-07
[RTE-428] Fetching QeIdentity with TCB Evaluation Data Number

raoulstrackx:
This PR:

- Adds support to dcap artifact retrieval tool to fetch QE identity data based on TCB Evaluation Data Number
- Changes the API to fetch TCB info to keep it in line with how it stores/retrieves stored artifacts

### DIFF
--- a/intel-sgx/dcap-artifact-retrieval/src/provisioning_client/intel.rs
+++ b/intel-sgx/dcap-artifact-retrieval/src/provisioning_client/intel.rs
@@ -336,22 +336,15 @@ impl TcbInfoApi {
 }
 
 impl<'inp> TcbInfoService<'inp> for TcbInfoApi {
-    fn build_input_from_fmpsc(
+    fn build_input(
         &'inp self,
         fmspc: &'inp Fmspc,
+        tcb_evaluation_data_number: Option<u16>,
     ) -> <Self as ProvisioningServiceApi<'inp>>::Input {
         TcbInfoIn {
             api_version: self.api_version.clone(),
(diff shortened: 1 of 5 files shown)


intel-sgx/dcap-artifact-retrieval/src/provisioning_client/intel.rs
```diff
@@ -450,10 +443,17 @@ impl<'inp> ProvisioningServiceApi<'inp> for QeIdApi {
...
-            INTEL_BASE_URL, api_version,
-        );
+        let url = if let Some(tcb_evaluation_data_number) = input.tcb_evaluation_data_number {
+            format!(
+                "{}/sgx/certification/v{}/qe/identity?tcbEvaluationDataNumber{}",
+                INTEL_BASE_URL, api_version, tcb_evaluation_data_number
+            )
+        } else {
+            format!(
+                "{}/sgx/certification/v{}/qe/identity?update=early",
```
aditijannu, line 453: Is there a use case where we would want to get the standard access instead of early without providing the TCB number?

aditijannu (approved):

intel-sgx/dcap-artifact-retrieval/src/provisioning_client/intel.rs
raoulstrackx: No as soon as the standard/early option was introduced, we always used the "early" one. It already used to be hard coded. Now we want either "early" or a specific TCB evaluation data number

PR #744, opened 2025-04-10
[RTE-436] Bugfix qe identity fetch from pccs

raoulstrackx:


intel-sgx/dcap-artifact-retrieval/src/provisioning_client/pccs.rs
```diff
@@ -772,7 +773,31 @@ mod tests {
     #[ignore = "PCCS service needs an update to support the new endpoint"]
     #[test]
     pub fn tcb_evaluation_data_numbers() {
+        let root_ca = include_bytes!("../../tests/data/root_SGX_CA_der.cert");
+        let root_cas = [&root_ca[..]];
         let client = make_client(PcsVersion::V4);
-        client.tcb_evaluation_data_numbers().unwrap();
```
aditijannu, line 776: Did this unwrap() not report an error before? Can we update the client to ensure that an error is reported when the url is ill formed?

raoulstrackx: No, the bug resulted in a url of `https://pccs.fortanix.com/sgx/certification/v4/qe/identity?tcbEvaluationDataNumber18`. That's a valid url, and the service simply silently ignored the `tcbEvaluationDataNumber18` parameter. I discovered this by coincidence by seeing that the content it returned actually had a `tcbEvaluationDataNumber` of 17 instead of 18.

aditijannu (approved):

PR #747, opened 2025-04-14
Make filename method public for DCAP artifacts

gausk:


intel-sgx/dcap-artifact-retrieval/Cargo.toml
```diff
@@ -28,7 +28,7 @@ mbedtls = { version = "0.12.3", features = [
     "std",
 ], default-features = false }
 num_enum = { version = "0.7", features = ["complex-expressions"] }
-pcs = { version = "0.6.0", path = "../pcs" }
```
aditijannu, line 31: The latest published version of pcs on crates.io is 0.6.0, shouldn't this be 0.6.1?

intel-sgx/pcs/src/pckcrl.rs
```diff
@@ -29,21 +29,21 @@ impl PckCrl {
         Ok(crl)
     }
 
-    pub fn filename() -> String {
-        Self::DEFAULT_FILENAME.to_string()
+    pub fn filename() -> &'static str {
```
Pagten, line 32: This is a backwards-incompatible change. Let's not change the return type.

intel-sgx/pcs/src/tcb_evaluation_data_numbers.rs
```diff
@@ -114,6 +114,10 @@ impl RawTcbEvaluationDataNumbers {
         }
     }
 
+    pub fn filename() -> &'static str {
```
Pagten, line 117: Let's return a `String` so we have more flexibility for future alternative implementations of this method.

intel-sgx/dcap-artifact-retrieval/Cargo.toml
Pagten, line 1: Are the changes to this file needed? Can't this tool keep depending on the older version of the `pcs` crate? It doesn't seem to depend on any of the changes made in `pcs`.

gausk: If we just use `rust-sgx` master for these two crates in our code base then we don't. But if we want to use published version of `pcs` and `dcap-artifact-retrieval` then I need to make this change as otherwise I get version mismatch of pcs struct type in `dcap-artifact-retrieval`

intel-sgx/dcap-artifact-retrieval/Cargo.toml
gausk: The last version in pcs was `0.6.1` so I just updated the patch version by 1.

aditijannu (approved):

PR #770, opened 2025-06-05
Remove `compiler-builtins` from `rustc-dep-of-std` dependencies

tgross35:
Since [1], this will come automatically from `rustc-std-workspace-core` and the crates.io dependency should no longer be specified.

[1]: https://github.com/rust-lang/rust/pull/141993

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -550,9 +550,9 @@ dependencies = [
 
 [[package]]
 name = "compiler_builtins"
-version = "0.1.152"
+version = "0.1.160"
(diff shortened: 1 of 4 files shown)


tgross35:
https://github.com/fortanix/serde/blob/1755f934942d36f9caf6cb73177fda1d25ca20f4/serde/Cargo.toml#L21 should also be able to be deleted, I believe

tgross35:
@jethrogb mind restarting the CI? The failures from the last run were more than likely unrelated, hence the rebase.

tgross35:
Thanks!

PR #775, opened 2025-06-12
[RTE-487] Update fortanixvme tools to pick up nitro_cli 1.4.2

raoulstrackx:
The old ftxvme-elf2eif and eif-sign depend on older versions of nitro_cli. Unfortunately those are incompatible with newer versions of Docker. This PR:
- Updates these tools to pick up nitro_cli 1.4.2 from upstream. They avoid having to rely on a custom branch.
- Keeps the functionality of eif-sign intact, but it has been reduced to a very thin wrapper around nitro_cli.

### DIFF
--- a/intel-sgx/aesm-client/src/lib.rs
+++ b/intel-sgx/aesm-client/src/lib.rs
@@ -11,7 +11,7 @@
 #![doc(html_logo_url = "https://edp.fortanix.com/img/docs/edp-logo.svg",
        html_favicon_url = "https://edp.fortanix.com/favicon.ico",
        html_root_url = "https://edp.fortanix.com/docs/api/")]
-#![deny(warnings)]
+#![allow(warnings)]
 
 extern crate byteorder;
 pub extern crate failure;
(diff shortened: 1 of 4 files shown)


fortanix-vme/eif-tools/src/bin/eif-sign.rs
```diff
@@ -66,27 +63,21 @@ fn main() {
...
+        eif_path: output_path.to_string(),
+        signing_certificate: Some(signing_certificate.to_string()),
+        private_key: Some(private_key.to_string()),
     };
-
-    let mut output_file = File::create(output_path).expect("Could not create output file");
-    let measurements = builder.write_to(&mut output_file);
-
-    println!("EIF successfully signed: `{}`", output_path);
-    println!("{:#?}", measurements);
```
aditijannu, line 91: Do we need these measurements to be printed?

aditijannu:
Do we have any unit tests to check if these changes work provide expected error or result based on specific inputs?

fortanix-vme/eif-tools/src/bin/eif-sign.rs
raoulstrackx: They are still being printed, only by the `nitro_cli` tool  now

PR #777, opened 2025-06-17
Do not let tokio swallow forwarded panics

### LINKED ISSUE #763: [BUG]: enclave-runner does not join on tokio tasks it spawn
Tokio swallows task panics, and it's up to the spawner to check the result of that task. In the case of the enclave runner, if an enclave panics that should propagate all the way up to a full enclave-runner panic (if forward_panics is set). Currently, those panics are ignored and the enclave-runner instead hangs.

s-arash:
We forward enclave panics (i.e., panic if the enclave panics) inside a tokio task, which results in tokio eating up the panic. This is the exact opposite of the intended behavior of `forward_panics`. This PR makes the runner panic outside tokio tasks, so the panic won't be caught by tokio.

Closes #763

### DIFF
--- a/intel-sgx/enclave-runner/src/usercalls/mod.rs
+++ b/intel-sgx/enclave-runner/src/usercalls/mod.rs
@@ -880,10 +880,9 @@ impl EnclaveState {
             .expect("failed to create tokio Runtime");
         let local_set = tokio::task::LocalSet::new();
 
-        let forward_panics = enclave.forward_panics;
         let return_future = async move {
             while let Some((my_result, mode)) = rx_return_channel.recv().await {
-                if forward_panics {
+                if enclave_clone.forward_panics {
                     if let Err(EnclaveAbort::Exit { panic: Some(panic) }) = &my_result {
                         panic!("{}", panic);
                     }
--- a/intel-sgx/enclave-runner/src/usercalls/mod.rs
+++ b/intel-sgx/enclave-runner/src/usercalls/mod.rs
@@ -855,9 +855,6 @@ impl EnclaveState {
                         EnclavePanic::DebugStr("async exit with a panic".to_owned())
                     }
                 };
-                if enclave.forward_panics {
-                    panic!("{}", &panic);
-                }
                 Err(EnclaveAbort::Exit{ panic: Some(panic) })
             }
             Err(EnclaveAbort::Exit { panic: false }) => Err(EnclaveAbort::Exit{ panic: None }),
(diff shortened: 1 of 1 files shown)


Taowyoo (dismissed):

Goirad (dismissed):
Nice!

jethrogb:
Does this work? I'd think all aborts get converted to Secondary in `syscall_loop`.

jethrogb:
How about this instead?
```diff
diff --git a/intel-sgx/enclave-runner/src/usercalls/mod.rs b/intel-sgx/enclave-runner/src/usercalls/mod.rs
index bb7e51a..562b976 100644
--- a/intel-sgx/enclave-runner/src/usercalls/mod.rs
+++ b/intel-sgx/enclave-runner/src/usercalls/mod.rs
@@ -855,9 +855,6 @@ impl EnclaveState {
                         EnclavePanic::DebugStr("async exit with a panic".to_owned())
                     }
                 };
-                if enclave.forward_panics {
-                    panic!("{}", &panic);
-                }
                 Err(EnclaveAbort::Exit{ panic: Some(panic) })
             }
             Err(EnclaveAbort::Exit { panic: false }) => Err(EnclaveAbort::Exit{ panic: None }),
@@ -899,7 +896,13 @@ impl EnclaveState {
 
                     (Err(e @ EnclaveAbort::Exit { panic: Some(_) }), _) |
                     (Err(e @ EnclaveAbort::InvalidUsercall(_)), _) => {
-                        let e = e.map_panic(|opt| opt.unwrap());
+                        let e = e.map_panic(|opt| {
+                            let panic_msg = opt.expect("in Some match arm");
+                            if enclave_clone.forward_panics {
+                                panic!("{}", &panic_msg);
+                            }
+                            panic_msg
+                        });
                         let cmd = enclave_clone.kind.as_command().unwrap();
                         let mut cmddata = cmd.panic_reason.lock().await;
 
```

intel-sgx/enclave-runner/src/usercalls/mod.rs
```diff
@@ -883,8 +880,15 @@ impl EnclaveState {
             .expect("failed to create tokio Runtime");
         let local_set = tokio::task::LocalSet::new();
 
+        let forward_panics = enclave.forward_panics;
         let return_future = async move {
             while let Some((my_result, mode)) = rx_return_channel.recv().await {
+                if forward_panics {
```
jethrogb:
```suggestion
                if enclave_clone.forward_panics {
```

jethrogb (approved):

intel-sgx/enclave-runner/src/usercalls/mod.rs
s-arash: Done!

Taowyoo (approved):

Goirad (approved):

PR #790, opened 2025-08-04
Update GitHub Action for creating GitHub release & publishing crates to crates.io

Taowyoo:
Migrate current publish crate GitHub Action to a new GitHub Action that:
- Create a GitHub release on tag pushed
  - Action will also build and upload binaries if released package has `bin` targets.
- Publish new version of crate to `crates.io` on tag pushed.

This action is tested in my fork:
- https://github.com/Taowyoo/rust-sgx/actions/runs/16789919202 creates https://github.com/Taowyoo/rust-sgx/releases/tag/pcs_v0.0.2
  - This shows the case when crate is type of `lib`.
- https://github.com/Taowyoo/rust-sgx/actions/runs/16789764405 creates https://github.com/Taowyoo/rust-sgx/releases/tag/sgxs-tools_v0.0.8
  - This shows the case when crate is type of `bin` and have multiple binaries.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -65,7 +65,7 @@ jobs:
       - name: Publish crate to crates.io
         run: |
           echo "Publishing crate: $CRATE_NAME"
-          cargo publish --dry-run --locked --package "$CRATE_NAME"
+          cargo publish --locked --package "$CRATE_NAME"
         env:
           CARGO_REGISTRY_TOKEN: ${{ secrets.CARGO_REGISTRY_TOKEN }}
 


.github/workflows/release.yml
```diff
@@ -0,0 +1,67 @@
+name: Create Release and Publish
+
+on:
+  push:
+    tags:
+      - 'dcap-artifact-retrieval_v*'
+      - 'dcap-retrieve-pckid_v*'
```
raoulstrackx: Should we also do this for the other executables to remain consistent?

.github/workflows/release.yml
```diff
@@ -0,0 +1,67 @@
...
+          cargo build --release --package "$PKG"
+          mv "target/release/$PKG" dist/
+          cd dist
+          tar -czf "$PKG-x86_64-unknown-linux-gnu.tar.gz" "$PKG"
+          rm "$PKG"
+
+      - name: Publish GitHub Release using GitHub CLI
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
```
raoulstrackx, line 100: When we publish a release on GitHub it'd make sense to also publish to crates.io

.github/workflows/release.yml
```diff
@@ -0,0 +1,67 @@
...
+            echo "Unknown package for tag: ${GITHUB_REF_NAME}"
+            exit 1
+          fi
+          cargo build --release --package "$PKG"
+          mv "target/release/$PKG" dist/
+          cd dist
+          tar -czf "$PKG-x86_64-unknown-linux-gnu.tar.gz" "$PKG"
+          rm "$PKG"
+
+      - name: Publish GitHub Release using GitHub CLI
```
raoulstrackx, line 97: Shouldn't someone still approve like we do for publishing to crates.io?

.github/workflows/release.yml
Taowyoo: I would suggest this GitHub action is only for github release.
For automate publishing to crates.io, need another PR to update the existing action: https://github.com/fortanix/rust-sgx/blob/master/.github/workflows/publish-crates.yml

.github/workflows/release.yml
Taowyoo: What other executables you want to add?
Also I think adding other executables is out of scope for support offline DCAP attestation.
I could add them in another PR.

.github/workflows/release.yml
Taowyoo: Maybe not necessary? People who have permission to push tag also have permission to create release in this Github repo?

.github/workflows/release.yml
aditijannu: You could rename this workflow to be dcap-binary-release.yml

raoulstrackx: I'm mainly thinking of `sgx-detect` as that's something users do execute repeatedly. But all SGX related executables could be in scope such as `sgxs-info`, ...

.github/workflows/release.yml
raoulstrackx: If we keep publishing to crates.io and publishing a release on GH separate, we may make a mistake and publish one but not the other.
As it's possible to force update a tag, we could even publish both from the same version number but from actual different commit hashes.

.github/workflows/release.yml
raoulstrackx: If we want to keep versions on crates.io and GH releases in sync, we should only let people with publishing rights to crates.io, create releases in GH.

.github/workflows/release.yml
Taowyoo: Hi @raoulstrackx , the GitHub Action here is for crates with executables.
Do you mean you suggest to have Github releases for all crates?

.github/workflows/release.yml
```diff
@@ -0,0 +1,106 @@
...
+
+      - name: Setup Rust toolchain
+        run: rustup update
+
+      - name: Get name of crate to be published
+        run: |
+          # Extract the crate name from the GITHUB_REF_NAME environment variable
+          # GITHUB_REF_NAME contains the Tag name (e.g. rs-lic_v0.2.4) associated with the event
+          export CRATE_NAME=$(python3 -c "print('$GITHUB_REF_NAME'.rsplit('_v', 1)[0])")
+          echo "CRATE_NAME=$CRATE_NAME" >> $GITHUB_ENV
```
raoulstrackx, line 64: Does this still work? How is the `GITHUB_REF_NAME` environment variable set?

Taowyoo: - GITHUB_REF , For tags it is `refs/tags/<tag_name>`. For example, `refs/tags/tag-name-1`.
- GITHUB_REF_NAME, it's `tag-name-1`


Ref: https://docs.github.com/en/actions/reference/workflows-and-actions/variables

Taowyoo: Here I assume the out tag convention is `<crate-name>_v<version number>`.

And this is partially checked by `on tags` triggers with tag name filter at very beginning of this action file.

- For publish , `cargo publish -p` will check if given crate exist in workspace.
- For release, the python script will check if given crate name exist in output of `cargo metadata`, see
https://github.com/fortanix/rust-sgx/blob/2d85757ad9ba2da678526186a945f5507287f2fd/build-release-artifacts.py#L48-L56

.github/workflows/release.yml
```diff
@@ -0,0 +1,106 @@
...
+      - name: Publish crate to crates.io
+        run: |
+          echo "Publishing crate: $CRATE_NAME"
+          cargo publish --locked --package "$CRATE_NAME"
+        env:
+          CARGO_REGISTRY_TOKEN: ${{ secrets.CARGO_REGISTRY_TOKEN }}
+
+  release:
+    name: Create GitHub release
+    environment: "Publish to crates.io"
```
raoulstrackx, line 74: Where is this used? I was expecting something like "Publish GitHub release"

.github/workflows/release.yml
```diff
@@ -0,0 +1,106 @@
...
+          export CRATE_NAME=$(python3 -c "print('$GITHUB_REF_NAME'.rsplit('_v', 1)[0])")
+          echo "CRATE_NAME=$CRATE_NAME" >> $GITHUB_ENV
+      - name: Publish crate to crates.io
+        run: |
+          echo "Publishing crate: $CRATE_NAME"
+          cargo publish --locked --package "$CRATE_NAME"
+        env:
+          CARGO_REGISTRY_TOKEN: ${{ secrets.CARGO_REGISTRY_TOKEN }}
+
+  release:
```
raoulstrackx, line 72: Publishing to crates.io or releasing to GitHub may fail independently of the other. It'd be nice if there are some checks that avoid such cases; either both just succeed, or both should fail. That's never 100% guaranteed as when taking network issues into account as such, but with dry-runs we may come close. But maybe that can be handled in another PR.

.github/workflows/release.yml
Taowyoo: The environment `"Publish to crates.io"` only contains API KEY for crates.io so I did not change the name of it.

For publish GitHub release, there is not external API key needed. I put the environment `"Publish to crates.io"` here just for ensuing two steps both using same approval process.

Taowyoo: > Where is this used?

This is defined at https://github.com/fortanix/rust-sgx/settings/environments
GitHub environment provide user a way to define an environment with specific secrets or env vars. It also provides way to define protection rules such as "require approval".

PR #793, opened 2025-08-08
[RTE-527] Bugfix pckcerts select pck

raoulstrackx:
There's a `PckCerts::select_pck` function that selects the best matching PCK for a given TCB configuration. Unfortunately, it's buggy. When the given TCB configuration matches _exactly_ what's specified in one of these PCK certs, it wasn't selected. This PR:
- corrects the bug
- adds a unit test for this case
- bumps the patch version of the `pcs` crate


Taowyoo (approved):

intel-sgx/pcs/src/pckcrt.rs
```diff
@@ -580,7 +580,7 @@ impl<V: VerificationType> PckCert<V> {
             .sgx_extension()
             .map_err(|_| Error::InvalidPck("Failed to parse SGX extension".into()))?;
         let tcb = sgx_extension.tcb;
-        if sgx_extension.pceid == pceid && tcb.tcb_components < *comps {
+        if sgx_extension.pceid == pceid && tcb.tcb_components <= *comps {
```
jethrogb, line 583:
```suggestion
        if sgx_extension.pceid <= pceid && tcb.tcb_components <= *comps {
```
?

raoulstrackx: There must be an exact match for the PCE ID, right? I think you're confusing PCE ID with the PCE svn (which already is part of the TCB components).

jethrogb: You're exactly right :)

PR #795, opened 2025-08-15
Bump version for tools that need release binaries

Taowyoo:
- This PR bumps version of following crates to prepare binary release:
  - dcap-artifact-retrieval
  - fortanix-sgx-tools
  - sgxs-tools
- This PR also fixes the readme path issue in Cargo.toml

After bump version, we can use this newly added GitHub Action to create release with binary + crates.io new version: https://github.com/fortanix/rust-sgx/actions/workflows/release.yml

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -726,7 +726,7 @@ dependencies = [
 
 [[package]]
 name = "dcap-artifact-retrieval"
-version = "0.4.1"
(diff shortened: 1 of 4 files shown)


Taowyoo:
I will release new version for following crates first:
- fortanix-sgx-tools
- sgxs-tools

Issue #811, opened 2025-09-12
Fix PCCS TcbEvaluationDataNumbers related tests


Taowyoo:
Recently Intel did following changes to TcbEvaluationDataNumber API:
```
Intel® SGX and Intel® TDX Provisioning Certification Service – Adding TCB Evaluation Data Number Historical Data

Dear Intel® SGX and Intel® TDX Customers,

The “Retrieve TCB Evaluation Data Numbers” endpoint (/{sgx/tdx}/certification/v4/tcbevaluationdatanumbers) response has been supplemented with additional historical data back to late 2022 (tcbEvaluationDataNumber = 13), which is the TCB Recovery immediately prior to the introduction of version 4 APIs of the Intel® SGX and Intel® TDX Provisioning Certification Service.

This has been done to facilitate using date-based appraisal policies, as Intel SGX and Intel TDX ecosystem adoption of attestation appraisal capabilities increases. Reference the Attestation Appraisal section of the [Trusted Computing Base Recovery guidance page](https://www.intel.com/content/www/us/en/developer/topic-technology/software-security-guidance/trusted-computing-base-recovery-attestation.html) or the Guidance for Infrastructure Providers section of the [Trusted Computing Base Recovery technical paper](https://www.intel.com/content/www/us/en/developer/articles/technical/software-security-guidance/best-practices/trusted-computing-base-recovery.html) for more details.

Thank You,

Intel SGX & TDX Services Team
```
So this API now will response outdated TcbEvaluationDataNumber.

Since PCCS is cache service and not able to cache the `Gone` response from Intel PCS, We need to change the test behavior to call QE ID API with update=standard to get the smallest TcbEvaluationDataNumber that's still available.
Then filter out those outdated TcbEvaluationDataNumbers.

PR #812, opened 2025-09-12
Update qe id and tcb info APIs then fix pccs tests

Taowyoo:
This PR:
- Updates QE ID and TCB info API implementation to be able to call these API with given `update` type.
- In `pccs.rs` , update tests to not calling QE ID and TCB info APIs with outdated TCB evaluation numbers.
- Fixs #811

### DIFF
--- a/intel-sgx/dcap-artifact-retrieval/src/cli.rs
+++ b/intel-sgx/dcap-artifact-retrieval/src/cli.rs
@@ -32,7 +32,7 @@ enum Origin {
     Pccs,
 }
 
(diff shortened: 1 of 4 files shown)


raoulstrackx:
This seems complex. Can't we just match the error on `Error::PCSError(StatusCode::Gone)` and ignore it in the tests when it's gone? Creating a function `resource_gone` on `Error` would be nice :)

raoulstrackx:
Yes PCCS should definitely get an update to correctly report "Gone" when it needs to

PR #813, opened 2025-09-15
[RTE-557] Ensure Azure is tested in V3 and V4

raoulstrackx:
The Azure dcap artifact retrieval tool wasn't being tested for the V4 api of the PCS service. This PR corrects that

Blocked by #811


Taowyoo:
Since the solution for #811 is still in discussion and may be solved on PCCS side.
Now we are using  f09a262b41859b8201ebc2ed44856ce44e53f00f as temporary fix.

PR #814, opened 2025-09-15
[RTE-558] Fix ToPrimitive and FromPrimitive warnings

raoulstrackx:
Fix `ToPrimitive` and `FromPrimitive` warnings

Blocked by #811


Taowyoo:
Since the solution for #811 is still in discussion and may be solved on PCCS side.
Now we are using  f09a262b41859b8201ebc2ed44856ce44e53f00f as temporary fix.

PR #825, opened 2025-10-24
[RTE-597] Bugfix time runs too fast when system is under heavy load

raoulstrackx:
Newer processors do allow the use of the `rdtsc` instruction inside of an SGX enclave. In contrast to SGXv1 this means we no longer require an explicit `insecure_time` usercall to get the time inside of the enclave. The frequency that this TSC clock runs at, however is not easy to derive. On some processors we are able to get the frequency directly from the hardware. On others we can get an approximations. None of these however gives us the correct frequency for very long time periods. We fix this by calling out to the enclave runner, and (i) get the correct time, and (ii) compute the exact frequency based on this.
Unfortunately there was a bug in the implementation. Even when we knew the approximate frequency, we still computed the frequency once a `frequency_learning_period`. This is problematic when this learning period is rather short, and the time it takes to receive the result is has large variations. This can happen when the system is under very heavy load.
This PR:
* Improves up the CI tests
* Ensures we use the approximate frequency during the `frequency_learning_period`
* Ensures time cannot run backwards when a monotonic time is configured
* Ensures frequency is rather overestimated than underestimated. This results in time running too slow, rather than too fast, which can easily be fixed in contrast to time running too fast for monotonic clocks.
* A test is added that mimics a system under very heavy load.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1782,7 +1782,7 @@ dependencies = [
 
 [[package]]
 name = "insecure-time"
-version = "0.1.0"
+version = "0.2.0"
 dependencies = [
  "clap 4.5.20",
  "rand 0.8.5",
--- a/intel-sgx/enclave-runner/Cargo.toml
+++ b/intel-sgx/enclave-runner/Cargo.toml
@@ -23,7 +23,7 @@ exclude = ["fake-vdso/.gitignore", "fake-vdso/Makefile", "fake-vdso/main.S"]
 sgxs = { version = "0.8.0", path = "../sgxs" }
 fortanix-sgx-abi = { version = "0.6.1", path = "../fortanix-sgx-abi" }
 sgx-isa = { version = "0.4.0", path = "../sgx-isa" }
-insecure-time = { version = "0.1", path = "../insecure-time", features = ["estimate_crystal_clock_freq"] }
+insecure-time = { version = "0.2", path = "../insecure-time", features = ["estimate_crystal_clock_freq"] }
 ipc-queue = { version = "0.4.0", path = "../../ipc-queue" }
 
 # External dependencies
--- a/intel-sgx/insecure-time/Cargo.toml
+++ b/intel-sgx/insecure-time/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "insecure-time"
-version = "0.1.0"
+version = "0.2.0"
 authors = ["Fortanix, Inc."]
 license = "MPL-2.0"
 edition = "2021"


intel-sgx/insecure-time/src/lib.rs
```diff
@@ -523,12 +523,27 @@ impl<T: NativeTime> TscBuilder<T> for LearningFreqTscBuilder<T> {
...
         let LearningFreqTscBuilder { frequency, max_sync_interval, max_acceptable_drift, frequency_learning_period, time_mode, } = self;
-        let tsc_mode = TscMode::Learn {
-            max_sync_interval,
-            max_acceptable_drift,
-            frequency_learning_period,
-            next_sync: Ticks::new(0),
-            frequency: frequency.unwrap_or(Freq(AtomicU64::from(0))),
+        // When the initial frequency is set in the builder, we assume we don't need to resync for
+        // `max_sync_interval` duration
+        let tsc_mode = if let Some((Ok(next_sync), frequency)) = frequency.and_then(|f| Some((Ticks::from_duration(max_sync_interval, &f), f))) {
```
Pagten: Better to use `map(|f| [...])` instead of `and_then(|f| Some([...]))

intel-sgx/insecure-time/src/lib.rs
```diff
@@ -523,12 +523,27 @@ impl<T: NativeTime> TscBuilder<T> for LearningFreqTscBuilder<T> {
...
+                frequency,
+            }
+        } else {
+            TscMode::Learn {
+                max_sync_interval,
+                max_acceptable_drift,
+                frequency_learning_period,
+                next_sync: Ticks::new(0),
+                frequency: Freq(AtomicU64::from(0)),
+            }
```
Pagten, line 546: Nit: the common parts of both branches could be factored out:
```
let (next_sync, frequency, frequency_learning_period) =
    if let Some(next_sync) = frequency.as_ref().and_then(|f| Ticks::from_duration(max_sync_interval, f).ok()) {
        // We won't resync until `max_sync_interval`, use the full period to learn the exact
        // frequency
        (next_sync, frequency, frequency_learning_period.max(max_sync_interval))
    } else {
        (Ticks::new(0), Freq(AtomicU64::from(0)), frequency_learning_period)
    };
let tsc_mode = TscMode::Learn {
    max_sync_interval,
    max_acceptable_drift,
    frequency_learning_period,
    next_sync,
    frequency,
}
```

intel-sgx/insecure-time/src/lib.rs
```diff
@@ -563,6 +578,10 @@ impl<T: NativeTime> FixedFreqTscBuilder<T> {
     }
 }
 
+enum ResyncResult<T> {
```
Pagten, line 581: This seems more like an error rather than a result?

intel-sgx/insecure-time/src/lib.rs
```diff
@@ -27,7 +27,7 @@ use core::ops::Add;
 
 const NANOS_PER_SEC: u64 = 1_000_000_000;
 
-pub trait NativeTime: PartialOrd + Copy + Add<core::time::Duration, Output = Self> + ToOwned {
+pub trait NativeTime: Debug + PartialOrd + Copy + Add<core::time::Duration, Output = Self> + ToOwned {
```
Pagten, line 30: This is a major change according to SemVer.

raoulstrackx: Oh yes good catch!

intel-sgx/insecure-time/src/lib.rs
Taowyoo: This looks like a temporary type used below as value container. So should be fine?

Taowyoo (approved):

PR #833, opened 2025-10-29
[RTE-606] Merge fortanixvme branch

sardok:
This PR intends to merge `fortanixvme-2023-05-07-alpha` into `master`.

### DIFF
--- a/intel-sgx/pcs/src/tcb_evaluation_data_numbers.rs
+++ b/intel-sgx/pcs/src/tcb_evaluation_data_numbers.rs
@@ -339,20 +339,11 @@ impl TcbPolicy {
     }
 
     fn minimum_tcb_evaluation_data_number_ex<V: VerificationType>(&self, tcb_eval: &TcbEvaluationDataNumbers<V>, now: &DateTime<Utc>) -> Option<TcbEvalNumber> {
-        tcb_eval.numbers().fold(None, |last, number| {
-            match (last, self.needs_to_be_enforced(number, &now)) {
-                (last, false) => last,
-                (Some(last), true) => {
-                    if last.number() <= number.number() {
-                        // We need the highest TCB Eval Data Number that needs to be enforced
-                        Some(number.clone())
-                    } else {
-                        Some(last)
-                    }
-                },
-                (None, true)  => Some(number.clone()),
-            }
-        })
+        // We need the highest TCB Eval Data Number that needs to be enforced
+        tcb_eval.numbers()
+            .filter(|number| self.needs_to_be_enforced(number, now))
+            .max_by_key(|number| number.number)
+            .cloned()
     }
 
     pub fn minimum_tcb_evaluation_data_number<V: VerificationType>(&self, tcb_eval: &TcbEvaluationDataNumbers<V>) -> Option<TcbEvalNumber> {


.github/workflows/build.yml
```diff
@@ -32,6 +32,8 @@ jobs:
 
     steps:
     - uses: actions/checkout@v5
+      with:
+        lfs: 'true'
```
raoulstrackx: This shouldn't be needed anymore

fortanix-vme/fortanix-vme-eif/src/enclave.eif
raoulstrackx, line 1: Is this file required?

fortanix-vme/fortanix-vme-eif/tests/mod.rs
```diff
@@ -0,0 +1,41 @@
...
+void main() {
+       int cnt = 0;
+       while(1) {
+               printf("[%3i] Hello world!\n", cnt);
+               sleep(1);
+               cnt++;
+       }
+}' > main.c
+gcc -o a.out main.c -static -static-libgcc -flto
+*/
```
raoulstrackx: Can you create a build script for this instead of linking in the binary directly?

fortanix-vme/fortanix-vme-runner/src/lib.rs
```diff
@@ -172,13 +175,13 @@ impl Connection {
                     //  - reflect this change on the other connection
                     //  - avoid reading from the socket again
                     // https://doc.rust-lang.org/std/io/trait.Read.html#tymethod.read
-                    if Self::transfer_data(remote, &self.remote_name, enclave, "enclave")? == 0 {
+                    if Self::transfer_data(remote, &self.remote_name, enclave, "enclave0")? == 0 {
```
raoulstrackx: I think this is for debugging, can you remove it?

fortanix-vme/fortanix-vme-runner/src/lib.rs
```diff
@@ -172,13 +175,13 @@ impl Connection {
...
                         enclave.shutdown(Shutdown::Write)?;
                         golden_set.remove(remote.as_raw_fd());
                     }
                 }
                 if read_set.contains(enclave.as_raw_fd()) {
-                    if Self::transfer_data(enclave, "enclave", remote, &self.remote_name)? == 0 {
+                    if Self::transfer_data(enclave, "enclave1", remote, &self.remote_name)? == 0 {
```
raoulstrackx: I think this is for debugging, can you remove it?

intel-sgx/ias/Cargo.toml
```diff
@@ -18,7 +18,7 @@ serde_bytes = "0.11"
 serde-bytes-repr = { version = "0.1", optional = true }
 serde_json = { version = "1", optional = true }
 serde = { version = "1.0.7", features = ["derive"] }
-url = "2.2"
+url = "2.4"
```
raoulstrackx: Can you revert this?

fortanix-vme/aws-nitro-enclaves/blobs-tool/src/utils.rs
```diff
@@ -0,0 +1,17 @@
+use anyhow::{self, Context};
+use std::os::unix::fs;
+use std::path::{Path, PathBuf};
+
+pub fn create_symlink(src: &String, dst_dir: &mut PathBuf) -> anyhow::Result<()> {
+    let path = Path::new(src);
+    let filename = path
+        .file_name()
+        .with_context(|| format!("unable to get filename from {}", src))?;
+    dst_dir.push(filename);
```
raoulstrackx: Can you create a new variable `dst` for this? The remainder of the code is confusing now

fortanix-vme/aws-nitro-enclaves/blobs-tool/src/utils.rs
```diff
@@ -0,0 +1,17 @@
...
+    if dst_dir.exists() {
+        std::fs::remove_file(&mut *dst_dir)?;
```
raoulstrackx: It seems dangerous to just delete the file, and not needed? If it's the wrong destination, won't the build script detect this at the moment it computes the hash?

fortanix-vme/fortanix-vme-eif/build.rs
```diff
@@ -0,0 +1,34 @@
...
+    println!("cargo:rerun-if-env-changes=CARGO_MANIFEST_DIR");
+
+    let mut test_dir = PathBuf::from(format!(
+        "{}/tests/data",
+        env::var("CARGO_MANIFEST_DIR").expect("cargo manifest dir"),
+    ));
+    let output_dir = Path::new(BLOB_DEFAULT_DIR);
+    for blob in download_blobs(output_dir).unwrap() {
+        println!("cargo:rerun-if-changed={}", blob);
+        create_symlink(&blob, &mut test_dir).unwrap();
```
raoulstrackx: This seems unnecessarily complex. Can't we move this logic to download the aws blobs to `blobs-tools` (maybe rename it to `aws-nitro-blobs`), expose the blobs using a ` pub const KERNEL: &[u8] = include_bytes!(concat!(env!("OUT_DIR"), "/bzImage"));` and create a dependency on that crate? Then only one build script on the `aws-nitro-blobs` crate is required

fortanix-vme/fortanix-vme-eif/build.rs
```diff
@@ -0,0 +1,34 @@
...
+    println!("cargo:rerun-if-changed=tests/hello_world.c");
+    let compiler = cc::Build::new().get_compiler();
+    let mut cmd = compiler.to_command();
+    cmd.arg("tests/hello_world.c")
+        .args(["-static", "-static-libgcc", "-flto"])
+        .args(["-o", "tests/data/hello_world"]);
+    let status = cmd.status().expect("Failed to execute C compiler");
+    if !status.success() {
+        panic!("Compilation failed, command: {:?}", cmd);
+    }
```
raoulstrackx: The hello_world executable is only used in tests. There shouldn't be a need to create a build script for it here. You can move this code to the tests themselves.

build_eif.sh
```diff
@@ -0,0 +1,91 @@
...
+	--output ${output}
+popd
+
+echo "eif file written to ${output}"
+# To debug one can use:
+#   qemu-system-x86_64 -kernel ./bzImage -initrd ./initramfs -nographic -append "console=ttyS0"
+# however keep in mind that the AWS init process tries to connect to VSOCK address 3 port 9000
+# (see https://github.com/aws/aws-nitro-enclaves-sdk-bootstrap/blob/746ec5d2713e539b94e651601b5c24ec1247c955/init/init.c#L370)
+
+#nitro-cli run-enclave --cpu-count 2 --eif-path ${output}  --memory 512 --debug-mode
```
raoulstrackx: This script isn't being tested, and maybe it shouldn't be checked in. Maybe we can move it the a confluence page? With some context of what we re-implemented in Rust.

fortanix-vme/fortanix-vme-eif/src/enclave.eif
sardok: Removed.

.github/workflows/build.yml
sardok: Removed.

fortanix-vme/aws-nitro-enclaves/blobs-tool/src/utils.rs
sardok: Removed this file. Its is no longer needed with the new `aws-nito-blobs` crate.

fortanix-vme/fortanix-vme-eif/build.rs
sardok: Added `aws-nitro-blobs` crate and kept `blobs-tools` (renamed it `aws-nitro-blobs-downloader`). I needed a separate crate for downloading as downloaded files are consumed in `aws-nitro-blobs/lib.rs`. Alternatively, i can move all functions from `aws-nitro-blobs-downloader` to `aws-nitro-blobs`'s `build.rs` but preferred to keep it as separate crate.

fortanix-vme/fortanix-vme-runner/src/lib.rs
sardok: Updated `enclave0 -> enclave`.

fortanix-vme/fortanix-vme-runner/src/lib.rs
sardok: Updated `enclave1 -> enclave`.

fortanix-vme/aws-nitro-enclaves/blobs-tool/src/utils.rs
sardok: With new structure this file is no longer needed. Removed the file.

fortanix-vme/fortanix-vme-eif/build.rs
sardok: I restructured the tests. Here are some info about them;
* Now there is `src/test_support/mod.rs` that has a helper function to build this file. Since `hello_world` build no longer takes place in a build script, `cc` requires some explicit parameters such as target, host & opt-level (if `cc` is called within a build script, then these parameters are fetched from env variables implicitly). 
* The `hello_world` binary is place under `src/test_support` when compiled (because there is `OUT_DIR` env variable to read from). Consequently, this filepath is added to the `.gitignore` file.
* Using function & variables from `test_support` required a feature flag in order to be accessible from integration tests. So the test invocation for this crate would look like `cargo test --feature=...`. Instead of going this direction, i moved the test in integration tests ( `eif_creation_and_extraction`) to `src/lib.rs` and made it a unit test.

Alternative to this restructuring, would be introducing a sub-crate with a build script (can be named as `test_setup`) where this crate can be added as a dependency to `dev-dependencies` to the project (so would only be compiled during tests). The structure would look like;
```
.
├── Cargo.toml
├── test_setup/
│   ├── Cargo.toml
│   ├── build.rs     <-- Put the hello_world compilation here
│   └── src/
│       └── lib.rs   <-- Empty
└── src/
    └── lib.rs
└── tests/
    └── mod.rs
```
With this structure, the build script can use `OUT_DIR`, cc would have access to necessary env variables (`OPT_LEVEL`, `TARGET`, `HOST`) and there would be minimal change in tests.

build_eif.sh
sardok: Added a confluence page (search for Building EIF Binary)  and removed the file. I added the script as is to the confluence page. To improve the document and add more context, IMO better to add a ticket for that.

intel-sgx/ias/Cargo.toml
sardok: Done.

fortanix-vme/aws-nitro-enclaves/aws-nitro-blobs-downloader/src/lib.rs
```diff
@@ -0,0 +1,116 @@
...
+pub fn download_blobs(output_dir: &Path) -> anyhow::Result<Vec<String>> {
+    if !output_dir.is_dir() {
+        anyhow::bail!("{:?} must be a directory.", output_dir);
+    }
+
+    let client = reqwest::blocking::Client::new();
+    let mut blobs = Vec::new();
+    for blob in &BLOBS {
+        let blob_file = blob.download(&client, output_dir)?;
+        blobs.push(blob_file.to_string_lossy().to_string());
```
raoulstrackx: Can you return a `Vec` of `PathBuf` instead? In most cases if you want to do anything with the path, you need to turn it into a `Path` again, and turning it into a string is easy anyway.

fortanix-vme/fortanix-vme-eif/src/test_support/mod.rs
```diff
@@ -0,0 +1,38 @@
...
+fn read_test_binary() -> Vec<u8> {
+    let binary_path = Path::new(concat!(
+        env!("CARGO_MANIFEST_DIR"),
+        "/src/test_support/hello_world"
+    ));
+    if !binary_path.exists() {
+        let compiler = cc::Build::new()
+            .opt_level(0)
+            .target("x86_64-unknown-linux-gnu")
+            .host("x86_64-unknown-linux-gnu")
```
raoulstrackx: Why do you need to specify the host target?

fortanix-vme/fortanix-vme-eif/src/test_support/mod.rs
raoulstrackx, line 1: Separating the hello_world compilation is nice!
But why is it under `src`? It's only used for testing, and doesn't add new test functionality to the crate itself. Can't it just be moved as a function within the unit test's module?

fortanix-vme/fortanix-vme-runner/Cargo.toml
```diff
@@ -1,15 +1,19 @@
 [package]
-name = "fortanix-vme-runner"
+name = "ftxvme-runner"
```
raoulstrackx: This is on me, but having a different name for the package than the directory name is odd.
I think this was changed to bring it in line with other places where we use "ftx" as well. But then the FTX bitcoin exchange went down and we weren't very happy with FTX being the abbreviation for Fortanix. We should have a discussion on this. At least things should be consistent.

fortanix-vme/fortanix-vme-runner/src/main.rs
```diff
@@ -1,13 +1,144 @@
...
+            enclave_cid: None,
+            memory_mib,
+            cpu_ids: None,
+            debug_mode: Some(false),
+            cpu_count: Some(cpu_count),
+            enclave_name: None,
+            //TODO: With `fortanixvme-2023-05-07` branch merge the nitro-cli
+            //package is downgraded due to dependency conflicts.
+            //The following should be enabled with/after nitro-cli v1.2.2
+            //attach_console: true,
```
raoulstrackx: I think we need to be on the latest nitro-cli tools to ensure we are compatible with the latest Docker version, which is needed for security reasons.

fortanix-vme/fortanix-vme-eif/src/test_support/mod.rs
sardok: Because when `cc::Build::new()` is called in a build script (i.e `build.rs`), these parameters (host, target & opt_level) are derived from env variables where cargo passes specifically for build scripts (you can find them [here](https://doc.rust-lang.org/cargo/reference/environment-variables.html#environment-variables-cargo-sets-for-build-scripts)). But these env variables are missing when cargo is compiling a crate.  Thats why i think it is more convenient to build `hello_world` in a build script. Take a look at my comment here: https://github.com/fortanix/rust-sgx/pull/833#discussion_r2511042612.

fortanix-vme/fortanix-vme-runner/Cargo.toml
sardok: Ok then, there are a few more places to fix this;
* `ftxvme-runner` -> `fortanix-vme-runner`
* `ftxvme-elf2eif` -> `fortanix-vme-elf2eif`
* `ftxvme-runner-cargo` -> `fortanix-vme-runner-cargo`
* `FtxEif` -> `FortanixEif`

fortanix-vme/fortanix-vme-eif/src/test_support/mod.rs
sardok: Its in `src` because i moved the integration test to `src/lib.rs` to avoid duplicating `hello_world` compilation in unit tests & integration tests. I put some details here https://github.com/fortanix/rust-sgx/pull/833#discussion_r2511042612.

fortanix-vme/fortanix-vme-runner/src/main.rs
sardok: Use `nitro-cli v1.4.2` if possible.

fortanix-vme/fortanix-vme-eif/src/test_support/mod.rs
sardok: Unfortunately `host` triple is expected by the `cc` crate. I tried to get host triple in a better way, but couldn't find anything better than the following options;
* Forward it from build script via a custom env variable, via something like `println!("cargo:rustc-env=MY_HOST_TRIPLE={}", host_triple);`
* execute `rustc -vV` and capture from it.

fortanix-vme/aws-nitro-enclaves/aws-nitro-blobs-downloader/src/lib.rs
sardok: Done.

fortanix-vme/fortanix-vme-runner/src/main.rs
sardok: Upgraded.

fortanix-vme/fortanix-vme-eif/src/test_support/mod.rs
sardok: To tackle the issue mentioned in my previous comment, i added `test_resources` crate which builds the `hello_world` binary in its build script. Added this crate to the project's `dev-dependencies`. So the `hello_world` test binary is only compiled for tests.

fortanix-vme/fortanix-vme-runner/tree.txt
raoulstrackx, line 1: This shouldn't have been checked in

fortanix-vme/aws-nitro-enclaves/aws-nitro-blobs-downloader/src/lib.rs
```diff
@@ -0,0 +1,116 @@
...
+const INIT_X86_64_SHA256: &str = "755e650b732777b798cb9ec243ee402bef4826f789cf01a1e453bb724207c005";
+const NSM_X86_64_URL: &str = "https://raw.githubusercontent.com/aws/aws-nitro-enclaves-cli/ec130adc1fd86f2489f482d3f4a02676d3a748a7/blobs/x86_64/nsm.ko";
+const NSM_X86_64_SHA256: &str = "48904e520db2541ca4378da29d85791749408febc81987ade56cc5c556bd90df";
+
+struct Blob {
+    url: &'static str,
+    sha256: &'static str,
+}
+
+const BLOBS: [Blob; 5] = [
```
raoulstrackx: Nitpick: It'd make sense if you moved this `BLOBS` definition (or only its components) to `aws-nitro-blobs` and rename this crate into `verified-downloader` or something better. That way it can be easily re-used.

raoulstrackx (dismissed):

fortanix-vme/fortanix-vme-runner/tree.txt
sardok: Removed.

fortanix-vme/aws-nitro-enclaves/aws-nitro-blobs-downloader/src/lib.rs
sardok: I removed `aws-nitro-blobs-downloader` completely.

fortanix-vme/aws-nitro-enclaves/aws-nitro-blobs/build.rs
```diff
@@ -1,8 +1,68 @@
...
 fn main() {
-    let out_dir = env::var("OUT_DIR").expect("OUT_DIR is missing");
-    download_blobs(Path::new(&out_dir)).expect("Download nitro blobs failed");
+    let mut out_dir = PathBuf::from(env::var("OUT_DIR").expect("OUT_DIR is missing"));
+    let downloader = Downloader::builder()
+        .storage_dir(out_dir.clone())
+        .build()
+        .expect("Unable to initialize downloader");
+
+    let mut lines = Vec::with_capacity(NITRO_BLOBS.len());
```
raoulstrackx: Nit: the capacity is certainly incorrect it `NITRO_BLOBS.len()` will just return 5, but you need way more space than that. I'd just use the default capacity. (You could also just write things to file directly, without using this `Vec`)

raoulstrackx (approved):

fortanix-vme/aws-nitro-enclaves/aws-nitro-blobs/build.rs
sardok: I simplified this file. it doesn't collect all lines in advance now. It just writes a line as a blob becomes ready.

PR #834, opened 2025-11-01
[RTE-610] Fix `insecure_time()` returning a pointer to the stack

s-arash:
Instead of passing as pointer to the global static `InsecureTimeInfo` struct, the code first created a copy of that struct and returned a pointer to that copy. This PR fixes that.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -962,7 +962,7 @@ dependencies = [
 
 [[package]]
 name = "enclave-runner"
-version = "0.7.2"
+version = "0.7.3"
 dependencies = [
  "anyhow",
  "crossbeam",
--- a/intel-sgx/enclave-runner/Cargo.toml
+++ b/intel-sgx/enclave-runner/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "enclave-runner"
-version = "0.7.2"
+version = "0.7.3"
 authors = ["Fortanix, Inc."]
 license = "MPL-2.0"
 edition = "2018"


intel-sgx/enclave-runner/src/usercalls/mod.rs
```diff
@@ -1658,22 +1658,28 @@ impl<'tcs> IOHandlerInput<'tcs> {
...
+    #[inline(always)]
+    fn insecure_time(&mut self) -> (u64, *const InsecureTimeInfo) {
+        let (ts, info) = self.insecure_time_ref();
+        let info = match info {
+            Some(info) => info,
+            None => ptr::null(),
+        };
+        (ts, info)
+    }
+
```
raoulstrackx: Wouldn't this be cleaner:
```
diff --git a/intel-sgx/enclave-runner/src/usercalls/mod.rs b/intel-sgx/enclave-runner/src/usercalls/mod.rs
index 3a7e6995..2caa9d75 100644
--- a/intel-sgx/enclave-runner/src/usercalls/mod.rs
+++ b/intel-sgx/enclave-runner/src/usercalls/mod.rs
@@ -54,7 +54,7 @@ lazy_static! {
 
 const NANOS_PER_SEC: u64 = 1_000_000_000;
 
-static mut TIME_INFO: LazyLock<Option<InsecureTimeInfo>> = LazyLock::new(|| {
+static TIME_INFO: LazyLock<Option<InsecureTimeInfo>> = LazyLock::new(|| {
     if Rdtscp::is_supported() {
         if let Ok(frequency) = Freq::get() {
             return Some(InsecureTimeInfo {
@@ -1663,14 +1663,11 @@ impl<'tcs> IOHandlerInput<'tcs> {
             .duration_since(time::UNIX_EPOCH)
             .unwrap();
         let t = (time.subsec_nanos() as u64) + time.as_secs() * NANOS_PER_SEC;
-        let info = unsafe { match (*TIME_INFO, self.enclave.force_time_usercalls) {
-            (Some(info), false) => {
-                &info
-            },
-            _ => {
+        let info = if let (Some(info), false) = ((&*TIME_INFO).as_ref(), self.enclave.force_time_usercalls) {
+                info
+            } else {
                 ptr::null()
-            }
-        } };
+            };
         (t, info)
     }
 ```

raoulstrackx (commented):
Great catch!

intel-sgx/enclave-runner/src/usercalls/mod.rs
s-arash: I want to make it very explicit that we are returning a reference to a static. Hence my diff. With this diff the compiler isn't guaranteeing that the pointed-to object is alive, and it's very easy to introduce the bug again.

s-arash: In hindsight, it might have been better to define `insecure_time` itself as having the signature `fn insecure_time(&mut self) -> (u64, Option<&'static InsecureTimeInfo>)`, taking advantage of the fact that `Option<&T>` and `*const T` have identical layouts.

PR #837, opened 2025-11-12
[Feature] Allow passing enclave signature file path to `ftxsgx-runner`

Pagten:
This is a small feature addition to `ftxsgx-runner` to allow users to pass a path to the enclave signature file, instead of requiring the signature file to be in the same directory as the enclave. This is convenient for Bazel builds of enclaves.

Note: I considered using the existing `--signature` argument for this, but it would make the argument ambiguous if the signature file happens to be named "dummy" or "coresident". Probably not a concern in practice, but still...

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1626,7 +1626,7 @@ dependencies = [
 
 [[package]]
 name = "fortanix-sgx-tools"
-version = "0.6.1"
+version = "0.6.2"
 dependencies = [
  "aesm-client",
  "anyhow",
--- a/intel-sgx/fortanix-sgx-tools/Cargo.toml
+++ b/intel-sgx/fortanix-sgx-tools/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "fortanix-sgx-tools"
-version = "0.6.1"
+version = "0.6.2"
 authors = ["Fortanix, Inc."]
 license = "MPL-2.0"
 description = """


intel-sgx/fortanix-sgx-tools/src/bin/ftxsgx-runner.rs
```diff
@@ -78,10 +84,13 @@ fn main() -> Result<(), anyhow::Error> {
...
-        Some(Signature::coresident) => { enclave_builder.coresident_signature().context("While loading coresident signature")?; }
-        Some(Signature::dummy) => { enclave_builder.dummy_signature(); },
-        None => (),
+    let signature = args.value_of("signature").map(|v| v.parse().expect("validated"));
+    let signature_file = args.value_of("signature-file");
+    match (signature, signature_file) {
+        (Some(Signature::coresident), _) => { enclave_builder.coresident_signature().context("While loading coresident signature")?; }
+        (Some(Signature::dummy), _) => { enclave_builder.dummy_signature(); }
+        (None, Some(path)) => { enclave_builder.signature(path).with_context(|| format!("Failed to load signature file '{}'", path))?; }
+        _ => (),
```
Taowyoo:
```suggestion
        _ => unreachable!(),
```

Taowyoo (approved):

intel-sgx/fortanix-sgx-tools/src/bin/ftxsgx-runner.rs
Pagten: No, this is not unreachable. We reach this code path when neither the `--signature` nor the `--signature-file` argument is supplied.

I could refactor the match statement to
```
    match (signature, signature_file) {
        (Some(Signature::coresident), None) => { enclave_builder.coresident_signature().context("While loading coresident signature")?; }
        (Some(Signature::dummy), None) => { enclave_builder.dummy_signature(); }
        (None, Some(path)) => { enclave_builder.signature(path).with_context(|| format!("Failed to load signature file '{}'", path))?; }
        (None, None) => (),
        _ => unreachable!(),
    }
```
if you think that's better.

Taowyoo: This looks better!

PR #838, opened 2025-11-12
[RTE-608] Remove `mopa` crate from the cargo

gilanghamidy:
`mopa` is used only in `sgx-detect` tool, as a legacy workaround on generic programming with polymorphic types. Newer Rust now supports type upcasting, allowing `std::any::Any` to be used instead of using `mopa::Any`.

This should fix the Dependabot alert related to `mopa` crate: 
- [https://github.com/fortanix/rust-sgx/security/dependabot/5](https://github.com/fortanix/rust-sgx/security/dependabot/5)
- [https://github.com/fortanix/rust-sgx/security/dependabot/40](https://github.com/fortanix/rust-sgx/security/dependabot/40)

Credit fo solution: @tvsfx

Verified the tools are working on the SGX machine.

### DIFF
--- a/intel-sgx/sgxs-tools/Cargo.toml
+++ b/intel-sgx/sgxs-tools/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "sgxs-tools"
-version = "0.9.2"
+version = "0.9.3"
 authors = ["Fortanix, Inc."]
 license = "MPL-2.0"
 description = """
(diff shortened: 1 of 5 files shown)


Taowyoo (dismissed):

intel-sgx/sgxs-tools/Cargo.toml
Taowyoo: Nit:
You need to bump version if need to publish new version.
(This could be done in separate PR.)

gilanghamidy: I just update it now so to not make a separate PR. Thanks for noting this. :)

raoulstrackx (approved):

PR #855, opened 2026-01-01
[RTE-659] Adding TDX Collateral collection in the `dcap-artifact-retrieval` and  `pcs` crate

gilanghamidy:
Adding the support for TDX collateral data and extending the existing `TcbInfo` types to be able to distinguish over SGX or TDX collateral types.

The PCCS functionality has no unit test yet awaiting the confirmation of the update on Fortanix PCCS server.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1215,7 +1215,7 @@ dependencies = [
 
 [[package]]
 name = "dcap-artifact-retrieval"
-version = "0.4.2"
+version = "0.5.0"
 dependencies = [
  "assert_matches",
  "backoff",
@@ -3364,7 +3364,7 @@ dependencies = [
 
 [[package]]
 name = "pcs"
-version = "0.7.3"
+version = "0.8.0"
 dependencies = [
  "anyhow",
  "b64-ct",
--- a/intel-sgx/dcap-artifact-retrieval/Cargo.toml
+++ b/intel-sgx/dcap-artifact-retrieval/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "dcap-artifact-retrieval"
-version = "0.4.2"
+version = "0.5.0"
 authors = ["Fortanix, Inc."]
 license = "MPL-2.0"
 edition = "2018"
@@ -28,7 +28,7 @@ mbedtls = { version = ">=0.12.0, <0.14.0", features = [
     "std",
 ], default-features = false }
 num_enum = { version = "0.7", features = ["complex-expressions"] }
-pcs = { version = "0.7.3", path = "../pcs" }
+pcs = { version = "0.8", path = "../pcs" }
 percent-encoding = "2.3.2"
 pkix = "0.2.0"
 quick-error = "1.1.0"
@@ -45,7 +45,7 @@ rustls-tls = ["reqwest?/rustls-tls"]
 [dev-dependencies]
 assert_matches = "1.5.0"
 yasna = { version = "0.3", features = ["num-bigint", "bit-vec"] }
-pcs = { version = "0.7", path = "../pcs", features = ["verify"] }
+pcs = { version = "0.8", path = "../pcs", features = ["verify"] }
 
 [build-dependencies]
 mbedtls = { version = ">=0.12.0, <0.14.0", features = ["ssl", "x509"] }
--- a/intel-sgx/pcs/Cargo.toml
+++ b/intel-sgx/pcs/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "pcs"
-version = "0.7.3"
+version = "0.8.0"
 authors = ["Fortanix, Inc."]
 license = "MPL-2.0"
 edition = "2018"


intel-sgx/dcap-artifact-retrieval/src/provisioning_client/intel.rs
```diff
@@ -565,6 +570,69 @@ impl<'inp> ProvisioningServiceApi<'inp> for TcbEvaluationDataNumbersApi {
     }
 }
 
+pub struct FmspcsApi;
```
raoulstrackx: I don't think we'll ever use this API. But by adding it, we'll need to maintain it, and test it. Best to remove it (you can keep it on a separate branch if you think you need it personally)

intel-sgx/dcap-artifact-retrieval/src/provisioning_client/intel.rs
```diff
@@ -359,21 +363,21 @@ impl<'inp> TcbInfoService<'inp> for TcbInfoApi {
...
 
     fn build_request(&self, input: &Self::Input) -> Result<(String, Vec<(String, String)>), Error> {
         let api_version = input.api_version as u8;
         let fmspc = input.fmspc.as_bytes().to_hex();
         let url = if let Some(evaluation_data_number) = input.tcb_evaluation_data_number {
             format!(
-                "{}/sgx/certification/v{}/tcb?fmspc={}&tcbEvaluationDataNumber={}",
-                INTEL_BASE_URL, api_version, fmspc, evaluation_data_number)
+                "{}/{}/certification/v{}/tcb?fmspc={}&tcbEvaluationDataNumber={}",
+                INTEL_BASE_URL, T::new(), api_version, fmspc, evaluation_data_number)
```
raoulstrackx: Nice trick, but it's a bit brittle here. Changing the `Display` implementation of `SGX`  (e.g., to "Intel SGX") shouldn't result in errors here. You could introduce a new trait in this `dcap-artifact-retrieval` crate (e.g., `AsUrlPart`) and implement it here for `SGX` and `TDX`. You also don't need to create an instance of those types.

intel-sgx/dcap-artifact-retrieval/src/provisioning_client/mod.rs
```diff
@@ -134,6 +135,21 @@ impl WithApiVersion for PcsVersion {
     }
 }
 
+#[derive(Copy, Clone, PartialEq, Eq, Hash)]
+pub enum TeePlatform {
```
raoulstrackx: There's already a `Platform` enum with those values in `pcs` can't you reuse that one?

intel-sgx/enclave-runner/src/lib.rs
```diff
@@ -5,7 +5,6 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 #![allow(non_local_definitions)] // Required by failure
-#![deny(warnings)]
```
raoulstrackx, line 8: Please put that back

intel-sgx/pcs/src/qe_identity.rs
```diff
@@ -341,14 +347,14 @@ impl QeIdentitySigned {
     }
 
     pub fn write_to_file(&self, output_dir: &str) -> Result<String, Error> {
-        let id = QeIdentity::<Unverified>::try_from(self)?;
+        let id = QeIdentity::<T, Unverified>::try_from(self)?;
         let filename = Self::create_filename(Some(id.tcb_evaluation_data_number));
         io::write_to_file(&self, output_dir, &filename)?;
         Ok(filename)
```
raoulstrackx: If you have both an SGX and TDX QeIdentity, you won't be able to store them both in the same directory. That's problematic.

intel-sgx/pcs/src/tcb_info.rs
```diff
@@ -227,10 +227,11 @@ pub struct TcbData<V: VerificationType = Verified> {
     tcb_evaluation_data_number: u64,
     tcb_levels: Vec<TcbLevel>,
     type_: PhantomData<V>,
+    platform_: PhantomData<T>
```
raoulstrackx: For TDX there's also `tdxModule`

intel-sgx/pcs/src/qe_identity.rs
```diff
@@ -95,7 +95,7 @@ impl TcbLevel {
 
 #[derive(Clone, Serialize, Debug)]
 #[serde(rename_all = "camelCase")]
-pub struct QeIdentity<V: VerificationType = Verified> {
+pub struct QeIdentity<T: PlatformType, V: VerificationType = Verified> {
```
raoulstrackx: This changes the public API, so this would be a breaking change. Hence, the version of `pcs` should be bumped to `0.8.0`. Similarly for the `dcap-artifact-retrieval` crate

intel-sgx/enclave-runner/src/lib.rs
gilanghamidy: This will be removed we should land the other PR first. This is for ensuring the build only

intel-sgx/pcs/src/qe_identity.rs
gilanghamidy: Okay, I was bumping the patch version, but I will update the minor now.

intel-sgx/dcap-artifact-retrieval/src/provisioning_client/mod.rs
gilanghamidy: This was an experimental one I forgot to remove :)

intel-sgx/pcs/src/tcb_info.rs
gilanghamidy: Ah yeah I completely missed that one. On it now.

intel-sgx/dcap-artifact-retrieval/src/provisioning_client/azure.rs
```diff
@@ -265,6 +267,13 @@ mod tests {
         }
     }
 
+    pub fn qe_identity_tdx() {
```
raoulstrackx: Intel refers to this enclave as TD QE. So can you call this function `td_qe_identity` instead?

intel-sgx/dcap-artifact-retrieval/src/provisioning_client/azure.rs
```diff
@@ -265,6 +267,13 @@ mod tests {
         }
     }
 
+    pub fn td_qe_identity() {
```
raoulstrackx: You're missing `#[test]` here.

intel-sgx/dcap-artifact-retrieval/src/provisioning_client/intel.rs
```diff
@@ -12,18 +12,18 @@
...
 use std::borrow::Cow;
+use std::marker::PhantomData;
 use std::time::Duration;
 
 use super::common::*;
 use super::{
     Client, ClientBuilder, Fetcher, PckCertIn, PckCertService, PckCertsIn, PckCertsService,
     PckCrlIn, PckCrlService, PcsVersion, ProvisioningServiceApi, QeIdIn, QeIdService, StatusCode,
-    TcbEvaluationDataNumbersIn, TcbEvaluationDataNumbersService, TcbInfoIn, TcbInfoService, WithApiVersion,
+    TcbEvaluationDataNumbersIn, TcbEvaluationDataNumbersService, TcbInfoIn, TcbInfoService, WithApiVersion,
```
raoulstrackx: Unintended change.

intel-sgx/pcs/src/tcb_info.rs
```diff
@@ -226,14 +333,31 @@ pub struct TcbData<V: VerificationType = Verified> {
...
+    fn deserialize<D>(deserializer: D) -> Result<TcbData<T, Unverified>, D::Error>
     where
         D: Deserializer<'de>,
     {
+        T::deserialize(deserializer)
+    }
+}
+
+pub trait PlatformTypeWithTcbDeserializer<T: PlatformType> : PlatformType {
+    fn deserialize<'de, D>(deserializer: D) -> Result<TcbData<T, Unverified>, D::Error>
```
raoulstrackx: `PlatformType` is already a super trait of `PlatformTypeWithTcbDeserializer` so why do you need a generic type here?

intel-sgx/pcs/src/pckcrt.rs
```diff
@@ -114,6 +114,7 @@ impl From<u8> for IntelSgxTcbComponentV4 {
 struct IntelSgxTcbComponentsV4 {
     pub sgxtcbcomponents: [IntelSgxTcbComponentV4; 16],
     pub pcesvn: u16,
+    pub tdxtcbcomponents: Option<[IntelSgxTcbComponentV4; 16]>
```
raoulstrackx: An `IntelSgxTcbComponentV4` must not be confused with an TDX TCB component. So it's better to use a different type for it.

intel-sgx/pcs/src/pckcrt.rs
```diff
@@ -114,6 +114,7 @@ impl From<u8> for IntelSgxTcbComponentV4 {
 struct IntelSgxTcbComponentsV4 {
```
raoulstrackx: This name is confusing. Components that apply to both SGX and TDX can use TEE as name.

intel-sgx/pcs/src/tcb_info.rs
```diff
@@ -226,14 +333,31 @@ pub struct TcbData<V: VerificationType = Verified> {
     tcb_type: u16,
     tcb_evaluation_data_number: u64,
     tcb_levels: Vec<TcbLevel>,
+    tdx_module: Option<TdxModule>,
+    tdx_module_identities: Option<Vec<TdxModuleIdentity>>,
```
raoulstrackx: How can these fields be accessed? Since you already have a type specifier, can you avoid an `unwrap` on this `Option`? For a `TcbData<SGX>` those getters simply shouldn't exist

intel-sgx/pcs/src/qe_identity.rs
```diff
@@ -325,42 +351,50 @@ impl QeIdentitySigned {
...
 
     pub fn write_to_file(&self, output_dir: &str) -> Result<String, Error> {
         let id = QeIdentity::<Unverified>::try_from(self)?;
-        let filename = Self::create_filename(Some(id.tcb_evaluation_data_number));
+        let filename = Self::create_filename(&self.enclave_type, Some(id.tcb_evaluation_data_number));
         io::write_to_file(&self, output_dir, &filename)?;
         Ok(filename)
     }
 
     pub fn write_to_file_if_not_exist(&self, output_dir: &str) -> Result<Option<PathBuf>, Error> {
```
raoulstrackx: Since we're changing the interface of these types anyway, can you do something like:
```
struct WriteOption {
    overwrite: bool
}
...
pub fn write_to_file(&self, outdir: &Path, options: WriteOptions){}
```

intel-sgx/pcs/src/tcb_info.rs
gilanghamidy: This is to introduce the `T` of `PlatformType` type so it can be passed as generic to `TcbData<T, Unverified>` in the `Result` type. Then the making it super trait allows to simplify the other user codes. 

Instead of using the generic constraint `PlatformType + PlatformTypeDeserializer<T>`, the relevant code can just use a single `PlatformTypeWithDeserializer<T>` which will also definitely requires the type to have `PlatformType` trait.

intel-sgx/pcs/src/pckcrt.rs
gilanghamidy: Basically the content is the same. I propose to just rename the type instead of copy and pasting the same code.

intel-sgx/pcs/src/pckcrt.rs
gilanghamidy: Yeah we can rename this instead.

intel-sgx/pcs/src/qe_identity.rs
gilanghamidy: Ah yeah I remember we discussed about it "last year". Okay, I think we should simplify this API to avoid confusion.

intel-sgx/pcs/src/tcb_info.rs
gilanghamidy: Ah yeah you are right. I think I miss to make the acessor function for that field

gilanghamidy:
I am thinking about it again, I think the `PckCrt` needs to also carry the generic type, as SGX and TDX `PckCrt` will carry different `TcbComponent` element depending on the TEE platform type. I think I need another major refactoring in this

intel-sgx/pcs/src/tcb_info.rs
gilanghamidy: Refactored with more complex type systems

intel-sgx/pcs/src/pckcrt.rs
gilanghamidy: Renamed, but I think some other types are still confusing. Waiting for further suggestion

intel-sgx/pcs/src/pckcrt.rs
gilanghamidy: Refactored with more complex type system

intel-sgx/pcs/src/io.rs
```diff
@@ -13,30 +13,40 @@ use serde::de::DeserializeOwned;
 
 use crate::Error;
 
+pub struct WriteOption {
```
raoulstrackx: Can you make this plural?

intel-sgx/pcs/src/io.rs
```diff
@@ -13,30 +13,40 @@ use serde::de::DeserializeOwned;
...
+    pub overwrite: bool
```
raoulstrackx: Can you make this private? The idea is to have a default implementation, and have a setter. (A `new` function would also be nice.) Future changes would could then use an additional field, and no code changes would be required. You can take a look at [Thread Builder](https://doc.rust-lang.org/std/thread/struct.Builder.html) for an example of a Builder pattern.

intel-sgx/dcap-artifact-retrieval/src/provisioning_client/intel.rs
```diff
@@ -450,16 +455,27 @@ impl<'inp> ProvisioningServiceApi<'inp> for QeIdApi {
     type Output = QeIdentitySigned;
 
     fn build_request(&self, input: &Self::Input) -> Result<(String, Vec<(String, String)>), Error> {
+        let base_platform: &'static str = match self.enclave_identity_to_request {
+            EnclaveIdentity::QE | EnclaveIdentity::QVE | EnclaveIdentity::QAE => "sgx",
+            EnclaveIdentity::TDQE => "tdx",
```
raoulstrackx: This is confusing. Above it is stated that this is for `QeIdentitySigned`. So this should only be for the QE3 identity. Have the same type for all enclave identities may result in mistakenly using the wrong identity. You could make this generic for Enclave Identities and add a generic type, but that also wouldn't need an enum here.
We don't have plans to use the TDQE, QVE nor QAE. You could still use a generic to make it compatible should we ever want to add such support. For now, let's ensure we don't have to support things we don't need.

intel-sgx/dcap-artifact-retrieval/src/provisioning_client/mod.rs
```diff
@@ -562,12 +589,16 @@ pub trait ProvisioningClient {
         qe_id: Option<&QeId>,
     ) -> Result<PckCert<Unverified>, Error>;
 
-    fn tcbinfo(&self, fmspc: &Fmspc, evaluation_data_number: Option<u16>) -> Result<TcbInfo, Error>;
+    fn tcbinfo(&self, fmspc: &Fmspc, evaluation_data_number: Option<u16>) -> Result<TcbInfo<platform::SGX>, Error>;
+
+    fn tcbinfo_tdx(&self, fmspc: &Fmspc, evaluation_data_number: Option<u16>) -> Result<TcbInfo<platform::TDX>, Error>;
```
raoulstrackx: Could you use the generic in the function call as well?

intel-sgx/dcap-artifact-retrieval/src/provisioning_client/mod.rs
```diff
@@ -562,12 +589,16 @@ pub trait ProvisioningClient {
...
 
     fn pckcrl(&self, ca: DcapArtifactIssuer) -> Result<PckCrl<Unverified>, Error>;
 
     fn qe_identity(&self, evaluation_data_number: Option<u16>) -> Result<QeIdentitySigned, Error>;
 
+    fn tdqe_identity(&self, evaluation_data_number: Option<u16>) -> Result<QeIdentitySigned, Error>;
```
raoulstrackx: Could you use the generic in the function call as well?

intel-sgx/pcs/src/lib.rs
```diff
@@ -55,6 +56,70 @@ pub enum Platform {
     TDX,
 }
 
+pub trait PlatformType : Into<Platform> + Display + Clone {
+    fn new() -> Self;
+    fn tag() -> &'static str;
```
raoulstrackx: Isn't `tag` only used to construct urls in the DCAP artifact retrieval? Then it'd be cleaner to have a separate trait for that in that crate, that's _not_ public.

intel-sgx/pcs/src/lib.rs
```diff
@@ -55,6 +56,70 @@ pub enum Platform {
     TDX,
 }
 
+pub trait PlatformType : Into<Platform> + Display + Clone {
+    fn new() -> Self;
```
raoulstrackx: Is this used?

intel-sgx/pcs/src/lib.rs
```diff
@@ -55,6 +56,70 @@ pub enum Platform {
     TDX,
 }
 
+pub trait PlatformType : Into<Platform> + Display + Clone {
```
raoulstrackx: This defeats the use of a generic a bit, because new variants need to map back to either `SGX` or `TDX`. I don't think you need this `Platform` enum, but only require the `PlatformType`

intel-sgx/pcs/src/pckcrt.rs
```diff
@@ -69,7 +70,7 @@ impl TryFrom<i64> for SGXType {
 
 /// TCB component as specified in the Intel PCKCrt API v3 and v4
 #[derive(Serialize, Deserialize, Debug, Default)]
-struct IntelSgxTcbComponentsV3 {
+pub struct IntelSgxTcbComponentsV3 {
```
raoulstrackx: Why does this need to be `pub` now?

intel-sgx/pcs/src/tcb_info.rs
```diff
@@ -214,26 +217,141 @@ impl<'de> Deserialize<'de> for AdvisoryID {
...
+    #[derive(Debug)]
+    pub struct TDX {
+        pub(crate) tdx_module: TdxModule,
+        pub(crate) tdx_module_identities: Vec<TdxModuleIdentity>,
+    }
 }
 
-impl<'de> Deserialize<'de> for TcbData<Unverified> {
-    fn deserialize<D>(deserializer: D) -> Result<TcbData<Unverified>, D::Error>
+impl PlatformTypeForTcbInfo<platform::SGX> for platform::SGX {
```
raoulstrackx: I think you can do this simpler by parsing with a "unified" struct (using resulting in default impl where fields are missing). And ensuring that you have different impl for each generic. For example: `TcbInfo<TDX>` has a `tdx_module` function, that `TcbInfo<SGX>` simply does not have).

intel-sgx/pcs/src/io.rs
gilanghamidy: Okay, I was thinking of making it builder pattern as well last time, but I tried to timebox myself and came up with the easiest solution for now. I'll note it down

intel-sgx/dcap-artifact-retrieval/src/provisioning_client/intel.rs
gilanghamidy: Yeah last time we are still considering using TDQE, so it is there. I will just revert this one completely and remains that we support only the regular QE.

intel-sgx/dcap-artifact-retrieval/src/provisioning_client/mod.rs
gilanghamidy: I will check again about it, I thought because the generic is not used in the type parameter, it would not work.

intel-sgx/dcap-artifact-retrieval/src/provisioning_client/mod.rs
gilanghamidy: Same as above

intel-sgx/pcs/src/lib.rs
gilanghamidy: The `Platform` enum is there for backward compatibility to the existing enum types initially. I haven't check completely where the `Platform` enum is being used and if it is safe to remove it completely.

intel-sgx/pcs/src/tcb_info.rs
gilanghamidy: But then there will be `unwrap` as we may need to use `Option` type or the likes for that. It was previously written like that until your last comment to avoid `unwrap` in the accessor function.

intel-sgx/pcs/src/pckcrt.rs
gilanghamidy: Because the type is used in function signature of a `pub trait`

intel-sgx/pcs/src/lib.rs
gilanghamidy: Just rechecked the code, the enum is used as a field in struct that uses by Serde JSON parsing. So it may still need to exist, unless we want to use the `struct SGX` and `struct TDX` as the "field" and make custom deserializer to match that ID value.

Example: tcb_info.rs:357

intel-sgx/pcs/src/lib.rs
gilanghamidy: Yes for default instantiation of the generic type in custom serializer/deserializer

intel-sgx/pcs/src/lib.rs
gilanghamidy: I am in progress removing this enum altogether and use type system as differentiator

intel-sgx/dcap-artifact-retrieval/src/provisioning_client/mod.rs
gilanghamidy: Checked, this is impossible as it requires the entire `Client`, `Fetcher` type to be parameterized as well. We have to do it like this.

intel-sgx/dcap-artifact-retrieval/src/provisioning_client/pccs.rs
```diff
@@ -814,51 +819,4 @@ mod tests {
...
-            assert_eq!(verified_qe_id.tcb_evaluation_data_number(), u64::from(number));
-
-            let tcb_info = client
-                    .tcbinfo(&fmspc, Some(number))
-                    .unwrap()
-                    .verify(&root_cas, Platform::SGX, 2)
-                    .unwrap();
-            assert_eq!(tcb_info.tcb_evaluation_data_number(), u64::from(number));
-        }
-    }
```
gilanghamidy, line 874: I remove this test because it is exactly the same as the `intel` provider unit test and the `pccs` provider actually uses the `intel` provider instead. I assume currently PCCS server does not directly support this API.

intel-sgx/pcs/src/pckcrt.rs
```diff
@@ -508,7 +675,7 @@ impl PckCert<Unverified> {
 impl PckCert<Verified> {
     /// Selects the highest matching TCB level
     /// see <https://api.portal.trustedservices.intel.com/documentation#pcs-tcb-info-v2>
-    pub fn find_tcb_state<V: VerificationType>(&self, tcb_data: &TcbData<V>) -> Option<TcbLevel> {
+    pub fn find_tcb_state<V: VerificationType>(&self, tcb_data: &TcbData<platform::SGX, V>) -> Option<TcbLevel<platform::SGX>> {
```
gilanghamidy: When checking our code that is doing verification, I think this method should accept any TCB data, not specifically only SGX TCB data. We found out that SGX and TDX TCB may have different SVN results, which I currently don't know which one that is embedded into the PCK Certificate.

I may have to refactor this again to reflect this function accepting TCB data for any `T`, not only for `platform::SGX`

intel-sgx/pcs/src/io.rs
```diff
@@ -13,27 +13,43 @@ use serde::de::DeserializeOwned;
...
-///
-/// - Returns `Ok(None)` if file already exist.
-/// - Returns `Ok(Some(filename))` if succeed to write to new file.
-pub fn write_to_file_if_not_exist<T: serde::ser::Serialize>(
-    obj: &T,
-    dir: &str,
-    filename: &str,
-) -> Result<Option<PathBuf>, Error> {
+/// Write given object in json to given filename under given dir.
+pub fn write_to_file<T: serde::ser::Serialize>(obj: &T, dir: &str, filename: &str, options: WriteOptions) -> Result<Option<PathBuf>, Error> {
```
Taowyoo, line 50: Could we keep the rustdoc as detailed as old code ?

intel-sgx/pcs/src/tcb_info.rs
```diff
@@ -214,31 +217,209 @@ impl<'de> Deserialize<'de> for AdvisoryID {
...
+        })
+
+    }
+    
+    fn extra_extension() -> &'static str {
+        ""
+    }
+}
+
+#[allow(dead_code)]
```
Taowyoo, line 354: I think these dead_code and unused should be able to remove since they should be coverred by your tests?

intel-sgx/pcs/src/pckcrt.rs
```diff
@@ -92,69 +93,238 @@ struct IntelSgxTcbComponentsV3 {
...
+    
+    fn try_from_v3(_: TcbComponentsV3) -> Result<TcbComponents<platform::TDX>, Error> {
+        Err(Error::InvalidTcbInfo("Attempting to convert TcbComponentsV3 into TcbComponentsV4<TDX>".to_string()))
+    }
+}
+
+
+#[derive(Clone, Debug, Default, Eq, PartialEq)]
+pub struct TcbComponentsV4<T: PlatformTypeForTcbComponent<T>> {
+    sgxtcbcomponents: [TcbComponentEntry; 16],
```
Taowyoo, line 175: typo `sgxtcbcomponents` ? I assume this should be generic to both sgx and tdx?

intel-sgx/dcap-artifact-retrieval/src/provisioning_client/azure.rs
Taowyoo, line 1: could we add tdx test cases

intel-sgx/pcs/src/pckcrt.rs
gilanghamidy: Nope, the TDX TCB has also `sgxtcbcomponents` as part of its TCB Info

intel-sgx/dcap-artifact-retrieval/src/provisioning_client/azure.rs
gilanghamidy: For Azure, it resorts to the Intel PCS server anyway. So there's no need for specific test case. Perhaps we can just generalize the test script for all different provisioning clients to reduce code repetition.

intel-sgx/pcs/src/tcb_info.rs
gilanghamidy: I'll check again, but I think it was introduced because if we don't build for the test, it will be a dead code as it contains `Serializer` or `Deserializer` that is not called within the code.

intel-sgx/pcs/src/io.rs
gilanghamidy: There are many missing documentations actually. We'll go after this PR to improve the documentations

Taowyoo: sure, we can do this later

intel-sgx/dcap-artifact-retrieval/src/provisioning_client/azure.rs
```diff
@@ -46,10 +46,13 @@ impl AzureProvisioningClientBuilder {
         let pck_cert = PckCertApi::new(self.api_version.clone());
         let pck_crl = PckCrlApi::new(self.api_version.clone());
         let qeid = QeIdApi::new(self.api_version.clone());
-        let tcbinfo = TcbInfoApi::new(self.api_version.clone());
-        let evaluation_data_numbers = TcbEvaluationDataNumbersApi::new(INTEL_BASE_URL.into());
+        let sgx_tcbinfo = TcbInfoApi::new(self.api_version.clone());
+        let tdx_tcbinfo = TcbInfoApi::new(self.api_version.clone());
+        let sgx_evaluation_data_numbers = TcbEvaluationDataNumbersApi::new(INTEL_BASE_URL.into());
+        let tdx_evaluation_data_numbers = TcbEvaluationDataNumbersApi::new(INTEL_BASE_URL.into());
+
```
tvsfx: There's spaces on this line and L55 is too long; try to (more-or-less; doesn't have to be perfect) format new code that you add. (Rust analyzer supports region formatting, or you can format the entire thing and git add selectively)

intel-sgx/pcs/src/tcb_info.rs
```diff
@@ -214,71 +217,183 @@ impl<'de> Deserialize<'de> for AdvisoryID {
...
+#[allow(unused)]
+fn tdx_attribute_serializer<S>(attribute: u64, serializer: S) -> ::std::result::Result<S::Ok, S::Error>
+where
+    S: ::serde::Serializer,
+{
+    let attribute = format!("{attribute:16x}");
+    serializer.serialize_str(&attribute)
+}
+
+pub trait PlatformTypeForTcbInfo<T: PlatformTypeForTcbInfo<T>> : PlatformType + PlatformTypeForTcbComponent<T> {
```
tvsfx: I dont get why `trait PlatformTypeForTcbInfo<T: PlatformTypeForTcbInfo<T>>` isnt just `trait PlatformTypeForTcbInfo`; this bound causes a proliferation of bounds throughout your PR.

tvsfx:
Did one round of simplifying trait bounds.
Tasks to complete before further review of `pcs` crate:

- [ ]  Correct naming ([reference](https://rust-lang.github.io/api-guidelines/naming.html)) - e.g., use `Sgx` and `Tdx` instead of all caps, don't start errors/logs with capital letters
- [ ] Combine the 3 different traits (`PlatformTypeForTcbComponent` , `PlatformType`, `PlatformTypeforTcbInfo`) into 1; I dont think we really need these separately often enough to split them. Move the resulting trait to some logical location.
- [ ] Get rid of `platform_specific` modules and make type names more descriptive (e.g. `SgxComponents` instead of `Sgx`) to help programmers (when you define an inline module, consider if it should be its own module file instead). Watch for naming. Move to same file as trait above, together with all `impl` blocks related to this trait (and perhaps the auxiliary `TryFromTcbComponentsV3` I defined?). Shorten all usage paths once types are descriptive (e.g. `SgxComponents` instead of `platform_specific::SGX`).
- [ ] Reduce number of type bounds in all traits. You almost never need stuff like `type PlatformSpecificTcbComponentData : Debug + Eq + Clone + Default + Serialize + DeserializeOwned + TryFromTcbComponentsV3` or `pub trait PlatformType : Display + Clone + Send + Default + DeserializeOwned {`; you can probably remove most of these bounds.
- [ ] decide what to do with `decompose_raw_cpusvn` function

gilanghamidy:
>    Correct naming ([reference](https://rust-lang.github.io/api-guidelines/naming.html)) - e.g., use `Sgx` and `Tdx` instead of all caps, don't start errors/logs with capital letters

Then it will be inconsistent with the rest of existing naming convention in the crates, e.g.: https://github.com/fortanix/rust-sgx/blob/5f3fdcfddcd0b62391aee965b8979214664ca44d/intel-sgx/pcs/src/pckcrt.rs#L43C1-L50C1

>   Combine the 3 different traits (`PlatformTypeForTcbComponent` , `PlatformType`, `PlatformTypeforTcbInfo`) into 1; I dont think we really need these separately often enough to split them. Move the resulting trait to some logical location.

`PlatformType` is to be used in other context outside of the `TcbInfo` structures. It was initially also to be used for `QeIdentity` type as well and it has zero relation with `TcbInfo` structures. I only agree merging `PlatformTypeForTcbComponent` and `PlatformTypeforTcbInfo` which is definitely more make sense.

>Get rid of `platform_specific` modules and make type names more descriptive (e.g. `SgxComponents` instead of `Sgx`) to help programmers (when you define an inline module, consider if it should be its own module file instead). Watch for naming. Move to same file as trait above, together with all `impl` blocks related to this trait (and perhaps the auxiliary `TryFromTcbComponentsV3` I defined?). Shorten all usage paths once types are descriptive (e.g. `SgxComponents` instead of `platform_specific::SGX`).

As we discussed online, there is no such thing as `SgxComponents` and it does not make sense to make an `SgxComponents` as empty struct as it is semantically confusing to the reader. `platform_specific` nomenclature is clearer as it indicates there is no platform-specific data for SGX type, and there are some for TDX type. Besides, that `platform_specific` module is used as internal namespace and is not intended for external user of this crate to use. That's why I insist on using `platform::SGX` to indicate every generic type as it will not burden every other programmers that use our crate to find out which proper type to call, such as: `TcbComponent<SgxComponent>` that is uglier than `TcbComponent<platform::SGX>` as in current state.

> Reduce number of type bounds in all traits. You almost never need stuff like `type PlatformSpecificTcbComponentData : Debug + Eq + Clone + Default + Serialize + DeserializeOwned + TryFromTcbComponentsV3` or `pub trait PlatformType : Display + Clone + Send + Default + DeserializeOwned {`; you can probably remove most of these bounds.

Many of those bounds are required to get compiled, because the call side (for instance, calling `println!("{:#?}", tcb_info);`) requires `Debug` to be implemented over the generic.

>  decide what to do with `decompose_raw_cpusvn` function

We can just move it as free function, although I currently also don't understand the purpose of checking `self.tcb_type` there that basically does not make any much difference as `tcb_type` is always zero at the moment (and perhaps in the foreseeable future)

intel-sgx/dcap-artifact-retrieval/src/provisioning_client/pccs.rs
```diff
@@ -254,21 +257,23 @@ impl<'inp> ProvisioningServiceApi<'inp> for PckCrlApi {
     }
 }
 
-pub struct TcbInfoApi {
+pub struct TcbInfoApi<T: PlatformType> {
```
tvsfx: We should definitely add the following to our TODOs ticket:
Investigate proliferation of `T` parameters to distinguish between SGX and TDX. 
I think we can make the distinction between both cases more judiciously, but that will take serious refactoring.

intel-sgx/pcs/src/lib.rs
```diff
@@ -49,25 +50,58 @@ pub type PceIsvsvn = u16;
 pub type QeId = [u8; 16];
 pub use crate::pckid::PckID;
 
-#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, Eq)]
-pub enum Platform {
-    SGX,
-    TDX,
+///Global trait that specify the required interface for typesafe enumeration of platforms.
+pub trait PlatformType : Display + Clone + Send + Default + DeserializeOwned {
```
tvsfx: Reduce bounds?

tvsfx: You said that `PlatformType` is used externally/independently and can therefore not be merged into `PlatformTypeForTcbComponent` and `PlatformTypeforTcbInfo`. However, I'm not seeing where that is the case.

intel-sgx/pcs/src/lib.rs
```diff
@@ -185,18 +219,28 @@ impl TryFrom<&str> for DcapArtifactIssuer {
     }
 }
 
-pub trait VerificationType {}
+pub trait VerificationType : Serialize {
```
tvsfx: do we have scenarios were we want to serialize unverified structures?

intel-sgx/pcs/src/lib.rs
```diff
@@ -185,18 +219,28 @@ impl TryFrom<&str> for DcapArtifactIssuer {
...
 
 impl VerificationType for Verified {}
 
-#[derive(Clone, Debug, Eq, PartialEq)]
+#[derive(Clone, Serialize, Debug, Eq, PartialEq)]
 pub struct Unverified;
 
 impl VerificationType for Unverified {}
 
+impl<'de> Deserialize<'de> for Unverified {
```
tvsfx, line 226: Can you put a big warning on `Verified` explaining that we purposefully do not implement `Deserialize` for it? Also mention that it should _never_ be combined with `serde(skip)`. Since `Verified` does not implement `Default`, it cannot be used with `serde(skip)` currently, which is good, but if you make a second mistake and add `PhantomData<Verified>` then things can go wrong, since `PhantomData` is always `Deserialize`.

Also, why can you not just `#[derive(Deserialize)]`  for Unverified? Does that give problems with the old serde version for some reason?

intel-sgx/pcs/src/pckcrl.rs
```diff
@@ -19,41 +17,21 @@ use {
...
+use crate::io::WriteOptions;
 use crate::io::{self};
 use crate::{DcapArtifactIssuer, Error, Unverified, VerificationType, Verified};
 
-#[derive(Clone, Serialize, Debug, Eq, PartialEq)]
+#[derive(Clone, Serialize, Deserialize, Debug, Eq, PartialEq)]
 pub struct PckCrl<V: VerificationType = Verified> {
     crl: String,
     ca_chain: Vec<String>,
-    #[serde(skip)]
```
tvsfx, line 29: Very happy to see how much this simplified things 🥳

intel-sgx/pcs/src/pckcrl.rs
```diff
@@ -19,41 +17,21 @@ use {
...
-        }
-
-        let Dummy { crl, ca_chain } = Dummy::deserialize(deserializer)?;
-        Ok(PckCrl::<Unverified> {
-            crl,
-            ca_chain,
-            type_: PhantomData,
-        })
-    }
+    type_: V,
```
tvsfx, line 29: naming: `_<field>` signifies that it will not be used - not sure why the `_` is at the end here (I know you didnt introduce this)

intel-sgx/pcs/src/pckcrt.rs
```diff
@@ -92,69 +93,157 @@ struct IntelSgxTcbComponentsV3 {
...
+    #[derive(Serialize, Deserialize, Clone, Debug, Default, Eq, PartialEq)]
+    pub struct SGX {}
+
+    #[derive(Serialize, Deserialize, Clone, Debug, Default, Eq, PartialEq)]
+    pub struct TDX {
+        pub(crate) tdxtcbcomponents: [TcbComponentEntry; 16],
+    }
+}
+
+pub trait PlatformTypeForTcbComponent
```
tvsfx, line 122: Please review this: this is my code. (All traits stuff should probably be moved to its own file)

intel-sgx/pcs/src/pckcrt.rs
```diff
@@ -92,69 +93,157 @@ struct IntelSgxTcbComponentsV3 {
...
+        TcbComponents(TcbComponentsV4 {
+            sgxtcbcomponents: raw_cpusvn.map(|svn| svn.into()),
+            pcesvn,
+            platform_specific_data: platform_specific::SGX {}
+        })
+    }
+}
+
+impl TcbComponents<platform_specific::TDX> {
+    pub fn from_raw(raw_cpusvn: [u8; 16], pcesvn: u16, raw_tdxsvn: [u8; 16]) -> Self {
```
tvsfx, line 224: note that this function is only used in tests, apart from being in our public interface
which begs the question - do the from_raw functions need to be public? if clients are supposed to always deserialize, they might not have to be

intel-sgx/pcs/src/pckcrl.rs
```diff
@@ -117,27 +95,14 @@ impl<V: VerificationType> PckCrl<V> {
...
-    }
-
-    pub fn write_to_file_as(&self, output_dir: &str, ca: DcapArtifactIssuer) -> Result<String, Error> {
-        let filename = Self::filename_from_ca(ca);
-        io::write_to_file(&self, output_dir, &filename)?;
-        Ok(filename)
-    }
-
-    #[cfg(feature = "verify")]
-    pub fn write_to_file_if_not_exist(&self, output_dir: &str) -> Result<Option<PathBuf>, Error> {
```
tvsfx, line 133: Do we expect other options than `disallow_overwrite`? If not you could just have an enum `AllowsOverwrite`.
I think having `WriteOptionsBuilder` is overkill right now; just make `disallow_overwrite` a function on `WriteOptions` for now.

intel-sgx/dcap-artifact-retrieval/src/provisioning_client/azure.rs
```diff
@@ -304,12 +307,12 @@ mod tests {
...
 
     #[test]
-    pub fn tcb_evaluation_data_numbers() {
+    pub fn sgx_tcb_evaluation_data_numbers() {
         for api_version in [PcsVersion::V3, PcsVersion::V4] {
             let client = AzureProvisioningClientBuilder::new(api_version)
                 .set_retry_timeout(TIME_RETRY_TIMEOUT)
                 .build(reqwest_client());
-            assert!(client.tcb_evaluation_data_numbers().is_ok());
+            assert!(client.sgx_tcb_evaluation_data_numbers().is_ok());
```
tvsfx, line 352: does azure not support TDX, or why do we not need tests for that?

intel-sgx/dcap-artifact-retrieval/src/provisioning_client/intel.rs
```diff
@@ -853,13 +905,52 @@ mod tests {
...
 
+    #[test]
+    pub fn tcb_info_tdx() {
+        let intel_builder = IntelProvisioningClientBuilder::new(PcsVersion::V4)
+            .set_retry_timeout(TIME_RETRY_TIMEOUT);
+        let client = intel_builder.build(reqwest_client());
+        let root_ca = include_bytes!("../../tests/data/root_SGX_CA_der.cert");
+        let root_cas = [&root_ca[..]];
+
+        // List of knowns FMSPCS that has valid TDX TCB
```
tvsfx, line 928: is this from an authoritative source? if so, can you add it? Is there a risk that these numbers change and the test starts failing?

intel-sgx/dcap-artifact-retrieval/src/provisioning_client/intel.rs
```diff
@@ -853,13 +905,52 @@ mod tests {
...
+            let tdx_tcbinfo = client.tdx_tcbinfo(&item, None);
+            println!("FMSPC: {} => {}", item.to_string(), tdx_tcbinfo.is_ok());
+            assert!(tdx_tcbinfo.is_ok());
+
+            let _ = tdx_tcbinfo.unwrap().verify(&root_cas, 2).unwrap();
+        }
+    }
+
     #[test]
     pub fn tcb_info_with_evaluation_data_number() {
```
tvsfx, line 956: Could you reuse this test for TDX by inputting a different list of fmspc's, and passing in a couple closures where the behavior is different? (e.g. `sgx_tcbinfo` vs `tdx_tcbinfo`) - I'm trying to understand why the SGX test above is separate and looks so different from this one.

intel-sgx/dcap-artifact-retrieval/src/provisioning_client/intel.rs
```diff
@@ -1091,43 +1192,58 @@ mod tests {
...
         let eval_numbers2 = serde_json::ser::to_vec(&eval_numbers)
-            .and_then(|v| serde_json::from_slice::<RawTcbEvaluationDataNumbers>(&v))
+            .and_then(|v| serde_json::from_slice::<RawTcbEvaluationDataNumbers<T>>(&v))
             .unwrap();
         assert_eq!(eval_numbers, eval_numbers2);
 
-        let fmspc = Fmspc::try_from("90806f000000").unwrap();
-        let eval_numbers: TcbEvaluationDataNumbers =
-            eval_numbers.verify(&root_cas, Platform::SGX).unwrap();
+        let fmspc = Fmspc::try_from("50806f000000").unwrap();
```
tvsfx, line 1224: Why did this change?

intel-sgx/dcap-artifact-retrieval/src/provisioning_client/azure.rs
```diff
@@ -46,10 +46,13 @@ impl AzureProvisioningClientBuilder {
...
-        let tcbinfo = TcbInfoApi::new(self.api_version.clone());
-        let evaluation_data_numbers = TcbEvaluationDataNumbersApi::new(INTEL_BASE_URL.into());
+        let sgx_tcbinfo = TcbInfoApi::new(self.api_version.clone());
+        let tdx_tcbinfo = TcbInfoApi::new(self.api_version.clone());
+        let sgx_evaluation_data_numbers = TcbEvaluationDataNumbersApi::new(INTEL_BASE_URL.into());
+        let tdx_evaluation_data_numbers = TcbEvaluationDataNumbersApi::new(INTEL_BASE_URL.into());
+         
         self.client_builder
-            .build(pck_certs, pck_cert, pck_crl, qeid, tcbinfo, evaluation_data_numbers, fetcher)
+            .build(pck_certs, pck_cert, pck_crl, qeid, sgx_tcbinfo, tdx_tcbinfo, sgx_evaluation_data_numbers, tdx_evaluation_data_numbers, fetcher)
```
tvsfx: I wish we could do the following:
 
```rust
pub struct TcbEvaluationDataNumbers {
    base_url: Cow<'static, str>,
}

pub struct TcbEvaluationDataNumbersApi<T> { 
    data: TcbEvaluationDataNumbers
    _platform: PhantomData<T>
}
```
and similar for `TcbInfoApi`, then we could make the distinction between the 2 cases as late as possible in our builder, and we wouldnt have to expose this to our clients. 
Unfortunately, that would require refactoring `ClientBuilder::build`

intel-sgx/dcap-artifact-retrieval/src/provisioning_client/intel.rs
```diff
@@ -1091,43 +1192,58 @@ mod tests {
         }
     }
 
-    #[test]
-    pub fn tcb_evaluation_data_numbers() {
+    fn tcb_evaluation_data_numbers_test_base<T: PartialEq>()
```
tvsfx, line 1208: Note that I put the `PartialEq` bound on `T` here, so you do not have to require it on the Trait

intel-sgx/dcap-artifact-retrieval/src/provisioning_client/mod.rs
```diff
@@ -645,9 +690,38 @@ pub trait ProvisioningClient {
...
     }
 
-    fn tcb_evaluation_data_numbers(&self) -> Result<RawTcbEvaluationDataNumbers, Error>;
+    fn sgx_tcb_evaluation_data_numbers(&self) -> Result<RawTcbEvaluationDataNumbers<platform::SGX>, Error>;
+
+    fn tdx_tcb_evaluation_data_numbers(&self) -> Result<RawTcbEvaluationDataNumbers<platform::TDX>, Error>;
+}
+
+
+pub trait ProvisioningClientFuncSelector<T: PlatformTypeForTcbInfo> {
```
tvsfx: This is only used in `test`, right? Make it `cfg(test)` and not public. Also, you can avoid having to define a trait by just passing in closures (although this is fine by me) . Additionally, you can remove the `T: PlatformTypeForTcbInfo` and just write 
```rust
    fn get_tcb_evaluation_data_numbers<'pc>(pc: &'pc dyn ProvisioningClient) -> Result<RawTcbEvaluationDataNumbers<Self>, Error> where Self: Sized;
```

intel-sgx/dcap-artifact-retrieval/src/provisioning_client/mod.rs
```diff
@@ -645,9 +690,38 @@ pub trait ProvisioningClient {
...
+    fn get_tcb_evaluation_data_numbers<'pc>(pc: &'pc dyn ProvisioningClient) -> Result<RawTcbEvaluationDataNumbers<T>, Error>;
```
tvsfx: nit: no need to specify `pc`

intel-sgx/dcap-artifact-retrieval/src/provisioning_client/mod.rs
```diff
@@ -675,9 +749,14 @@ impl<F: for<'a> Fetcher<'a>> ProvisioningClient for Client<F> {
         self.pckcert_service.call_service(&self.fetcher, &input)
     }
 
-    fn tcbinfo(&self, fmspc: &Fmspc, tcb_evaluation_data_number: Option<u16>) -> Result<TcbInfo, Error> {
-        let input = self.tcbinfo_service.pcs_service().build_input(fmspc, tcb_evaluation_data_number);
-        self.tcbinfo_service.call_service(&self.fetcher, &input)
+    fn sgx_tcbinfo(&self, fmspc: &Fmspc, tcb_evaluation_data_number: Option<u16>) -> Result<TcbInfo<platform::SGX>, Error> {
+        let input = self.sgx_tcbinfo_service.pcs_service().build_input(fmspc, tcb_evaluation_data_number);
```
tvsfx, line 753: For the TODO list: We really have to find a way to have 1 generic service that we can invoke with a platform, so we can get rid of this duplication.

intel-sgx/dcap-artifact-retrieval/src/provisioning_client/pccs.rs
tvsfx: Get rid of this bound on `T` (below as well)

intel-sgx/pcs/src/tcb_evaluation_data_numbers.rs
```diff
@@ -13,74 +16,43 @@ use {
...
     version: u16,
     #[allow(unused)]
+    #[serde(with = "crate::iso8601")]
     issue_date: DateTime<Utc>,
     #[allow(unused)]
+    #[serde(with = "crate::iso8601")]
     next_update: DateTime<Utc>,
     #[allow(unused)]
     tcb_eval_numbers: Vec<TcbEvalNumber>,
+    #[serde(skip)]
```
tvsfx: Why do you have `skip`? That seems dangerous. I think you should do as above: make `V` a field.

intel-sgx/pcs/src/tcb_evaluation_data_numbers.rs
```diff
@@ -13,74 +16,43 @@ use {
...
-            issue_date,
-            next_update,
-            tcb_eval_numbers,
-            type_: PhantomData,
-        })
-    }
-}
-
-impl<V: VerificationType> TcbEvaluationDataNumbers<V> {
+impl<T: PlatformTypeForTcbInfo, V: VerificationType> TcbEvaluationDataNumbers<T, V> {
```
tvsfx, line 39: bounds?

intel-sgx/pcs/src/tcb_evaluation_data_numbers.rs
```diff
@@ -13,74 +16,43 @@ use {
...
-    pub fn select_best(input_dir: &str, fmspc: &Fmspc, tcb_components: &TcbComponents, qesvn: u16) -> Result<TcbEvalNumber, Error> {
-        let evalnums = RawTcbEvaluationDataNumbers::read_from_file(input_dir)?.evaluation_data_numbers()?;
+    pub fn select_best(input_dir: &str, fmspc: &Fmspc, tcb_components: &TcbComponents<T::PlatformSpecificTcbComponentData>, qesvn: u16) -> Result<TcbEvalNumber, Error> {
+        let evalnums = RawTcbEvaluationDataNumbers::<T>::read_from_file(input_dir)?.evaluation_data_numbers()?;
         let mut tcb_levels: std::collections::HashMap<_, _> = evalnums.numbers().map(|num| (num.number as u64, (num, None, None))).collect();
 
-        for tcbinfo in TcbInfo::read_all(input_dir, fmspc) {
-            let tcb_data = TcbData::parse(tcbinfo?.raw_tcb_info())?;
+        for tcbinfo in TcbInfo::<T>::read_all(input_dir, fmspc) {
+            let tcb_data = TcbData::<T, Unverified>::parse(tcbinfo?.raw_tcb_info())?;
```
tvsfx: nit: spurious type annotation?

intel-sgx/pcs/src/tcb_evaluation_data_numbers.rs
```diff
@@ -175,25 +147,29 @@ impl TcbEvalNumber {
...
     raw_tcb_evaluation_data_numbers: String,
     signature: Vec<u8>,
     ca_chain: Vec<String>,
+    #[serde(skip)]
+    platform_: PhantomData<T>,
 }
 
-impl RawTcbEvaluationDataNumbers {
-    const DEFAULT_FILENAME: &'static str = "tcb_evaluation_data_numbers.numbers";
+impl<T: PlatformTypeForTcbInfo> RawTcbEvaluationDataNumbers<T> {
```
tvsfx, line 157: bound not needed for some of these methods (potentially add to TODOs)

intel-sgx/pcs/src/tcb_evaluation_data_numbers.rs
```diff
@@ -447,31 +414,31 @@ mod tests {
     #[cfg(not(target_env = "sgx"))]
     #[test]
     fn select_best() {
```
tvsfx, line 415: test for TDX necessary?

intel-sgx/pcs/src/tcb_info.rs
```diff
@@ -214,71 +221,182 @@ impl<'de> Deserialize<'de> for AdvisoryID {
...
+    }
+}
+
+fn tdx_mrsigner_deserializer<'de, D: Deserializer<'de>>(deserializer: D) -> Result<[u8; 48], D::Error> {
+    let mrsigner = String::deserialize(deserializer)?;
+    let mrsigner = base16::decode(&mrsigner).map_err(de::Error::custom)?;
+    mrsigner.as_slice().try_into().map_err(de::Error::custom)
+}
+
+#[allow(unused)]
```
tvsfx, line 308: why is this annotation necessary? remove if unused, also below

intel-sgx/pcs/src/lib.rs
tvsfx:
```rust
/// ```rust
/// # use static_assertions::assert_not_impl_any;
/// assert_not_impl_any!(Unverified, Deserialize);
/// ```
```

we could consider adding `static_assertions` to the dev-dependencies and making a test like this

intel-sgx/dcap-artifact-retrieval/src/provisioning_client/intel.rs
```diff
@@ -853,13 +905,52 @@ mod tests {
...
+            Fmspc::try_from("50806f000000").unwrap(),
+            Fmspc::try_from("20a06e050000").unwrap(),
+            Fmspc::try_from("10a06f010000").unwrap(),
+            Fmspc::try_from("b0c06f000000").unwrap(),
+            Fmspc::try_from("20a06f000000").unwrap(),
+            Fmspc::try_from("60a06f000000").unwrap(),
+            Fmspc::try_from("c0806f000000").unwrap(),
+            Fmspc::try_from("20a06d080000").unwrap(),
+            Fmspc::try_from("10a06d000000").unwrap(),
+            Fmspc::try_from("00806f050000").unwrap(),
```
tvsfx, line 942: nit: use `map` or convert later so you dont need to repeat `try_from` 16 times.

intel-sgx/dcap-artifact-retrieval/src/provisioning_client/pccs.rs
tvsfx: > the pccs provider actually uses the intel provider
I'm not sure what this means; doesnt this test contact [a different server](https://github.com/fortanix/rust-sgx/blob/3cb2949ada88de1c041d499381e5af2f1283d465/intel-sgx/dcap-artifact-retrieval/src/provisioning_client/pccs.rs?plain=1#L466-L471)? If anything, does that mean this test should use a common implementation? (cc: @raoulstrackx )

intel-sgx/pcs/src/pckcrt.rs
tvsfx: consider renaming with `rename =` to `sgx_tcb_components` here and elsewhere to follow rust naming

intel-sgx/pcs/src/pckcrt.rs
```diff
@@ -180,18 +269,18 @@ impl TcbComponents {
...
+    pub fn tcb_component_index(&self, comp: TcbComponentType) -> Option<usize> {
         self.0.sgxtcbcomponents
             .iter()
-            .position(|c| TcbComponent::try_from(c.comp_type.as_str()) == Ok(comp))
+            .position(|c| TcbComponentType::try_from(c.comp_type.as_str()) == Ok(comp))
     }
 }
 
-impl PartialOrd for TcbComponents {
+impl<T, U> PartialOrd<TcbComponents<U>> for TcbComponents<T> {
```
tvsfx, line 272: Does is make sense to order when `T != U`? If not, restrict the impl

intel-sgx/pcs/src/pckcrt.rs
```diff
@@ -204,14 +293,24 @@ impl PartialOrd for TcbComponents {
...
         }
-
         prev
     }
 }
 
-impl std::convert::From<IntelSgxTcbComponentsV3> for IntelSgxTcbComponentsV4 {
-    fn from(c: IntelSgxTcbComponentsV3) -> Self {
-        IntelSgxTcbComponentsV4 {
+impl<T, U> PartialEq<TcbComponents<U>> for TcbComponents<T> {
```
tvsfx, line 293: Does it make sense to/do you need to compare Sgx and Tdx components? If not you can just derive(PartialEq)

intel-sgx/pcs/src/pckcrt.rs
```diff
@@ -368,7 +467,7 @@ impl PckCerts {
 
     /// Order all PCKs according to the tcb info
     /// TCB Info is carefully ordered by Intel
-    fn order_pcks<V: VerificationType>(&self, tcb_info: &TcbData<V>) -> Vec<PckCert<Unverified>> {
+    fn order_pcks<T: PlatformTypeForTcbInfo, V: VerificationType>(&self, tcb_info: &TcbData<T, V>) -> Vec<PckCert<Unverified>> {
```
tvsfx, line 463: bounds?

intel-sgx/pcs/src/pckcrt.rs
```diff
@@ -1276,20 +1357,22 @@ mod tests {
     #[test]
     fn tcb_components_cpu_svn() {
         let raw_cpu_svn = [10, 20, 30, 40, 50, 60, 70, 80, 90, 100, 110, 120, 130, 140, 150, 160];
-        let comp = TcbComponents::from_raw(raw_cpu_svn, 42);
+        let comp = TcbComponents::<platform_specific::SGX>::from_raw(raw_cpu_svn, 42);
```
tvsfx: tests for TDX?

intel-sgx/pcs/src/qe_identity.rs
```diff
@@ -423,7 +348,7 @@ impl QeIdentitySigned {
             mrsigner,
             isvprodid,
             tcb_levels,
-            type_: PhantomData,
+            ..
```
tvsfx: write `type_: _` ; no point using `..` if you list almost all fields

intel-sgx/pcs/src/tcb_evaluation_data_numbers.rs
```diff
@@ -447,31 +414,31 @@ mod tests {
...
-        use crate::pckcrt::TcbComponents;
-        fn select(tcb_components: &TcbComponents, qesvn: u16) -> Result<u16, Error> {
+        use crate::pckcrt::{platform_specific, TcbComponents};
+        fn select(tcb_components: &TcbComponents<platform_specific::SGX>, qesvn: u16) -> Result<u16, Error> {
             use std::convert::TryInto;
```
tvsfx, line 418: test for TDX?

intel-sgx/pcs/src/tcb_info.rs
```diff
@@ -214,71 +221,182 @@ impl<'de> Deserialize<'de> for AdvisoryID {
...
+
+impl Default for TdxModule {
+    fn default() -> Self {
+        Self { mrsigner: [0; 48], attributes: Default::default(), attributes_mask: Default::default() }
+    }
+}
+
+#[derive(Clone, Debug, Deserialize, Eq, PartialEq)]
+#[serde(rename_all = "camelCase")]
+pub struct TdxModuleIdentity {
```
tvsfx, line 243: Does this work?
```
#[derive(Clone, Debug, Deserialize, Eq, PartialEq)]
#[serde(rename_all = "camelCase")]
pub struct TdxModuleIdentity {
    #[serde(flatten)]
    tdx_module: TdxModule,
    tcb_levels: Vec<TdxModuleTcbLevel>,
}
```

intel-sgx/pcs/src/tcb_info.rs
```diff
@@ -214,71 +221,182 @@ impl<'de> Deserialize<'de> for AdvisoryID {
...
-            type_: PhantomData,
-        })
-    }
+    #[serde(rename = "tcbLevels")]
+    tcb_levels: Vec<TcbLevel<T::PlatformSpecificTcbComponentData>>,
+    #[serde(flatten)]
+    platform_specific_data: T::PlatformSpecificTcbData,
+    type_: V,
+    #[serde(skip)]
+    platform_: PhantomData<T>
```
tvsfx: why is this `Phantom` necessary?

intel-sgx/pcs/src/tcb_info.rs
tvsfx, line 308: how is this unused?

intel-sgx/pcs/src/tcb_info.rs
```diff
@@ -214,71 +221,182 @@ impl<'de> Deserialize<'de> for AdvisoryID {
...
+    }
+}
+
 #[allow(dead_code)]
-#[derive(Clone, Debug)]
-pub struct TcbData<V: VerificationType = Verified> {
-    id: Platform,
+#[derive(Clone, Debug, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct TcbData<T: PlatformTypeForTcbInfo, V: VerificationType = Verified> {
```
tvsfx, line 367: `V: VerificationType` bound is not necessary afaict, here and elsewhere.

intel-sgx/pcs/src/tcb_info.rs
```diff
@@ -578,4 +706,35 @@ mod tests {
...
+    fn parse_tcbinfo_json() {
+        let ca_chain = vec![
+            include_str!("../tests/data/cert-chain/cert-chain-1.cert").to_string(),
+            include_str!("../tests/data/cert-chain/cert-chain-2.cert").to_string()
+        ];
+
+        let root_certificate = include_bytes!("../tests/data/root_SGX_CA_der.cert");
+        let root_certificates = [&root_certificate[..]];
+
+        let januari_30_2026 = Utc.with_ymd_and_hms(2026, 1, 30, 12, 0, 0).unwrap();
```
tvsfx, line 707: all caps for constant

intel-sgx/pcs/src/tcb_info.rs
```diff
@@ -506,57 +636,55 @@ mod tests {
...
         let root_certificates = [&root_certificate[..]];
-        match info.verify(&root_certificates, Platform::SGX, 2) {
+        match info.verify(&root_certificates, 2) {
             Err(Error::InvalidTcbInfo(msg)) => assert_eq!(msg, String::from("TCB Info expired on 2020-06-17 17:49:24 UTC")),
             e => assert!(false, "wrong result: {:?}", e),
         }
         let juni_5_2020 = Utc.with_ymd_and_hms(2020, 6, 5, 12, 0, 0).unwrap();
-        assert!(info.verify_ex(&root_certificates, Platform::SGX, 2, &juni_5_2020).is_ok());
-
-        match info.verify(&root_certificates, Platform::TDX, 2) {
```
tvsfx, line 529: where did this call go? shouldnt we have `verify/verify_ex` calls for TDX here and below?

intel-sgx/pcs/tests/data/qe-identity/qeid-tdx.json
tvsfx, line 1: is this used anywhere?

intel-sgx/pcs/src/lib.rs
gilanghamidy: It will be used externally in our own internal repo. You'll see later ;)

intel-sgx/pcs/src/pckcrl.rs
gilanghamidy: I saw inconsistencies in using underscore to signify unused vars. I'll just use the `_<name>` format going forward

intel-sgx/pcs/src/pckcrt.rs
gilanghamidy: There's possibility since we can get `tdxsvn` from TDX's `REPORTDATA` structure which is in byte array, as in getting `cpusvn`.

intel-sgx/pcs/src/pckcrl.rs
gilanghamidy: This is Raoul's request for future-proofing

intel-sgx/dcap-artifact-retrieval/src/provisioning_client/azure.rs
gilanghamidy: because Azure's provider also contact Intel's PCS server for getting TCB info. So the test is actually covered there.

intel-sgx/dcap-artifact-retrieval/src/provisioning_client/intel.rs
gilanghamidy: FMSPC is a device model identifier. Once it supports TDX, it **will always** support TDX. I enumerate that by spamming the PCS server initially.

intel-sgx/dcap-artifact-retrieval/src/provisioning_client/intel.rs
gilanghamidy: No the thing is, TDX's FMSPC is a subset of SGX one. So a valid FMSPC will always have SGX's TCB info, but not TDX.

intel-sgx/dcap-artifact-retrieval/src/provisioning_client/intel.rs
gilanghamidy: I think I am just trying out another valid FMSPC for that API

intel-sgx/dcap-artifact-retrieval/src/provisioning_client/azure.rs
gilanghamidy: Let's make another ticket for KTLO :P

intel-sgx/dcap-artifact-retrieval/src/provisioning_client/mod.rs
gilanghamidy: No, that trait will be used in our internal codes. That's why now it is in `mod.rs`. It is basically a simple selector for getting `tcbinfo` from the PCS client based on `PlatformType`

intel-sgx/pcs/src/tcb_evaluation_data_numbers.rs
gilanghamidy: Wdym. TDX **is** SGX

intel-sgx/pcs/src/pckcrt.rs
gilanghamidy: We don't have that yet

intel-sgx/pcs/src/tcb_info.rs
gilanghamidy: Technically this is valid. And I think I also miss a field there.

intel-sgx/pcs/src/tcb_info.rs
gilanghamidy: It is there. Since beginning.

intel-sgx/pcs/tests/data/qe-identity/qeid-tdx.json
gilanghamidy: Not anymore because we decide to not use TDQE

intel-sgx/pcs/src/tcb_info.rs
gilanghamidy: The code can no longer mistakenly receive TDX TCB Info if it is called via the SGX API as it is typechecked. The original code is no longer relevant nor valid.

intel-sgx/pcs/src/lib.rs
gilanghamidy: Reduced to `Clone` and `Default`

intel-sgx/pcs/src/lib.rs
gilanghamidy: We always serialize unverified structure as verified structure is only occur in runtime. We never persist verified structure. But removing this bound will trigger bound error.

intel-sgx/pcs/src/lib.rs
gilanghamidy: Test case is failed if using attribute `Deserialize`. So I have to handcraft the `Deserialize`. The thing is, this value is nonexistent in the JSON, so using the attribute will expect there is some "field" named `verification` or something that has some values.

intel-sgx/dcap-artifact-retrieval/src/provisioning_client/mod.rs
gilanghamidy: Created a ticket for it.

intel-sgx/pcs/src/tcb_evaluation_data_numbers.rs
gilanghamidy: Changed with non `PhantomData` type

intel-sgx/pcs/src/pckcrt.rs
```diff
@@ -92,69 +93,157 @@ struct IntelSgxTcbComponentsV3 {
...
+    V3(TcbComponentsV3),
+    V4(TcbComponentsV4<P>),
 }
 
-#[derive(Serialize, Deserialize, Clone, Debug, Default, Eq, PartialEq)]
-#[serde(from = "IntelSgxTcbComponents")]
-pub struct TcbComponents(IntelSgxTcbComponentsV4);
+#[derive(Serialize, Deserialize, Clone, Debug, Default, Eq)]
+#[serde(try_from = "TcbComponentsCompatibilitySelector<P>")]
+#[serde(bound(deserialize = "TcbComponents<P>: TryFrom<TcbComponentsV3>, <TcbComponents<P> as TryFrom<TcbComponentsV3>>::Error: std::fmt::Display, P : Deserialize<'de>"))]
```
tvsfx: please drop commit `42320d88ce55e4b467f71b6c684768724fae788f` and run `cargo update -p syn@2.0.100` instead; that will update syn to a version that supports bounds on associated types
(see https://github.com/serde-rs/serde/issues/3033)

intel-sgx/pcs/src/tcb_info.rs
gilanghamidy: The other code writes like this: tcb_info.rs:659. I am just following what's done in previous codes, unless you want it to change there as well.

intel-sgx/pcs/src/tcb_info.rs
gilanghamidy: I guess it will be compiled if and only if the library user uses the `Serialize` against `TdxModuleIdentity`. Inside this library, there's no such use yet, but maybe eventually will be.

intel-sgx/pcs/src/tcb_evaluation_data_numbers.rs
gilanghamidy: Because I need to specify `Unverified`. Otherwise, the `TcbData` defaults to resolve to `Verified`. We should change the default into `Unverified` if we want to simplify this, but I think that was the original intention anyway.

intel-sgx/pcs/src/tcb_info.rs
gilanghamidy: May be required.

intel-sgx/pcs/src/pckcrt.rs
gilanghamidy: It may be allowed to compare `TcbComponents<TDX>` against `TcbComponents<SGX>` because the ordering is currently only on the component that they share each other. It is used in PCK Cert comparison that only has SGX TCB components.

intel-sgx/pcs/src/lib.rs
gilanghamidy: Changed into `default`. I should have been used that trait in the first place.

intel-sgx/pcs/src/pckcrt.rs
tvsfx: In fact, this partial-eq implementation is almost certainly not what you want; it ignores platform_specific_data, whereas the PartialEq for TcbComponentsV4 does account for platform_specific_data. So by unwrapping 1 layer 2 instances might suddenly become unequal. If you do really need this, define it as a function, not a PartialEq`.

intel-sgx/pcs/src/pckcrt.rs
tvsfx: I'm not really asking if it may happen; I'm saying that if you don't really need this, I would not implement cross-type partial-ord like this; you can never account for the third field.

gilanghamidy: Then it cannot be compiled.

intel-sgx/pcs/src/tcb_evaluation_data_numbers.rs
gilanghamidy: Need to collect all test case from actual TDX machine which is time consuming in the current state of work. Will be done later.

intel-sgx/pcs/src/tcb_info.rs
gilanghamidy: Yeah forgot to remove it actually.

intel-sgx/dcap-artifact-retrieval/src/provisioning_client/pccs.rs
gilanghamidy: If you see the `Client` implementation for `PCCS`, the `tcb_evaluation_data_number` API actually calls the Intel's PCS instance, instead of implementing its own PCCS call. It may be because the PCCS server (at that time) does not implement the API yet, and/or our internal PCCS server does not support it yet.

PR #856, opened 2026-01-04
[RTE-476] Exposing private `PckCrl` function and Fix CI for unused `unsafe` block due to recent change

gilanghamidy:
Exposing `PckCrl::ca` and `PckCrl::as_mbedtls_crl` to public so library user can utilize the functions.

### DIFF
--- a/intel-sgx/enclave-runner/src/loader.rs
+++ b/intel-sgx/enclave-runner/src/loader.rs
@@ -167,6 +167,7 @@ impl<'a> EnclaveBuilder<'a> {
 
     fn generate_xfrm(max_ssaframesize_in_pages: u32) -> u64 {
         fn cpuid(eax: u32, ecx: u32) -> Option<CpuidResult> {
+            #![allow(unused_unsafe)]
             unsafe {
                 if eax <= x86_64::__get_cpuid_max(0).0 {
                     Some(x86_64::__cpuid_count(eax, ecx))


Taowyoo (approved):
CI failed with `#[warn(unused_unsafe)]` you can explicitly allow them now, sice internal project may not upgrade to latest nightly in short time.

intel-sgx/pcs/src/pckcrl.rs
```diff
@@ -153,15 +153,15 @@ impl<V: VerificationType> PckCrl<V> {
     }
 
     #[cfg(feature = "verify")]
-    pub(crate) fn as_mbedtls_crl(&self) -> Result<Crl, Error> {
+    pub fn as_mbedtls_crl(&self) -> Result<Crl, Error> {
```
raoulstrackx: This shouldn't be exposed for `PckCrl<Unverified>`. I'm also not too thrilled on the name for a public function

intel-sgx/enclave-runner/src/lib.rs
```diff
@@ -5,6 +5,7 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 #![allow(non_local_definitions)] // Required by failure
+#![allow(unused_unsafe)]
```
raoulstrackx: Can you put this on the relevant `unsafe` blocks instead?

intel-sgx/pcs/src/pckcrl.rs
gilanghamidy: Reverting it as maybe it is not going to be used

intel-sgx/pcs/src/pckcrl.rs
```diff
@@ -161,7 +161,7 @@ impl<V: VerificationType> PckCrl<V> {
     }
 
     #[cfg(feature = "verify")]
-    fn ca(&self) -> Result<DcapArtifactIssuer, Error> {
+    pub fn ca(&self) -> Result<DcapArtifactIssuer, Error> {
```
raoulstrackx, line 164: Does this need to be exposed on unverified PckCrls as well?

gilanghamidy: This is as required in our internal changes, and putting it only on `Verified` state is undesirable as per your request for flexibility.

PR #857, opened 2026-01-18
[RTE-680] Add TDX types to `sgx-isa` crate and create `tdx-ql` crate

Taowyoo:
## Summary
Add TDX attestation support via a new `tdx-ql` crate, split common layout macros into a standalone `memory-layout` crate, and update CI setup.

## Changes
- Add `intel-tdx/tdx-ql` crate with TDX report + RTMR extend APIs implemented by `nix::ioctl` (default) via `/dev/tdx_guest`.
- Move shared layout macros (`struct_def!`, `impl_default_clone_eq!`) into new `memory-layout` crate and wire `sgx-isa` feature flags through it.
- Add TDX report structs/constants to `intel-sgx/sgx-isa/src/tdx.rs` and export via `sgx-isa`.
- Add `tdx-ql` and `memory-layout` to the workspace.
- CI: use `install_build_deps.sh`, add `tdx-ql` test step; release workflow accepts `tdx-isa_vX.Y.Z` tags.
- Add `tdx_ql_cli` example cli program for manual testing.
- Add intel tdx report verify API through adding EVERIFYREPORT2 instruction support. Ref: [Intel CPU Architectural Extensions Specification](https://cdrdv2.intel.com/v1/dl/getContent/733582) section 3.1

## Testing
- Unit tests added in `tdx-ql` for report/extend error paths.


intel-tdx/tdx-isa/build.rs
```diff
@@ -0,0 +1,3 @@
+fn main() {
+    println!("cargo:rerun-if-env-changed=SGX_SDK");
```
raoulstrackx: We should probably document why this `SGX_SDK` variable exists.

raoulstrackx: I also saw that you stripped out the `tdx-resource` directory (or I forgot to commit them). Was that intentional?

intel-tdx/tdx-isa/src/lib.rs
```diff
@@ -0,0 +1,248 @@
...
+
+impl From<TdxReport> for tdx_report_t {
+    fn from(report: TdxReport) -> Self {
+        tdx_report_t { d: report.0 }
+    }
+}
+
+impl TdxReport {
+    /// Create an empty report buffer.
+    pub fn new() -> Self {
```
raoulstrackx: Do we need this?

intel-tdx/tdx-isa/src/lib.rs
```diff
@@ -0,0 +1,248 @@
...
+        Self([0; TDX_REPORT_LEN])
+    }
+
+    /// Return the raw report bytes.
+    pub fn read_raw(&self) -> &[u8; TDX_REPORT_LEN] {
```
raoulstrackx: Is this needed? I wanted to have a layout of `TdxReport`, similar to how we define `Report` in the `sgx-isa` crate. It makes accessing its fields quite easy. Exposing the raw data here complicates that.

intel-tdx/tdx-isa/src/lib.rs
```diff
@@ -0,0 +1,248 @@
...
+/// Size of the RTMR extend data field in bytes.
+pub const TDX_RTMR_EXTEND_DATA_LEN: usize = 48;
+
+/// The generated TDX Report returned by the TDX module.
+#[derive(PartialEq, Eq, Debug, Clone)]
+pub struct TdxReport([u8; TDX_REPORT_LEN]);
+
+/// A TDX Quote (opaque byte blob) returned by the platform quoting stack.
+#[derive(PartialEq, Eq, Debug, Clone)]
+pub struct TdxQuote(Vec<u8>);
```
raoulstrackx: We won't use the TDX remote attestation enclave (TD QE). Can we remove this, so we don't have to test it?

intel-tdx/tdx-isa/src/lib.rs
```diff
@@ -0,0 +1,248 @@
...
+/// Extend one of the TDX runtime measurement registers (RTMRs).
+///
+/// Please check doc of [`TdxRtmrEvent`] for details of input.
+pub fn extend_rtmr(rtmr_event: TdxRtmrEvent) -> Result<(), TdxAttestError> {
+    let bytes: Vec<u8> = rtmr_event.try_into()?;
+    tdx_attest::parse_tdx_attest_error(tdx_attest::tdx_att_extend(&bytes))
+}
+
+/// Retrieve the list of attestation key IDs supported by the platform.
+pub fn get_supported_tdx_attestation_key_ids() -> Result<Vec<Uuid>, TdxAttestError> {
```
raoulstrackx: This isn't needed

intel-tdx/tdx-isa/src/lib.rs
```diff
@@ -0,0 +1,248 @@
...
+
+impl TdxRtmrEvent {
+    /// Build a new RTMR extend event.
+    pub fn new(
+        version: u32,
+        rtmr_index: u64,
+        extend_data: [u8; TDX_RTMR_EXTEND_DATA_LEN],
+        event_type: u32,
+        event_data: Vec<u8>,
+    ) -> Self {
```
raoulstrackx: This interface is horrible. By wrapping the Intel `tdx-attest-rs` we can create a nicer one. Can we just do an `extend` function that exposes the only two rtmr indices supported, and the extend data?

intel-tdx/tdx-isa/build.rs
Taowyoo: it is intentional, since it's actually used by `tdx-attest-rs-sys` crate.
And original build method with `build.sh` does not handle "full compile" since build lib does not trigger link but when this library is built with any executable such as test.
When using your `build.sh`  with test, it will fail with cannot found `tdxattest` lib.

Now this is resolved by installing `libtdx-attest-dev`:
- `libtdx-attest-dev` deb pkg provides header file in system path.
- `libtdx-attest` deb pkg provides the tdx attest lib/  It is dependency of `libtdx-attest-dev`.

After installing,  `tdx-attest-rs-sys` will using header and library in system path.

I will add a README.md file to explain these details.

Taowyoo: Good catch. I just realize this crate actually does not need to be recompiled with `SGX_SDK` change but `tdx-attest-rs-sys` should. Should I suppose this will be handled by `cargo` correctly?

Anyway, as mentioned in other comment, I will add a README file to explain this.

intel-tdx/tdx-isa/src/lib.rs
Taowyoo: ok, Let me remove these not used function now.

intel-tdx/tdx-isa/src/lib.rs
Taowyoo: make sense!

intel-tdx/tdx-isa/src/lib.rs
Taowyoo: oh, i did not know this. Let me sync with you on Monday

intel-tdx/tdx-isa/build.rs
raoulstrackx: Yes the `tdx-resource` directory was to avoid having compilation issues when the tdx packages weren't installed. I didn't got that for to include tests. Well, they would fail on my machine anyway.

intel-tdx/tdx-isa/src/lib.rs
Taowyoo: updated please check

intel-tdx/tdx-isa/src/lib.rs
Taowyoo: I have updated `TdxReport` to use memory_layout macro, please review again. thx!

intel-tdx/tdx-isa/build.rs
Taowyoo: added README, please check

intel-tdx/tdx-ql/src/tdx_ioctl.rs
```diff
@@ -0,0 +1,79 @@
...
+    Ok(TdxReport::try_copy_from(&req.td_report).expect("validated size"))
+}
+
+#[repr(C)]
+struct TdxExtendRtmrReq {
+    data: [u8; TDX_RTMR_EXTEND_DATA_SIZE],
+    index: u8,
+}
+
+nix::ioctl_readwrite!(tdx_cmd_extend_rtmr, b'T', 3, TdxExtendRtmrReq);
```
jethrogb: Is this really read-write?

Taowyoo: good catch, updated

intel-sgx/tdx-ql/src/tdx_ioctl.rs
```diff
@@ -0,0 +1,81 @@
...
+    Ok(TdxReport::try_copy_from(&req.td_report).expect("validated size"))
+}
+
+#[repr(C)]
+struct TdxExtendRtmrReq {
+    data: [u8; TDX_RTMR_EXTEND_DATA_SIZE],
+    index: u8,
+}
+
+nix::ioctl_write_ptr!(tdx_cmd_extend_rtmr, b'T', 3, TdxExtendRtmrReq);
```
gilanghamidy, line 52: Is this testable? I could not find `TDX_CMD_EXTEND_RTMR` command in the `ioctl` of `sgx-guest` driver in upstream kernel

https://github.com/torvalds/linux/blob/master/include/uapi/linux/tdx-guest.h#L40

https://github.com/torvalds/linux/blob/162b42445b585cd89f45475848845db353539605/drivers/virt/coco/tdx-guest/tdx-guest.c#L350C1-L360C1

I also cannot use the sample CLI to extend the RTMR on our internal server. This IOCTL ops only exists in the DCAP library, under the compiler flag `V3_DRIVER`:

https://github.com/intel/confidential-computing.tee.dcap/blob/a6c3631330787aaf1e286f5dd06c39c50ac0d796/QuoteGeneration/quote_wrapper/tdx_attest/tdx_attest.c#L42C1-L51C7

Intel has the implementation of the `TDCALL` to extend the RTMR, but I am not sure if it is on the upstream kernel:

https://github.com/intel/confidential-computing.tdx.tdx-guest/blob/a475dd8e3bd20c242d568f4fa25351d6829e5061/src/tdcall.rs#L244C1-L252C2

gilanghamidy: I got more stuff from here, let's discuss internally.

intel-sgx/tdx-ql/src/tdx_ioctl.rs
```diff
@@ -0,0 +1,81 @@
...
+
+    let file = std::fs::OpenOptions::new()
+        .read(true)
+        .write(true)
+        .open(TDX_ATTEST_DEV_PATH)
+        .map_err(|_| TdxAttestErrorCode::DeviceFailure)?;
+
+    unsafe { tdx_cmd_extend_rtmr(file.as_raw_fd(), &mut req) }.map_err(|errno| match errno {
+        nix::errno::Errno::EINVAL => TdxAttestErrorCode::InvalidRtmrIndex,
+        _ => TdxAttestErrorCode::ExtendFailure,
```
gilanghamidy: Please also capture `ENOTTY` error as it signifies that the command is not supported, as implemented in their kernel driver, and many other errors (`EFAULT`. `ENOMEM`) so we can better understand what's the internal error spawned by the kernel driver.

Taowyoo:
> Maybe we should put the `tdx-ql` crate also under `intel-sgx` instead. Because the `sgx-isa` (and later the `pcs` and `dcap-artifact-retrieval`) also consists of TDX stuff, which makes `intel-sgx` directory also has TDX functionality in it. Adding separate `intel-tdx` directory specifically for TDX will make the structure inconsistent and confusing.

Hi @gilanghamidy , good suggestion. I have moved the crate

intel-sgx/sgx-isa/src/tdx.rs
```diff
@@ -0,0 +1,266 @@
...
+    #[repr(C, align(1024))]
+    #[cfg_attr(
+        feature = "large_array_derive",
+        derive(Clone, Debug, Eq, PartialEq)
+    )]
+    pub struct TdxReport {
+        /// (  0) Report mac struct for SGX report type 2
+        pub report_mac: TdxReportMac,
+        /// (256) Struct contains details about extra TCB elements not found in CPUSVN
+        pub tee_tcb_info: [u8; TEE_TCB_INFO_SIZE],
```
gilanghamidy: We should create `TeeTcbInfo` struct that adheres the `TEE_TCB_INFO` in the Table 2.3 of Intel® Trust Domain CPU Architectural Extensions document page 2-6

intel-sgx/sgx-isa/src/tdx.rs
```diff
@@ -0,0 +1,266 @@
...
+        /// (495) Reserved, must be zero
+        pub reserved: [u8; TDX_REPORT_RESERVED_SIZE],
+        /// (512) Struct contains the TEE Info
+        pub tee_info: [u8; TEE_INFO_SIZE],
```
gilanghamidy: This should also adhere to `TDINFO_STRUCT` definition for `tee_info` field as described in TDX Module 1.0 public spec: https://cdrdv2-public.intel.com/733568/tdx-module-1.0-public-spec-344425005.pdf page 163

intel-sgx/tdx-ql/src/tdx_ioctl.rs
Taowyoo: Based on https://github.com/torvalds/linux/blob/master/arch/x86/coco/tdx/tdx.c#L158, there seems no too much mapping provided from `ioctl` level.

So I try to map as much as possible based on general definition of linux errono.
Please check

intel-sgx/sgx-isa/src/tdx.rs
Taowyoo: added, pls check

intel-sgx/sgx-isa/src/tdx.rs
Taowyoo: added, pls check.

From doc, there are two variants of TDINFO_STRUCT with different size. So I introduce two type to represent them, as a result also two variant of TdReport.

I did not update `tdx-ql`, since upstream (kernel) only support first varaint (1024 Bytes) one now.

intel-sgx/tdx-ql/src/tdx_ioctl.rs
gilanghamidy: We don't have to cover everything for now. We can support what is visible from `tdx-guest` kernel driver as shown in their source: https://github.com/intel-staging/tdx/blob/748382694adb8c962751eb8fcbd536ba7af1467f/drivers/virt/coco/tdx-guest/tdx-guest.c#L480C1-L493C2

There are at least:
```
-ENOMEM;
-EFAULT;
-EINTR;
-EBUSY;
-EINVAL;
-EIO;
-EPERM;
-ENOTTY;
-ENODEV;
```

intel-sgx/sgx-isa/src/tdx.rs
```diff
@@ -0,0 +1,515 @@
...
+    /// Version: Sep 2025, 348551-007US
+    /// Link: <https://cdrdv2.intel.com/v1/dl/getContent/733579>
+    #[repr(C, align(512))]
+    #[cfg_attr(
+        feature = "large_array_derive",
+        derive(Clone, Debug, Eq, PartialEq)
+    )]
+    pub struct TdInfoV1 {
+        /// (  0) Base TDINFO fields
+        pub base: [u8; TDINFO_BASE_SIZE],
```
gilanghamidy: On page 65 of the document, `TDINFO_BASE` has a complete definition that we should also extract. This is the important bits

intel-sgx/sgx-isa/src/tdx.rs
```diff
@@ -0,0 +1,515 @@
...
+    ///
+    /// Ref: Intel TDX Module Application Binary Interface (ABI) Reference, table 3.49.
+    /// Version: Sep 2025, 348551-007US
+    /// Link: <https://cdrdv2.intel.com/v1/dl/getContent/733579>
+    #[repr(C)]
+    #[cfg_attr(
+        feature = "large_array_derive",
+        derive(Clone, Debug, Eq, PartialEq)
+    )]
+    pub struct TdInfoV2 {
```
gilanghamidy: There are lots of duplication of V1 vs V2 in here, while the V2 only changes the `TDINFO_EXTENSION` definition that extends the `TDREPORT` size. We should consider a more generic implementation.

At current state, wince you said that only V1 that seems to be supported by TDX system, we can just disregard the V2, remove the V2 from the code for now and put as a new RTE TODO ticket to be picked up in later sprints. So we don't have to maintain/test V2 for now.

Taowyoo: OK I will create the ticket.

Also regarding generic implementation, personally i dont think generic is worth here for FFI data type. We want these type to have size known on compile time. Then if using generic type parameter here, I feel it's kind of over engineering.

intel-sgx/tdx-ql/src/tdx_ioctl.rs
Taowyoo: Have updated the mapping and error variants to map error as much as possible. Please check

intel-sgx/sgx-isa/src/tdx.rs
Taowyoo: added pls check

intel-sgx/sgx-isa/src/tdx.rs
Taowyoo: Created and added doc: `RTE-805`

fortanix-vme/aws-nitro-enclaves/nsm/Cargo.toml
```diff
@@ -8,6 +8,6 @@ edition = "2021"
 
 [dependencies]
 nitro-attestation-verify = { path = "../nitro-attestation-verify" }
-nsm-driver = { git = "https://github.com/aws/aws-nitro-enclaves-nsm-api" }
-nsm-io = { git = "https://github.com/aws/aws-nitro-enclaves-nsm-api" }
+nsm-driver = { git = "https://github.com/aws/aws-nitro-enclaves-nsm-api", rev = "c6c94667709407c2f2f2d458b31f80743031221f" }
```
gilanghamidy: Do we need this in this PR?

intel-sgx/sgx-isa/src/tdx.rs
```diff
@@ -0,0 +1,430 @@
...
+        /// (  8) TEE_TCB_SVN array
+        pub tee_tcb_svn: [u8; TEE_TCB_INFO_TEE_TCB_SVN_SIZE],
+        /// ( 24) Measurement of the Intel TDX module
+        pub mrseam: [u8; TEE_TCB_INFO_MR_SIZE],
+        /// ( 72) Measurement of TDX module signer if valid
+        pub mrsigner_seam: [u8; TEE_TCB_INFO_MR_SIZE],
+        /// (120) Additional configuration ATTRIBUTES if valid
+        pub attributes: [u8; TEE_TCB_INFO_ATTRIBUTES_SIZE],
+        /// (128) Reserved, must be zero
+        pub reserved: [u8; TEE_TCB_INFO_RESERVED_SIZE],
```
gilanghamidy, line 169: For some reason, this is non-zero in our test environment. But we'll just proceed with it regardless.

Taowyoo: From "Intel TDX Module Application Binary Interface (ABI) Reference 348551-007US"
it lists one more field  here , let me add it

fortanix-vme/aws-nitro-enclaves/nsm/Cargo.toml
Taowyoo: Yes, we need this.
Otherwise, the code cannot compile.
This upstream have updated so that we must specify a old rev to able to use it,

Taowyoo: I have create a separate PR for this, please review: https://github.com/fortanix/rust-sgx/pull/897
I also rebased this pr on to it.

PR #858, opened 2026-01-21
[RTE-718] Add fortanix-vme-initramfs package.

sardok:
The new package leaves everything related to AWS Nitro in fortanix-vme-eif and add helpers to build customized initramfs image.

### DIFF
--- a/fortanix-vme/fortanix-vme-initramfs/src/fs_tree.rs
+++ b/fortanix-vme/fortanix-vme-initramfs/src/fs_tree.rs
@@ -29,7 +29,7 @@ const REL_TO_CUR: &str = "./";
 ///     .add_file("etc/config", Cursor::new(b"key=value\n"))
 ///     .build();
 /// ```
-#[derive(Debug, Eq, PartialEq, Default)]
+#[derive(Debug, Default)]
 pub struct FsTree(pub(crate) Vec<FsTreeEntry>);
 
 type CpioInput = (NewcBuilder, Box<dyn ReadSeek>);
@@ -37,6 +37,10 @@ type CpioInput = (NewcBuilder, Box<dyn ReadSeek>);
 impl FsTree {
     pub fn new() -> Self {
         Self(Vec::new())
+    }
+
+    pub fn eq_metadata(&self, other: &Self) -> bool {
+        self.0.iter().zip(&other.0).all(|(l, r)| l.eq_metadata(r))
     }
 
     pub fn add_directory(mut self, dirname: &str) -> FsTree {
@@ -141,7 +145,7 @@ impl FsTree {
 }
 
 #[derive(Derivative)]
-#[derivative(Debug, PartialEq)]
+#[derivative(Debug)]
 pub enum FsTreeEntry {
     Directory {
         path: PathBuf,
@@ -151,7 +155,6 @@ pub enum FsTreeEntry {
         path: PathBuf,
         mode: u32,
         #[derivative(Debug(format_with = "redact"))]
-        #[derivative(PartialEq = "ignore")]
         content: Box<dyn ReadSeek>,
     },
 }
@@ -159,8 +162,6 @@ pub enum FsTreeEntry {
 fn redact<T>(_: &T, f: &mut fmt::Formatter<'_>) -> fmt::Result {
     write!(f, "...")
 }
-
-impl Eq for FsTreeEntry {}
 
 impl FsTreeEntry {
     pub(crate) fn into_cpio_input(self) -> Result<CpioInput, Error> {
@@ -184,6 +185,28 @@ impl FsTreeEntry {
                 let builder = NewcBuilder::new(name).uid(0).gid(0).mode(mode);
                 Ok((builder, content))
             }
+        }
+    }
+
+    /// Compares the metadata of two entries while ignoring the file content.
+    ///
+    /// Returns `true` if both entries are of the same type (both files or both directories)
+    /// and their `path` and `mode` are identical.
+    pub fn eq_metadata(&self, other: &Self) -> bool {
+        match (self, other) {
+            (
+                FsTreeEntry::Directory { path: p1, mode: m1 },
+                FsTreeEntry::Directory { path: p2, mode: m2 },
+            ) => p1 == p2 && m1 == m2,
+            (
+                FsTreeEntry::File {
+                    path: p1, mode: m1, ..
+                },
+                FsTreeEntry::File {
+                    path: p2, mode: m2, ..
+                },
+            ) => p1 == p2 && m1 == m2,
+            _ => false,
         }
     }
 }
@@ -236,7 +259,7 @@ mod tests {
             make_file("./init", Cursor::new(content.clone())),
             make_file("./nsm.ko", Cursor::new(content.clone())),
         ]);
-        assert_eq!(files, expected);
+        assert!(files.eq_metadata(&expected));
     }
 
     #[test]
@@ -251,7 +274,7 @@ mod tests {
             make_directory("./a/b/c/d/e/f"),
             make_file("./a/b/c/d/e/f/g.txt", Cursor::new(vec![])),
         ]);
-        assert_eq!(files, expected);
+        assert!(files.eq_metadata(&expected));
     }
 
     #[test]
@@ -266,13 +289,13 @@ mod tests {
             make_directory("./x"),
             make_directory("./x/z"),
         ]);
-        assert_eq!(files, expected);
+        assert!(files.eq_metadata(&expected));
     }
 
     #[test]
     fn test_partial_eq() {
         let entry1 = make_file("/tmp/a.txt", Cursor::new(vec![1, 2, 3, 4]));
         let entry2 = make_file("/tmp/a.txt", Cursor::new(vec![5, 6, 7, 8]));
-        assert_eq!(entry1, entry2);
+        assert!(entry1.eq_metadata(&entry2));
     }
 }


Cargo.toml
```diff
@@ -35,6 +35,7 @@ members = [
     "intel-sgx/sgxs",
     "ipc-queue",
     "rs-libc",
+    "fortanix-vme/fortanix-vme-initramfs",
```
tvsfx: Nit: sort alphabetically

fortanix-vme/fortanix-vme-eif/Cargo.toml
```diff
@@ -7,18 +7,16 @@ edition = "2021"
...
 [dependencies]
 aws-nitro-enclaves-image-format = "0.5"
-cpio = "0.2"
-serde_json = "1.0"
-sha2 = "0.9"
+fortanix-vme-initramfs = { path = "../fortanix-vme-initramfs" }
 tempfile = "3"
-flate2 = "1.0"
 thiserror = "1.0"
+serde_json = "1.0"
```
tvsfx, line 11: Nit: sort

fortanix-vme/fortanix-vme-eif/src/error.rs
```diff
@@ -1,30 +1,17 @@
...
-    InitramfsExtractError(#[source] io::Error),
-    #[error("Expected trailer in initramfs missing")]
-    InitramfsExpectedTrailer,
-    #[error("Unexpected data in initramfs file {path:?} (found \"{found:?}\", expected \"{expected:?}\")")]
-    InitramfsUnexpectedData {
-        path: String,
-        found: String,
-        expected: String,
-    },
+    #[error("Initramfs error: {0}")]
```
tvsfx: If you include the `InitramfsError` as the `source` of this variant, there's no need to include it in the string representation again (i.e., remove the `:  {0}` part). Otherwise, if you print an error backtrace you will get somethinglike"Initamfs error : <error string> caused by <error string repeated>"

fortanix-vme/fortanix-vme-eif/src/error.rs
```diff
@@ -1,30 +1,17 @@
...
+    InitramfsError(#[source] InitramfsError),
```
tvsfx: if you use `#[from]` instead of `#[source]`, you don't need the `From` implementation below

fortanix-vme/fortanix-vme-eif/src/lib.rs
```diff
@@ -216,63 +245,120 @@ impl<R: Read + Seek + 'static, S: Read + Seek + 'static, T: Read + Seek + 'stati
...
         eifbuilder.write_to(tmp.as_file_mut());
-        tmp.rewind().map_err(|e| Error::EifWriteError(e))?;
-        io::copy(&mut tmp, &mut output).map_err(|e| Error::EifWriteError(e))?;
+        tmp.rewind().map_err(Error::EifWriteError)?;
+        io::copy(&mut tmp, &mut output).map_err(Error::EifWriteError)?;
         Ok(FtxEif::new(output))
     }
 }
 
+mod initramfs {
```
tvsfx: can this be a separate `initramfs.rs` file? that'd be tidier

fortanix-vme/fortanix-vme-eif/src/lib.rs
```diff
@@ -184,26 +195,44 @@ impl<T: Read + Seek> FtxEif<T> {
...
-    pub fn new(name: String, application: R, init: S, nsm: T, kernel: U, kernel_config: V, cmdline: &str) -> Self {
+impl<
+        R: Read + Seek + 'static,
+        S: Read + Seek + 'static,
+        T: Read + Seek + 'static,
+        U: Read + Seek + 'static,
+        V: Read + Seek + 'static,
+    > Builder<R, S, T, U, V>
+{
+    pub fn new(
```
tvsfx, line 228: Not your code, but I'm wondering if we actually need all of these generics? I'd only make things generic like this if we expect to use the generics.

fortanix-vme/fortanix-vme-eif/src/lib.rs
```diff
@@ -216,63 +245,120 @@ impl<R: Read + Seek + 'static, S: Read + Seek + 'static, T: Read + Seek + 'stati
...
+            nsm,
+            kernel: mut image,
+            kernel_config: mut image_config,
+            cmdline,
+        } = self;
 
         // Unfortunately `aws_nitro_enclaves_image_format::EifBuilder` forces us to have data in
         // files.
-        let initramfs = NamedTempFile::new().map_err(|e| Error::InitramfsWriteError(e))?;
-        let initramfs = InitramfsBuilder::new(application, init, nsm)
```
tvsfx, line 224: Also not your code, but I think it was a mistake to nest this `builder` inside another `Builder`. I think it would be clearer if `fortanix-vme/fortanix-vme-eif/lib.rs` first built the `initramfs` and then passed the `initramfs` to the builder in this file.

fortanix-vme/fortanix-vme-eif/src/lib.rs
```diff
@@ -216,63 +245,120 @@ impl<R: Read + Seek + 'static, S: Read + Seek + 'static, T: Read + Seek + 'stati
...
     }
 }
 
+mod initramfs {
+    use super::*;
+    use fortanix_vme_initramfs::Builder as InitramfsBuilder;
+    use fortanix_vme_initramfs::{FsTree, FsTreeBuilder};
+    use std::io::Cursor;
+
+    const CMD_CONTENT: &str = "/bin/a.out";
```
tvsfx: Why is this the hard-coded value?

fortanix-vme/fortanix-vme-initramfs/src/fs_tree.rs
```diff
@@ -0,0 +1,287 @@
...
+///
+/// The builder is consumed on each method call and produces a finalized
+/// [`FsTree`] when [`build`](Self::build) is invoked.
+///
+/// # Example
+///
+/// ```text
+/// use std::io::Cursor;
+///
+/// let fs_tree = FsTreeBuilder::new()
```
tvsfx, line 26: I appreciate the docs!

fortanix-vme/fortanix-vme-initramfs/src/fs_tree.rs
```diff
@@ -0,0 +1,287 @@
...
+    fn dir_exists(&self, dir: &PathBuf) -> bool {
+        self.entries
+            .iter()
+            .any(|e| matches!(e, FsTreeEntry::Directory { path, .. } if path == dir))
+    }
+
+    fn add_directory_with_parents(&mut self, dir: &Path, mode: u32) {
+        let mut to_add = Vec::new();
+        for dir in dir.ancestors() {
+            if dir == Path::new("") || dir == Path::new(".") {
```
tvsfx: why do you need to check "."?
also, is this check needed given that this function seems to only be called with normalized paths?

fortanix-vme/fortanix-vme-initramfs/src/fs_tree.rs
```diff
@@ -0,0 +1,287 @@
...
+        // Add parents first
+        if path.pop() {
+            self.add_directory_with_parents(path.as_path(), DEFAULT_DIR_PERMS);
+        }
+
+        self.entries.push(entry);
+        self
+    }
+
+    pub fn normalize_path(basename: &str) -> PathBuf {
```
tvsfx: Which paths might clients of this builder provide us? There's a lot of things this normalization doesn't currently support (like `../<dirname>` or `././` )
Is there a reason to allow relative paths in the first place? Can we just disallow them?
If not, this might be an option: https://docs.rs/normalize-path/latest/normalize_path/

fortanix-vme/fortanix-vme-initramfs/src/fs_tree.rs
```diff
@@ -0,0 +1,287 @@
...
+/// use std::io::Cursor;
+///
+/// let fs_tree = FsTreeBuilder::new()
+///     .add_directory("bin")
+///     .add_executable("bin/init", Cursor::new(b"#!/bin/sh\n"))
+///     .add_file("etc/config", Cursor::new(b"key=value\n"))
+///     .build();
+/// ```
+#[derive(Debug, Default)]
+pub struct FsTreeBuilder {
```
tvsfx: Could you remove `FsTreeBuilder` and just define all of these methods on `FsTree` directly? I use builders when I need complex state validation and have optional fields, but that does not really seem to be the case here?

fortanix-vme/fortanix-vme-initramfs/src/fs_tree.rs
```diff
@@ -0,0 +1,287 @@
...
+            FsTreeEntry::Directory { path, mode } => f
+                .debug_struct("FsTreeEntry::Directory")
+                .field("path", path)
+                .field("mode", mode)
+                .finish(),
+            FsTreeEntry::File { path, mode, .. } => f
+                .debug_struct("FsTreeEntry::File")
+                .field("path", path)
+                .field("mode", mode)
+                .field("content", &"...") // Omit content
```
tvsfx: This is OK, but you can use the derivative to do this concisely (works for enums afaik):

```rust
fn redact<T>(_t: &T, f: &mut fmt::Formatter<'_>) -> fmt::Result {
    write!(f, "...")
}

#[derive(Derivative)]
#[derivative(Debug)]
struct Foo {
    #[derivative(Debug(format_with = "redact"))]
    secret: String,

    visible: u32,
}
```

fortanix-vme/fortanix-vme-initramfs/src/fs_tree.rs
```diff
@@ -0,0 +1,287 @@
...
+                .finish(),
+        }
+    }
+}
+
+impl PartialEq for FsTreeEntry {
```
tvsfx: https://mcarton.github.io/rust-derivative/latest/cmp.html
same comment as above

fortanix-vme/fortanix-vme-initramfs/src/fs_tree.rs
```diff
@@ -0,0 +1,287 @@
...
+            FsTreeEntry::File {
+                path,
+                mode,
+                content,
+            } => {
+                let name = path
+                    .to_str()
+                    .ok_or(Error::PathError(path.display().to_string()))?;
+                let builder = NewcBuilder::new(name).uid(0).gid(0).mode(mode);
+                let buffer = Box::new(content);
```
tvsfx: the `content` is already boxed

fortanix-vme/fortanix-vme-initramfs/src/lib.rs
```diff
@@ -0,0 +1,270 @@
...
+pub use fs_tree::{FsTree, FsTreeBuilder, FsTreeEntry};
+
+const DEFAULT_UID: u32 = 0;
+const DEFAULT_GID: u32 = 0;
+
+pub struct Initramfs<R: Read>(R);
+
+#[derive(Error, Debug)]
+pub enum Error {
+    #[error("Can't extract data from initramfs: {0}")]
```
tvsfx: same comment as above: dont do `: {0}` if the error is in your source

fortanix-vme/fortanix-vme-initramfs/src/lib.rs
```diff
@@ -0,0 +1,270 @@
...
+        Self { fs_tree }
+    }
+
+    fn build_initramfs<U: Read + Write>(self, output: U) -> Result<U, Error> {
+        let mut inputs = Vec::with_capacity(self.fs_tree.0.len());
+        for entry in self.fs_tree.0 {
+            let input = entry.into_cpio_input()?;
+            inputs.push(input);
+        }
+        cpio::write_cpio(inputs.drain(..), output).map_err(Error::WriteError)
```
tvsfx: `write_cpio` expects an iterator, so there's no need to `collect` into a `Vec` first, just `map` over the entries
also, why do you need `drain(..)`?

fortanix-vme/fortanix-vme-initramfs/src/lib.rs
```diff
@@ -0,0 +1,270 @@
...
+pub struct Builder {
+    fs_tree: FsTree,
+}
+
+impl Builder {
+    pub fn new(fs_tree: FsTree) -> Builder {
+        Self { fs_tree }
+    }
+
+    fn build_initramfs<U: Read + Write>(self, output: U) -> Result<U, Error> {
```
tvsfx: is `Read` bound necessary on `U`? (here and elsewhere)

fortanix-vme/fortanix-vme-eif/src/lib.rs
```diff
@@ -216,63 +245,120 @@ impl<R: Read + Seek + 'static, S: Read + Seek + 'static, T: Read + Seek + 'stati
...
+            .add_directory("rootfs/dev")
+            .add_directory("rootfs/proc")
+            .add_directory("rootfs/run")
+            .add_directory("rootfs/sys")
+            .add_directory("rootfs/tmp")
+            .build()
+    }
+
+    pub fn build<
+        R: Read + Seek + 'static,
```
tvsfx: use `ReadSeek` abbreviation or remove `ReadSeek`?

fortanix-vme/fortanix-vme-initramfs/src/lib.rs
```diff
@@ -0,0 +1,270 @@
...
+    pub fn wrong_gid(found: u32, expected: u32) -> Self {
+        Error::WrongGid { found, expected }
+    }
+
+    pub fn wrong_mode(found: u32, expected: u32) -> Self {
+        Error::WrongMode { found, expected }
+    }
+}
+
+pub struct Builder {
```
tvsfx: There's too many builders in this repo :smile: 
Any reason to not define these methods on `InitRamFs` directly

fortanix-vme/fortanix-vme-initramfs/src/lib.rs
```diff
@@ -0,0 +1,270 @@
...
+                FsTreeEntry::File {
+                    path,
+                    mode,
+                    mut content,
+                } => {
+                    let path_str = path
+                        .as_path()
+                        .to_str()
+                        .ok_or(Error::PathError(path.display().to_string()))?;
+                    Initramfs::verify_entry(&reader, path_str, DEFAULT_UID, DEFAULT_GID, mode)?;
```
tvsfx, line 112: I would read from my `reader` here instead of passing it down, so you can separate checking from reading

fortanix-vme/fortanix-vme-initramfs/src/lib.rs
```diff
@@ -0,0 +1,270 @@
...
+impl<R: Read + Seek> ReadSeek for R {}
+
+impl<R: Read> From<R> for Initramfs<R> {
+    fn from(reader: R) -> Initramfs<R> {
+        Initramfs(reader)
+    }
+}
+
+impl<R: Read> Initramfs<R> {
+    pub fn verify(self, fs_tree: FsTree) -> Result<(), Error> {
```
tvsfx, line 97: this is only used in testing; does this need to be in the public API?

fortanix-vme/fortanix-vme-initramfs/src/lib.rs
```diff
@@ -0,0 +1,270 @@
...
+        assert_eq!(
+            sha256[..],
+            hex!("07b33a2c4abb76d839912f0dfeade8868cb79b28305c86fdfcf3646c5cf504c3")[..]
+        );
+
+        let path_bin_pairs = [("app", app0), ("init", init0), ("nsm", nsm0)];
+
+        // Parse initramfs and verify binaries in it.
+        for (path, expected) in path_bin_pairs {
+            let initramfs = Initramfs::from(Cursor::new(initramfs.clone()));
```
tvsfx: why do we need to wrap initramfs inside an `Initramfs`?

fortanix-vme/fortanix-vme-eif/src/initramfs.rs
```diff
@@ -1,223 +0,0 @@
...
-            is_executable(reader, "./nsm.ko")
-        }
-
-        fn next<R: Read>(reader: NewcReader<R>) -> Result<NewcReader<R>, Error> {
-            reader.finish()
-                .and_then(|r| NewcReader::new(r))
-                .map_err(|e| Error::InitramfsWriteError(e))
-        }
-
-        let decoder = GzDecoder::new(self.0);
```
tvsfx, line 145: I think your approach of separating the FS representation from the checking logic is much better than the old implementation :)

fortanix-vme/fortanix-vme-initramfs/src/lib.rs
```diff
@@ -0,0 +1,270 @@
...
+        }
+
+        if !reader.entry().is_trailer() {
+            return Err(Error::ExpectedTrailer);
+        }
+
+        Ok(())
+    }
+
+    pub fn find_entry_by_path(self, path: &str) -> Result<Vec<u8>, Error> {
```
tvsfx: naming: this actually reads the _contents_ of the entry

fortanix-vme/fortanix-vme-eif/src/lib.rs
```diff
@@ -184,26 +195,44 @@ impl<T: Read + Seek> FtxEif<T> {
             .ok_or(Error::EifParseError(String::from("No ramdisks found")))?;
 
         let initramfs = Initramfs::from(Cursor::new(initramfs.deref()));
-        let mut app = Vec::new();
-        initramfs.application(&mut app)?;
```
tvsfx, line 188: The old behavior did both `verify` and `find_entry_by_path`; the new behavior only does the latter - intentional?

fortanix-vme/fortanix-vme-initramfs/src/lib.rs
```diff
@@ -0,0 +1,270 @@
...
+
+#[cfg(test)]
+mod tests {
+    use super::{Builder, Initramfs, fs_tree::FsTreeBuilder};
+    use hex_literal::hex;
+    use sha2::{Digest, Sha256};
+    use std::io::Cursor;
+
+    #[test]
+    fn create_and_parse_initramfs() {
```
tvsfx, line 225: add test for `verify` here?

fortanix-vme/fortanix-vme-eif/src/lib.rs
sardok: Yes, i didn't want to mix `find_entry_by_path` into `verify` logic. Since these both functions require ownership of underlying initramfs data, `Clone` here would be needed to keep the old behaviour at least in this function. I am not sure, do you think is `verify` necessary here?  `verify` is already execercised by the unit tests and i think it should be enough.

fortanix-vme/fortanix-vme-eif/src/lib.rs
sardok: I think the rationale behind this was, Eif `Builder` would take binaries/files only to make a complete image and initramfs is one of the parts of the resulting image.

fortanix-vme/fortanix-vme-eif/src/lib.rs
sardok: Nitro's `init` read to content of this file to determine which command to execute after `chroot` to `rootfs` (see it [here](https://github.com/aws/aws-nitro-enclaves-sdk-bootstrap/blob/f718dea60a9d9bb8b8682fd852ad793912f3c5db/init/init.c#L413)). So this is a fixed pointer to the application. Yes, i assume passing an application name here would make better sense probably.

fortanix-vme/fortanix-vme-initramfs/src/lib.rs
sardok: Yes i thought about that too. I guess there is no harm to give this functionality to the non-test code just in case that might be needed later.

fortanix-vme/fortanix-vme-initramfs/src/lib.rs
sardok: Well i couldn't get rid of `Vec` & `collect`. Because `entry.into_cpio_input()` returns `Result`, so i guess there is no way to convert from `Vec<Result<..>>` to `Result<Vec<_>>` without using `collect`. If you do then explicit type info is required.
I got rid of `drain(..)` though.

fortanix-vme/fortanix-vme-initramfs/src/lib.rs
sardok: Removed from `fn build_initramfs`, kept at `fn build` because of `pub struct Initramfs<R: Read>(R)`.

fortanix-vme/fortanix-vme-initramfs/src/lib.rs
sardok: Renamed.

fortanix-vme/fortanix-vme-initramfs/src/lib.rs
sardok: Added.

fortanix-vme/fortanix-vme-initramfs/src/lib.rs
sardok: Its no longer `Initramfs` its `Vec` at that point. Letme rename it.

fortanix-vme/fortanix-vme-initramfs/src/lib.rs
sardok: Removed `Builder` here and added `Initramfs::from_fs_tree`.

fortanix-vme/fortanix-vme-initramfs/src/lib.rs
sardok: Done.

fortanix-vme/fortanix-vme-initramfs/src/fs_tree.rs
sardok: Removed.

fortanix-vme/fortanix-vme-initramfs/src/fs_tree.rs
sardok: Added for `Debug` & `PartialEq`. There is clippy warning now though. Postponing the warning for now.

fortanix-vme/fortanix-vme-initramfs/src/fs_tree.rs
sardok: Yes, `ancestors` end up returning `"", "."` .

fortanix-vme/fortanix-vme-eif/src/lib.rs
sardok: Its in initramfs crate. Maybe i should get rid of it from that crate.

fortanix-vme/fortanix-vme-eif/src/lib.rs
sardok: Done.

fortanix-vme/fortanix-vme-eif/src/error.rs
sardok: Done.

fortanix-vme/fortanix-vme-eif/src/error.rs
sardok: Done.

fortanix-vme/fortanix-vme-initramfs/src/fs_tree.rs
tvsfx: https://doc.rust-lang.org/std/path/struct.Path.html#method.ancestors
For normalized paths it will end on `/` if I am to believe the docs.

fortanix-vme/fortanix-vme-eif/src/lib.rs
tvsfx: this makes me wonder; why does `verify` need to take ownership of the reader; couldnt you pass it a `&mut Cursor<Reader>` so that you can rewind later?

I did not look into who we expect to call the `fn application` function that Raoul had written; it is `pub` so I just wanted to make sure that no clients relied upon this `verify` call.

fortanix-vme/fortanix-vme-initramfs/src/lib.rs
tvsfx: Ah yes, I missed the `?`, my bad.

fortanix-vme/fortanix-vme-eif/src/lib.rs
sardok: > this makes me wonder; why does verify need to take ownership of the reader; couldnt you pass it a &mut Cursor<Reader> so that you can rewind later?

Yes probably it makes sense to rewind to the pos got by `Cursor::position` at the beginnig of the function.

fortanix-vme/fortanix-vme-eif/Cargo.toml
sardok: Done.

fortanix-vme/fortanix-vme-initramfs/src/lib.rs
tvsfx: I meant that it makes more sense to not shadow `initramfs` above; 
`        let initramfs: Vec<u8> = initramfs.into_inner().into_inner();`
can you not just `.clone()` the cursor or initramfs and `let initramfs_data = ...` or some other name

fortanix-vme/fortanix-vme-initramfs/src/lib.rs
tvsfx: note: you used a different parameter `U` in `from_fs_tree` - you should add the `Write` bound to the existing parameter `R`  using a `where` clause or a more precise `impl` block

fortanix-vme/fortanix-vme-initramfs/src/lib.rs
```diff
@@ -0,0 +1,268 @@
...
+impl<R: Read + Seek> ReadSeek for R {}
+
+impl<R: Read> From<R> for Initramfs<R> {
+    fn from(reader: R) -> Initramfs<R> {
+        Initramfs(reader)
+    }
+}
+
+impl<R: Read> Initramfs<R> {
+    pub fn from_fs_tree<U: Read + Write>(
```
tvsfx: Related to the discussion of reusing readers/writers above; this API is a little odd to me. I'd prefer just passing in a mutable reference to a reader/writer.

Right now I have the following in my code:
```
    let initramfs = initramfs::build(application_elf, initramfs, initramfs)
```
Whereas `initramfs` could just be written in-place

fortanix-vme/fortanix-vme-initramfs/src/fs_tree.rs
sardok: The following;
```
    let p4 = Path::new("./a/b/c/d");
    let p5 = Path::new("a/b/c/d");
    let p4_a = p4.ancestors().collect::<Vec<_>>();
    let p5_a = p5.ancestors().collect::<Vec<_>>();
    println!("p4 ancestors: {:?}", p4_a);
    println!("p5 ancestors: {:?}", p5_a);
```
Results to;
```
p4 ancestors: ["./a/b/c/d", "./a/b/c", "./a/b", "./a", ".", ""]
p5 ancestors: ["a/b/c/d", "a/b/c", "a/b", "a", ""]
```
So there are `""` & `"."` to get rid of. For formalized, yes `"."` is gone but in this case we need to handle  `"/"` (i guess even in current state of code (no normalization), i need to handle `"/"`.
Here are some normalization experiments;
```
    let p1 = Path::new("/a/b/c/d").normalize();
    let p2 = Path::new("a/b/../c/../d").normalize();
    let p3 = Path::new("./a/b/../c/d/e").normalize();
    println!("p1: {:?}", p1);
    println!("p2: {:?}", p2);
    println!("p3: {:?}", p3);
    let p1_a = p1.ancestors().collect::<Vec<_>>();
    let p2_a = p2.ancestors().collect::<Vec<_>>();
    let p3_a = p3.ancestors().collect::<Vec<_>>();
    println!("p1 ancestors: {:?}", p1_a);
    println!("p2 ancestors: {:?}", p2_a);
    println!("p3 ancestors: {:?}", p3_a);
```
Results to;
```
p1: "/a/b/c/d"
p2: "a/d"
p3: "a/c/d/e"
p1 ancestors: ["/a/b/c/d", "/a/b/c", "/a/b", "/a", "/"]
p2 ancestors: ["a/d", "a", ""]
p3 ancestors: ["a/c/d/e", "a/c/d", "a/c", "a", ""]
```

fortanix-vme/fortanix-vme-initramfs/src/lib.rs
sardok: I avoided shadowing `initramfs`, renamed that to `initramfs_blob`. But the original `initramfs` doesn't have much value. Because `Cursor::clone` clone the position as well. So i guess its better to have explicit `let initramfs = Initramfs::from(Cursor::new(initramfs.clone()));` in for loop.

tvsfx: You can access private members in tests; you could call  `rewind()`

fortanix-vme/fortanix-vme-eif/src/lib.rs
sardok: Ok, thanks to @tvsfx `&mut Cursor<Reader>` & rewinding allow us to do `verify()` and `application()` calls in a row. But unfortunately, `FtxEif::application` can not run `verify` here due to missing `FsTree` structure which is needed by `verify`.
 @raoulstrackx Now `FtxEif::application()` does no longer implicitly verify other files as it used to. There is separate function called `verify()` and it is used in unit tests. Are you ok with this change?

fortanix-vme/fortanix-vme-initramfs/src/lib.rs
sardok: Good idea, but for now i would prefer not to do it, if you think it is not that important?

fortanix-vme/fortanix-vme-eif/src/lib.rs
raoulstrackx: I think that's fine.

fortanix-vme/fortanix-vme-initramfs/src/lib.rs
```diff
@@ -0,0 +1,267 @@
...
+        mode: u32,
+    ) -> Result<(), Error> {
+        let entry = reader.entry();
+        if entry.name() != path {
+            return Err(Error::wrong_entry_name(
+                entry.name().to_string(),
+                path.to_string(),
+            ));
+        }
+        if entry.uid() != 0 {
```
raoulstrackx: Shouldn't this be `entry.uid() != uid`?

fortanix-vme/fortanix-vme-initramfs/src/lib.rs
```diff
@@ -0,0 +1,267 @@
...
+            return Err(Error::wrong_uid(entry.uid(), uid));
+        }
+        if entry.gid() != 0 {
```
raoulstrackx: Shouldn't this be `entry.gid() != gid`?

fortanix-vme/fortanix-vme-initramfs/src/lib.rs
tvsfx: Ok, let's leave it.

fortanix-vme/fortanix-vme-initramfs/src/lib.rs
sardok: Fixed.

fortanix-vme/fortanix-vme-initramfs/src/lib.rs
sardok: Fixed.

fortanix-vme/fortanix-vme-initramfs/src/lib.rs
sardok: Updated.

fortanix-vme/fortanix-vme-initramfs/src/fs_tree.rs
```diff
@@ -0,0 +1,278 @@
...
+pub enum FsTreeEntry {
+    Directory {
+        path: PathBuf,
+        mode: u32,
+    },
+    File {
+        path: PathBuf,
+        mode: u32,
+        #[derivative(Debug(format_with = "redact"))]
+        #[derivative(PartialEq = "ignore")]
```
tvsfx: Now that I look at it again, this `PartialEq` implementation is probably surprising to users; you don't expect `FSTreeEntry`s or `FsTree`s with different contents to be equal.

This should probably be a static method on the `FsTreeEntry`/`FsTree` type called `eq_metadata` instead.

sardok: Yes, i agree. This bothered me as well. Let me add an explicit method for this kind of comparison.

PR #861, opened 2026-01-23
Add platform-agnostic enclave-runner abstraction

jethrogb:
This PR adds a platform-agnostic enclave-runner abstraction, to be able to use different enclave platforms in a similar way from the untrusted host.

<ins>Crate renames</ins>
    * The original `enclave-runner` crate is now renamed to `enclave-runner-sgx`.
    * A new crate `enclave-runner` has been created at version 0.8. This version change allows completely changing the API per semver.

<ins>New API for users</ins>
The specific platform will have a loader/builder type for each enclave type (Command or Library) it supports. You need to create that and set any platform-specific settings, then pass it to `enclave_runner::EnclaveBuilder::new`. Common builder settings need to be set on `enclave_runner::EnclaveBuilder`.

The `UsercallExtensions` API has been changed to `StreamRouter`. It basically the same, except that **the type that implements the trait is responsible for any fallback** to OS defaults (usually by forwarding to the `OsStreamRouter` implementation).

All **users of `enclave-runner` 0.7 and lower need to be updated** to use the new crate names and APIs. Examples of such changes are in `intel-sgx/fortanix-sgx-tools/src/bin/ftxsgx-runner.rs`, `intel-sgx/report-test/src/lib.rs`, `intel-sgx/sgxs-tools/src/sgx_detect/tests/mod.rs`

<ins>New API for platform developers</ins>
Platform developers need to implement the API in `enclave_runner::platform`. The main thing to do is to implement `EnclavePlatform` for each supported enclave type

```rust
impl EnclavePlatform<Command> for MyPlatform {
    ...
}

// if supported:
impl EnclavePlatform<Library> for MyPlatform {
    ...
}
```

### DIFF
--- a/intel-sgx/enclave-runner-sgx/src/usercalls/mod.rs
+++ b/intel-sgx/enclave-runner-sgx/src/usercalls/mod.rs
@@ -7,7 +7,7 @@
 use std::alloc::{GlobalAlloc, Layout, System};
 use std::cell::RefCell;
 use std::collections::{HashMap, VecDeque};
-use std::io::{self, ErrorKind as IoErrorKind, Read, Result as IoResult};
+use std::io::{self, ErrorKind as IoErrorKind, Result as IoResult};
 use std::pin::Pin;
 use std::ptr;
 use std::result::Result as StdResult;
@@ -19,6 +19,8 @@ use std::time::{self, Duration};
 use std::{cmp, fmt, str};
 
 use anyhow::bail;
+use enclave_runner::platform::CommandConfiguration;
+use enclave_runner::stream_router::{AsyncListener, AsyncStream, StreamRouter};
 use fnv::FnvHashMap;
 use fortanix_sgx_abi::*;
 use futures::future::{poll_fn, Either, Future, FutureExt};
@@ -32,7 +34,7 @@ use libc::*;
 #[cfg(unix)]
 use nix::sys::signal;
 use sgxs::loader::Tcs as SgxsTcs;
-use tokio::io::{AsyncRead, AsyncWrite, ReadBuf};
+use tokio::io::ReadBuf;
 use tokio::runtime::Builder as RuntimeBuilder;
 use tokio::sync::broadcast::error::RecvError;
 use tokio::sync::{broadcast, mpsc as async_mpsc, oneshot, Semaphore};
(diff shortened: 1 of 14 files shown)


enclave-runner/src/builder.rs
```diff
@@ -0,0 +1,83 @@
...
+            stream_router: None,
+            forward_panics: false,
+            type_builder: Default::default(),
+        }
+    }
+
+    /// The stream router that this enclave should use when the enclave is
+    /// creating any streams. Defaults to [`OsStreamRouter`].
+    pub fn stream_router<R: Into<Box<dyn StreamRouter>>>(&mut self, router: R) {
+        self.stream_router = Some(router.into());
```
Taowyoo:
```suggestion
    pub fn stream_router<R: Into<Box<dyn StreamRouter>>>(&mut self, router: R) -> &mut Self {
        self.stream_router = Some(router.into());
        self
```
keep API consistency with  `fn forward_panics`

enclave-runner/src/stream_router/mod.rs
```diff
@@ -0,0 +1,99 @@
...
+
+mod os;
+pub use self::os::OsStreamRouter;
+
+pub trait AsyncStream: AsyncRead + AsyncWrite + 'static + Send + Sync {
+    fn poll_read_alloc(self: Pin<&mut Self>, cx: &mut Context) -> Poll<IoResult<Vec<u8>>>
+    {
+        let mut v: Vec<u8> = vec![0; 8192];
+        let mut buffer = ReadBuf::new(&mut v);
+        self.poll_read(cx, &mut buffer).map(|b| b.map(|_| buffer.filled().to_vec()))
```
Taowyoo: nit
```suggestion
        self.poll_read(cx, &mut buffer).map(|b| b.map(|()| buffer.filled().to_vec()))
```

Taowyoo (approved):
The PR itself is a great job.


But to enable `UsercallExt/StreamRouter` for vme runner. I found there is a big obstacle here: fortanix-vme-runner is writtern in no-async. Need some work to migrate it to async first.

enclave-runner/src/stream_router/mod.rs
jethrogb: Needs rewrite, this function allocates every time!

enclave-runner/src/stream_router/mod.rs
```diff
@@ -0,0 +1,114 @@
...
+use std::task::{Context, Poll};
+
+use tokio::io::{AsyncReadExt, AsyncRead, AsyncWrite};
+
+mod os;
+pub use self::os::OsStreamRouter;
+
+pub trait AsyncStream: AsyncRead + AsyncWrite + 'static + Send + Sync {
+    fn poll_read_alloc(mut self: Pin<&mut Self>, cx: &mut Context) -> Poll<IoResult<Vec<u8>>> {
+        fn empty_buf() -> Vec<u8> {
```
Taowyoo, line 21: nit: 
```suggestion
        #[inline(always)]
        fn empty_buf() -> Vec<u8> {
```
would be beneficial to inline this ?

intel-sgx/enclave-runner-sgx/src/usercalls/mod.rs
```diff
@@ -1458,89 +1248,41 @@ impl<'tcs> IOHandlerInput<'tcs> {
...
-            }
-            return Ok(self.alloc_fd(AsyncFileDesc::listener(stream_ext)).await);
-        }
 
-        let socket = tokio::net::TcpListener::bind(addr).await?;
-        if let Some(local_addr) = local_addr {
-            local_addr.set(socket.local_addr()?.to_string().into_bytes());
-        }
-        Ok(self.alloc_fd(AsyncFileDesc::listener(Box::new(socket))).await)
+        let socket = self.enclave.stream_router.bind_stream(addr, local_addr.map(OutputBuffer::init_default)).await?;
```
Taowyoo: nit: The `init_default` function calls here are little misleading: feel like they are over writing default value into 
given &mut reference.
could we rename it to `init_default_if_empty` ?

intel-sgx/enclave-runner-sgx/src/usercalls/mod.rs
```diff
@@ -1458,89 +1248,41 @@ impl<'tcs> IOHandlerInput<'tcs> {
...
         let file_desc = self.lookup_fd(fd).await?;
-        let stream = file_desc.as_listener()?.async_accept(local_addr_str.as_mut(), peer_addr_str.as_mut()).await?.unwrap();
 
-        if let Some(local_addr) = local_addr {
-            local_addr.set(&local_addr_str.unwrap().into_bytes()[..])
-        }
-        if let Some(peer_addr) = peer_addr {
-            peer_addr.set(&peer_addr_str.unwrap().into_bytes()[..])
-        }
+        let stream = file_desc.as_listener()?.async_accept(local_addr.map(OutputBuffer::init_default), peer_addr.map(OutputBuffer::init_default)).await?.unwrap();
```
Taowyoo: Hi @jethrogb , could you help explain when `unwrap()` is acceptable here?

intel-sgx/enclave-runner-sgx/src/usercalls/mod.rs
jethrogb: I don't really like the current name either. That said, I spent too much time thinking about the function name. I don't like making it longer since it makes the usage sites less readable.

What do you think about just always overwriting the contents? It's not like we ever put something there and then look at it again later.

enclave-runner/src/stream_router/mod.rs
jethrogb: The compiler should be able to figure this out. We shouldn't add inline annotations unless we have proof that it makes things better.

intel-sgx/enclave-runner-sgx/src/usercalls/mod.rs
jethrogb: Yeah this is bogus. I think this is caused by a flaw in the original UsercallExtension design. It allowed returning `None` from `connect` and `bind` to fallback to the standard implementation. The `Option` in the return signature of those functions was then inadvertently copied to `accept` as well, which I don't think makes sense: there's no reasonable fallback, and the existing implementation also doesn't have a fallback.

With StreamRouter, the implementer is responsible for any fallback so we can just get rid of the `Option` in the return type, which automatically gets rid of the unwrap.

Taowyoo: Got it

Taowyoo: Got it

PR #862, opened 2026-01-26
[RTE-710] elf2uki converter CLI tool

tvsfx:


fortanix-vme/fortanix-vme-initramfs/Cargo.toml
```diff
@@ -1,7 +1,7 @@
 [package]
 name = "fortanix-vme-initramfs"
 version = "0.1.0"
-edition = "2024"
+edition = "2021"
```
sardok, line 4: If you need this change i assume that you're not using nightly compiler but using our toolchain? This change is ok but i would like to add that, the `rust-sgx` repository is compiled/tested agains nightly toolchain by its CI.

fortanix-vme/tools/elf2uki/src/main.rs
```diff
@@ -0,0 +1,176 @@
...
+        Ok(ValidatedCli {
+            non_defaulted_args: self.non_defaulted_args,
+            output_path,
+        })
+    }
+}
+
+/// Run the `ukify` tool after validation of the passed-in arguments
+fn build_uki(cli: &ValidatedCli, initramfs_path: &Path) -> Result<()> {
+    let mut command = Command::new(UKIFY_PATH);
```
sardok: So the `README.md` installation steps require you to create a virtual env directory (`.venv`) to install the `ukify.py`. But here, there is a direct invocation of `UKIFY_PATH` which would fail i believe if user does not invoke `source .venv/bin/activate` before doing `cargo run`.
There some alternatives, IMO better than what we have;
1- install `ukify` via package manager. For example, we as a company use ubuntu and ubuntu has `systemd-ukify` package. So you don't need to think about installing this scripts dependencies (`pefile`). You can invoke `ukify` just like here.

2- create a small project in `build_init` and create a small project there with `uv` package manager and a wrapper shell script. The wrapper shell script would do something like the following;
```
#!/bin/bash

PARSED_ARGS=()
parse_args() {
  --linux) ..
  --initrd)..
 --stub)..
}

uv run --project . ./ukify.py "${PARSED_ARGS[@]}"
```

In general, `elf2uki` tool doesn't have to be a rust project. I think, if we add `cli` tool for initramfs crate, It could be python project or could even be a bash script.

fortanix-vme/tools/elf2uki/.gitignore
```diff
@@ -0,0 +1 @@
+target/
```
sardok: i don't think this is necessary, its alread listed in `<repo_root>/.gitignore`.

fortanix-vme/tools/elf2uki/src/main.rs
```diff
@@ -0,0 +1,176 @@
...
+    kernel_cmdline: Option<PathBuf>,
+}
+
+pub fn open_file(path: &Path) -> Result<File> {
+    File::open(path).with_context(|| format!("failed to open file at path {}", path.display()))
+}
+
+impl Cli {
+    /// Validate the provided values, filling in defaults where necessary
+    fn validate(self) -> Result<ValidatedCli> {
```
sardok, line 80: Nit wondering if you can do destructuring here such as `let ValidatedCli { output_path, non_defaulted_args } = self;`

fortanix-vme/tools/elf2uki/src/main.rs
```diff
@@ -0,0 +1,176 @@
...
+    )]
+    // This is a required argument because `ukify` uses '/usr/lib/systemd/boot/efi/linux{opts.efi_arch}.efi.stub' by default, but this is not behavior we want to fall back on
+    uefi_stub_path: PathBuf,
+
+    #[arg(
+        long = "cmdline",
+        help = "String to pass as the kernel command line",
+        value_name = "STRING"
+    )]
+    kernel_cmdline: Option<PathBuf>,
```
sardok: Shouldnt this be `Option<String>`?

sardok: Also if this is a required value, shouldn't we remove `Option<>` type?

fortanix-vme/tools/elf2uki/src/main.rs
```diff
@@ -0,0 +1,176 @@
...
+/// * blob artefacts are set up as specified by the blobs README (TODO: RTE-739)
+///
+/// the following invocation works:
+///
+/// ```sh
+/// elf2uki \
+/// --kernel /tmp/kernel \
+/// --app /tmp/application \
+/// --uefi-stub /usr/lib/systemd/boot/efi/linuxx64.efi.stub \
+/// --cmdline "console=ttyS0 earlyprintk=serial" \
```
sardok: I think we can make this value `console=ttyS0 ...` as default value, don't think that it would be changed often.

fortanix-vme/tools/elf2uki/src/main.rs
tvsfx: That makes sense given that I use `non_defaulted_args` quite a few times. fixed.

fortanix-vme/tools/elf2uki/src/main.rs
tvsfx: It's not required; you can pass nothing and then the `--cmdline` argument will not be passed to `ukify`. It did need to be `String`, indeed; fixed.

fortanix-vme/tools/elf2uki/src/main.rs
tvsfx: I'm not sure we always want a cmdline value; this value is specifically good for debugging, so not sure we should default to it? It's fine to pass nothing; I removed the `--cmdline` from the example to illustrate that.

fortanix-vme/tools/elf2uki/.gitignore
tvsfx: the repo root only ignored `/target` at the top-level; if you start a path with `/` it's anchored where the `.gitignore` is.
I didn't know why we did that so left it alone for now

fortanix-vme/tools/elf2uki/src/main.rs
tvsfx: Discussed offline with Sinan and we'll assume `ukify` is installed externally for now.

fortanix-vme/fortanix-vme-initramfs/Cargo.toml
tvsfx: Discussed this offline with Sinan; I argued that we probably shouldn't use `edition = "2024"` if our default toolchain does not support it, even though it's technically fine in this case.
@sardok : if you do want me to fix this I will, let me know

fortanix-vme/tools/confidential-vm-blobs/blobs/README.md
```diff
@@ -0,0 +1,17 @@
+<!-- TODO (RTE-739, RTE-728): clean up setup -->
+# Blobs
+
+These blobs have been vendored from the following locations
+
+## init 
+Tweaked the code to build the `init` executable from [here](https://github.com/aws/aws-nitro-enclaves-sdk-bootstrap/blob/f718dea60a9d9bb8b8682fd852ad793912f3c5db) and placed it under `build_init/`.
+`build.rs` uses the shell script `build_init/update_init.sh` to place the generated `init` artefact in this folder and keep it up-to-date.
```
Taowyoo: nit
```suggestion
`build.rs` uses the shell script `build_init/update_init.sh` to place the generated `init` artifact in this folder and keep it up-to-date.
```

fortanix-vme/tools/confidential-vm-blobs/blobs/build_init/init.c
Taowyoo, line 1: I noticed this file is different from https://github.com/aws/aws-nitro-enclaves-sdk-bootstrap/blob/f718dea60a9d9bb8b8682fd852ad793912f3c5db/init/init.c  ( this is mentioned in README.md )

And the changes are:
1. Remove function `init_nsm_driver`: initialization of Nitro Secure Module driver
2. Remove function `enclave_ready`: sending signal to nitro-cli that the enclave has started
3. Remove global vars for above two functions.

I suggest to clarify this in `README.md`

fortanix-vme/fortanix-vme-initramfs/Cargo.toml
sardok: Let's keep it.

fortanix-vme/tools/elf2uki/src/main.rs
```diff
@@ -0,0 +1,176 @@
...
+        anyhow::Error::new(e).context(match error_kind {
+            std::io::ErrorKind::NotFound => {
+                "`ukify` tool not found in PATH; make it available or install `systemd-ukify`"
+            }
+            _ => "spawning ukify process failed",
+        })
+    })?;
+    if !output.status.success() {
+        return Err(anyhow!(
+            "ukify exited with non-zero status code and stdout : {} \n\n sterr : {}",
```
sardok: Typo: `sterr` -> `stderr`

fortanix-vme/tools/elf2uki/src/main.rs
```diff
@@ -0,0 +1,176 @@
...
+
+    let application_elf = open_file(&validated_args.non_defaulted_args.application_elf_path)?;
+    let init = Cursor::new(INIT);
+
+    // Unfortunately `ukify` forces us to have data in files.
+    let mut initramfs_file = NamedTempFile::new().context("failed to create initramfs file")?;
+    initramfs_file = initramfs::build(application_elf, init, initramfs_file)
+        .context("failed to create initramfs")?;
+
+    build_uki(&validated_args, initramfs_file.path())?;
```
sardok, line 169: Nit: if the process exits with an error code or crashes the created temporary files are not deleted : https://docs.rs/tempfile/latest/tempfile/struct.NamedTempFile.html#resource-leaking.
This behaviour is beneficial if you have a failing ukify command and you want to give it a try manually.

fortanix-vme/tools/confidential-vm-blobs/blobs/build_init/init.c
tvsfx: Added your clarification to the README, along with a shell command to generate the diff

fortanix-vme/tools/confidential-vm-blobs/blobs/README.md
tvsfx: argh, the British spelling gets me every time :)

fortanix-vme/tools/elf2uki/src/main.rs
tvsfx: That's true, but I dont call `exit(n)` or similar in my code, nor do I expect anyone to run the binary with `panic = 'abort'`, so I don't think a scenario where the `\tmp` files are not cleaned up will easily occur. I guess the main risk would be someone e.g. using `Ctrl-C` in the terminal. We could mitigate this by catching that signal (and potentially others), but that does seems like a pretty niche scenario

tvsfx: > This behaviour is beneficial if you have a failing ukify command and you want to give it a try manually.

If this turns out to be a problem, we can always make a `persist` flag or similar in the future, so that we don't delete the tmp files and log where they are.

fortanix-vme/tools/elf2uki/.gitignore
tvsfx: I removed this `.gitignore` again, because I added `elf2uki` to the top-level workspace, so it doesn't have its own `target` folder anymore

Taowyoo (approved):

Taowyoo (approved):

PR #863, opened 2026-01-30
[RTE-709][RTE-751] amd-sev runner + cargo runner

tvsfx:


fortanix-vme/fortanix-vme-tools/src/bin/ftxvme-runner-cargo.rs
```diff
@@ -152,107 +89,195 @@ struct CargoTomlMetadata {
...
+    memory: Option<u64>,
 
-    /// `false` by default. This enables debug mode of `nitro-cli run-enclave`.
+    // TODO: unused - remove this? Then we don't need to make `FortanixVmeConfig` generic
+    /// This enables debug mode of `nitro-cli run-enclave`, `false` by default.
     debug_mode: bool,
 }
 
 impl FortanixVmeConfig {
-    const DEFAULT_CPU_COUNT: isize = 2;
```
tvsfx, line 179: Note: defaults removed here because I set them in the runner. Made more sense to me that way, and avoids `Option`s or duplicate defaults.

fortanix-vme/fortanix-vme-runner/src/platforms/amdsevsnp.rs
```diff
@@ -0,0 +1,122 @@
...
+    let VmRunArgs {
+        uki_path,
+        firmware_image_path,
+        memory_mib,
+        cpu_count,
+    } = vm_run_args;
+    let memory_size = format!("{}M", memory_mib);
+
+    // TODO (RTE-740): id-block
+    let mut command = Command::new("sudo"); // TODO: look at "sudo" - needed for amd + `/dev/sev`?
```
Taowyoo: when running in stimulator, we should not need sudo right ?

fortanix-vme/fortanix-vme-runner/src/main.rs
```diff
@@ -1,51 +1,104 @@
...
+}
+
+#[derive(Clone, Debug, Args)]
+struct AmdSevSnpArgs {
+    /// Path to the firmware image file, defaulting to the relevant vendored image if not provided
+    #[arg(long = "firmware-image")]
+    firmware_image_path: Option<PathBuf>,
+
+    /// Name for the enclave in the runner
+    #[arg(long, default_value = "FortanixVm")]
```
Taowyoo: Nit: could we make the default name more informative by including `AmdSev`?

fortanix-vme/fortanix-vme-tools/src/bin/ftxvme-runner-cargo.rs
```diff
@@ -1,148 +1,82 @@
...
 
+    #[arg(trailing_var_arg = true)]
     others: Vec<String>,
 }
 
-impl Cli {
-    pub fn help() -> String {
-        String::from("Usage: <ftxvme-runner-cargo> [--simulate] [--verbose] <elf_path> [others]*")
+impl CargoArgs {
+    pub fn eif_path(&self) -> PathBuf {
```
Taowyoo, line 62: typo?

`eif` -> `elf`

fortanix-vme/fortanix-vme-tools/src/bin/ftxvme-runner-cargo.rs
```diff
@@ -1,148 +1,82 @@
...
-        let mut cli = env::args();
-        cli.next(); // Skip executable name
-        let cli = cli.fold(Ok(Cli::default()), |cli, arg| { cli.and_then(|cli| parse_arg(cli, arg) ) })?;
-
-        if cli.elf_path == PathBuf::default() {
-            Err(IoError::new(IoErrorKind::InvalidInput, String::from("No elf path provided")))
-        } else {
-            Ok(cli)
-        }
+    pub fn uki_path(&self) -> PathBuf {
```
Taowyoo, line 68: why this function code is identical to `eif_path`

fortanix-vme/fortanix-vme-tools/src/bin/ftxvme-runner-cargo.rs
```diff
@@ -152,107 +86,194 @@ struct CargoTomlMetadata {
...
 
-    let ftxvme_elf2eif = command! {
-        "ftxvme-elf2eif"   => args(
-        "--elf-path"       => cli.elf_path.clone(),
-        "--output-path"    => cli.eif_path()
-        )
-    };
+    let uki_path = amd_sev_snp_args.uki_path();
 
+    // TODO: we can assume this is installed right?
```
Taowyoo: I agree, we can add code here to check if this tool exist, if not pop up error msg and guide user the way to install it

fortanix-vme/fortanix-vme-tools/src/bin/ftxvme-runner-cargo.rs
```diff
@@ -152,107 +86,194 @@ struct CargoTomlMetadata {
...
+    let mut ftxvme_elf2uki = Command::new("ftxvme-elf2uki");
+    ftxvme_elf2uki
+        .arg("--app")
+        .arg(&amd_sev_snp_args.elf_path)
```
Taowyoo, line 171: as I mentioned above, with code above, `amd_sev_snp_args.elf_path` will be same to `uki_path` 

are you intential to just inplace generate the uki?

fortanix-vme/fortanix-vme-tools/src/bin/ftxvme-runner-cargo.rs
```diff
@@ -152,107 +86,194 @@ struct CargoTomlMetadata {
...
-    let ftxvme_elf2eif = command! {
-        "ftxvme-elf2eif"   => args(
-        "--elf-path"       => cli.elf_path.clone(),
-        "--output-path"    => cli.eif_path()
-        )
-    };
+    let uki_path = amd_sev_snp_args.uki_path();
 
+    // TODO: we can assume this is installed right?
+    let mut ftxvme_elf2uki = Command::new("ftxvme-elf2uki");
```
sardok: Sorry, i missed that but, we're not usigx `ftx*` prefix anymore, so this should be `fortanix-vme-elf2uki`.

fortanix-vme/fortanix-vme-runner/src/main.rs
```diff
@@ -57,20 +110,117 @@ impl TryFrom<&Cli> for NitroArgs {
...
+
+        run_to_completion::<NitroEnclaves>(
+            run_args,
+            metadata.img_name,
+            nitro_cli.aws_nitro_args.enclave_args,
+        )
+    }
+}
+
+fn run_to_completion<P: Platform + 'static>(
```
sardok: I think, it would be nice if `run_to_completion` & `create_runner` functions (maybe everything with no relation to `Cli` related types) move to `fortanix-vme-runner` crate (aka lib). They will be needed in common runner in the mono repo.

fortanix-vme/fortanix-vme-tools/src/bin/ftxvme-runner-cargo.rs
```diff
@@ -152,107 +86,194 @@ struct CargoTomlMetadata {
...
-#[derive(Serialize, Deserialize, Debug)]
+#[derive(Serialize, Deserialize, Debug, Default)]
 #[serde(rename_all = "kebab-case", default, deny_unknown_fields)]
 #[rustfmt::skip] // contains long lines because of links and they may wrapped by mistake
 /// This config is mainly intended for args of ftxvme-elf2eif and nitro-cli run-enclave.
 /// See their args documentation for more info about these opts:
 /// https://docs.aws.amazon.com/enclaves/latest/user/cmd-nitro-run-enclave.html#cmd-nitro-run-enclave-options
 /// https://docs.aws.amazon.com/enclaves/latest/user/cmd-nitro-build-enclave.html#cmd-nitro-build-enclave-options
 struct FortanixVmeConfig {
+    // TODO: unused - remove this? Then we don't need to make `FortanixVmeConfig` generic
```
sardok: This needs to be passed to `fortanix-vme-runner`. It is used by nitro, we may discard it for amd sev/snp.

fortanix-vme/fortanix-vme-runner/src/platforms/amdsevsnp.rs
sardok: Yes, i think it should be removed. Regarding `/dev/sev*`, for Amd VM, the devices should be available for the application running in qemu. For simulation, we can't guarantee their existence.

fortanix-vme/fortanix-vme-runner/src/platforms/amdsevsnp.rs
```diff
@@ -0,0 +1,122 @@
...
+    command.arg("-bios").arg(firmware_image_path);
+
+    if let RunMode::AmdSevVm = run_mode {
+        command
+            .arg("-machine")
+            .arg("confidential-guest-support=sev0");
+        command
+            .arg("-object")
+            .arg("sev-snp-guest,id=sev0,cbitpos=51,reduced-phys-bits=1,kernel-hashes=on");
+    }
```
sardok, line 174: When i run amd sev/snp simulated mode, all my terminal is cluttered with messages coming from qemu. I think we enable dumping the qemu message while `--verbose` is passed. Otherwise, we may ignore them via Command's `.stdout(Stdio::null())` or redirect to a file, what do you think?

fortanix-vme/fortanix-vme-runner/src/platforms/amdsevsnp.rs
tvsfx: We indeed don't need sudo; resolved this `TODO`

fortanix-vme/fortanix-vme-runner/src/main.rs
tvsfx: I didn't do that initially because it might be misleading in `simulator` mode, but since this arg is placed under `AmdSevSnpArgs` I'll add it.

fortanix-vme/fortanix-vme-tools/src/bin/ftxvme-runner-cargo.rs
tvsfx: good catch; that was already present in the old runner :P

fortanix-vme/fortanix-vme-tools/src/bin/ftxvme-runner-cargo.rs
tvsfx: it's not identical; the extension is different: uki files have extension `.efi`, and eif files have  `.eif`

fortanix-vme/fortanix-vme-tools/src/bin/ftxvme-runner-cargo.rs
tvsfx: Fixed for this one, discussed offline that we'd rename the other binaries later.

fortanix-vme/fortanix-vme-runner/src/main.rs
tvsfx: That's non-trivial to do, because `EnclaveRunner::new()` returns `ioError`, which cannot easily be converted to `VmeError`. We talked about it, and think there's a need for a new `Error` type that is different from `VmeError`.

fortanix-vme/fortanix-vme-tools/src/bin/ftxvme-runner-cargo.rs
tvsfx: improved error message handling in ftxvme cargo runner

fortanix-vme/fortanix-vme-tools/src/bin/ftxvme-runner-cargo.rs
tvsfx: see above: you misread the extension, right?

fortanix-vme/fortanix-vme-tools/src/bin/ftxvme-runner-cargo.rs
tvsfx: no what I meant is that `enclave_name` and `debug_mode` are ignored by `ftxvme-runner-cargo` _even_ in aws nitro mode - so should we just remove them, because in the old implementation they weren't doing anything either.

fortanix-vme/fortanix-vme-runner/src/platforms/amdsevsnp.rs
tvsfx: I tried doing that, but not getting any output from `qemu` was very annoying, since you cannot see what's going on in the console either. I investigated a bit more, and you can control the amount of logging the kernel does using the `loglevel` parameter on the kernel command line. I made that value scale with the verbosity instead (see `4d4b70a0989320d39794a7796deba257785618dd`) or [here](https://github.com/fortanix/rust-sgx/blob/4d4b70a0989320d39794a7796deba257785618dd/fortanix-vme/fortanix-vme-tools/src/bin/ftxvme-runner-cargo.rs?plain=1#L185-L186) and [here](https://github.com/fortanix/rust-sgx/blob/4d4b70a0989320d39794a7796deba257785618dd/fortanix-vme/fortanix-vme-tools/src/bin/ftxvme-runner-cargo.rs?plain=1#L208-L219). So now you still have some output, but if you don't specify extra `-v` arguments (note that you can specify multiple), it shouldn't be overwhelming

sardok: Ok, I think it is good enough now. its better to see everything at least till runner's stdio, stderr & stdin are wired into guest vm. We can think about more through approach later. Nitro for example dumps all VM output to log files.

fortanix-vme/fortanix-vme-runner/src/main.rs
sardok: Disccussed it offline.

fortanix-vme/fortanix-vme-tools/src/bin/ftxvme-runner-cargo.rs
tvsfx: removed these fields.

fortanix-vme/tools/confidential-vm-blobs/blobs/README.md
```diff
@@ -17,8 +17,11 @@ To see the diff:
 diff -c10 <(curl https://raw.githubusercontent.com/aws/aws-nitro-enclaves-sdk-bootstrap/f718dea60a9d9bb8b8682fd852ad793912f3c5db/init/init.c) blobs/build_init/init.c
 ```
 
-## vmlinuz-{version}
-Ubuntu kernel, extracted from noble package `linux-image-{version}`.
+## bzImage
+Upstream kernel, configured for the guest VM.
```
tvsfx: Can we mention that we started from an ubuntu kernel and baked in a vsock kernel module?

fortanix-vme/fortanix-vme-runner/doc/amd-sevsnp-host.md
```diff
@@ -0,0 +1,72 @@
...
+Expected output: `crw-rw---- 1 root kvm ... /dev/vhost-vsock`
+
+### 3. User Group Membership
+
+Ensure your user account is a member of the kvm group.
+
+```sh
+sudo usermod -aG kvm $USER
+```
+
```
tvsfx, line 71: TODO: Sinan: "we need to identify the things that require sudo on launchpad. Then add setup instructions for them in the readme doc."

fortanix-vme/fortanix-vme-runner/src/platforms/amdsevsnp.rs
```diff
@@ -20,18 +26,88 @@ pub struct VmRunArgs {
...
+        unsafe {
+            // We're deliberately omitting O_CLOEXEC here as we want
+            // the child process inherit the opened file descriptors.
+            let fd = open(VHOST_VSOCK_DEV, OFlag::O_RDWR, Mode::empty()).map_err(|e| {
+                map_nix_error(
+                    format!("Unable to open vhost-vsock device: {}", VHOST_VSOCK_DEV),
+                    e,
+                )
+            })?;
+            let res = set_guest_cid(fd.as_raw_fd(), &mut cid);
```
tvsfx: is this call to `as_raw_fd` necessary? `nix::open` returns a `RawFd` already

fortanix-vme/fortanix-vme-runner/src/platforms/amdsevsnp.rs
```diff
@@ -20,18 +26,88 @@ pub struct VmRunArgs {
...
+// In C: _IOW(VHOST_VIRTIO, 0x60, __u64)
+const VHOST_VIRTIO: u8 = 0xAF;
+const VHOST_VSOCK_DEV: &str = "/dev/vhost-vsock";
+nix::ioctl_write_ptr!(set_guest_cid, VHOST_VIRTIO, 0x60, u64);
+
+// This function basically opens vhost-vsock device and
+// tries to allocate a cid number. If allocation succeeds
+// we simply re-use it along with the file descriptor.
+fn get_available_guest_cid_with_fd() -> Result<(i32, u64), RunnerError> {
+    for mut cid in 3u64..=64 {
```
tvsfx: why is this the range you decide to check?

fortanix-vme/fortanix-vme-runner/src/platforms/amdsevsnp.rs
```diff
@@ -20,18 +26,88 @@ pub struct VmRunArgs {
...
+        unsafe {
```
tvsfx: try to make your `unsafe` block as small as possible

fortanix-vme/fortanix-vme-runner/src/platforms/amdsevsnp.rs
```diff
@@ -20,18 +26,88 @@ pub struct VmRunArgs {
...
+// Define the VHOST_VSOCK_SET_GUEST_CID ioctl
+// In C: _IOW(VHOST_VIRTIO, 0x60, __u64)
+const VHOST_VIRTIO: u8 = 0xAF;
+const VHOST_VSOCK_DEV: &str = "/dev/vhost-vsock";
+nix::ioctl_write_ptr!(set_guest_cid, VHOST_VIRTIO, 0x60, u64);
+
+// This function basically opens vhost-vsock device and
+// tries to allocate a cid number. If allocation succeeds
+// we simply re-use it along with the file descriptor.
+fn get_available_guest_cid_with_fd() -> Result<(i32, u64), RunnerError> {
```
tvsfx: nit: any reason not to use `c_int` for the `fd` return value? that's what `nix` uses

tvsfx: nit: return `VmRuntime` instance directly from this function?

fortanix-vme/fortanix-vme-runner/src/lib.rs
```diff
@@ -308,6 +308,9 @@ pub enum RunnerError {
     // TODO (RTE-770): refine error variants and return proper errors throughout crate
     #[error("io error occurred: {0:?}")]
     Io(Option<Cow<'static, str>>, #[source] io::Error),
+
+    #[error("runtime error occurred: {0}")]
+    Runtime(Cow<'static, str>),
```
tvsfx: unused - remove?

fortanix-vme/fortanix-vme-runner/src/platforms/amdsevsnp.rs
```diff
@@ -20,18 +26,88 @@ pub struct VmRunArgs {
...
-fn build_qemu_command(run_mode: RunMode, vm_run_args: VmRunArgs) -> Command {
+fn map_nix_error<I: Into<Cow<'static, str>>>(msg: I, err: Error) -> RunnerError {
+    RunnerError::Io(Some(msg.into()), err.into())
+}
+
+// Define the VHOST_VSOCK_SET_GUEST_CID ioctl
+// In C: _IOW(VHOST_VIRTIO, 0x60, __u64)
+const VHOST_VIRTIO: u8 = 0xAF;
+const VHOST_VSOCK_DEV: &str = "/dev/vhost-vsock";
+nix::ioctl_write_ptr!(set_guest_cid, VHOST_VIRTIO, 0x60, u64);
```
tvsfx, line 59: https://github.com/qemu/qemu/blob/28a6ca268c2cd3718b5bf24c2d97e3d1e95fc604/hw/virtio/vhost-vsock.c#L177
Not sure if you saw, but from this link I have the impression that QEMU sets the CID for you. Are you just using this to find out which `cid`s are unused? Seems like a pretty heavy-handed way to do that

fortanix-vme/fortanix-vme-runner/src/lib.rs
sardok: It's used in `get_available_guest_cid_with_fd` where if a cid value could not be found.

fortanix-vme/fortanix-vme-runner/src/platforms/amdsevsnp.rs
sardok: I thougt we maybe populate `VmRuntime` more with additional information aside from vsock related values (which handled by `get_available_guest_cid_with_fd`).

fortanix-vme/fortanix-vme-runner/src/platforms/amdsevsnp.rs
sardok: Yes QEMU sets the given cid as far as i can see. See it here: https://github.com/qemu/qemu/blob/28a6ca268c2cd3718b5bf24c2d97e3d1e95fc604/hw/virtio/vhost-vsock.c#L44. I assume `vsock->conf.guest_cid` is what we provide as qemu vsock parameters. The runner is just responsible for finding available cid via `set_guest_cid`. If we can find a way of having qemu find available `cid` value for the guest vm, we can remove that logic from the runner. I tried using some magic values such as `guest-cid=0` on qemu side but it didn't work.

fortanix-vme/tools/confidential-vm-blobs/blobs/README.md
sardok: Well, we'll tinker the kernel config more (see RTE-728). So are we going to mention all configs we apply? Not sure if this is a good idea. Thats why i think we keep the config here to give an idea about the customizations.

fortanix-vme/fortanix-vme-runner/src/platforms/amdsevsnp.rs
tvsfx: By "QEMU sets the CID" I meant that the call to `nix::ioctl_write_ptr!(set_guest_cid, VHOST_VIRTIO, 0x60, u64);` seems to be performed by QEMU, as you can see in the code I linked. If you follow the code I linked you end up [here](https://github.com/qemu/qemu/blob/28a6ca268c2cd3718b5bf24c2d97e3d1e95fc604/hw/virtio/vhost-backend.c#L271)

tvsfx: There seems to be no good way to have QEMU/KVM decide on a CID for us; we'll stick with the current impl of just looping to find one and setting it ourselves, to avoid TOCTOU issues.
For reference, libvirt does the same thing: https://github.com/libvirt/libvirt/blob/29aa5586425fd1910e986bb263adceee4829ffd0/src/util/virvsock.c#L82-L107

fortanix-vme/tools/confidential-vm-blobs/blobs/README.md
tvsfx: Ok, but it's impossible to say what these 13k lines of config mean without a reference. Can you do one of the following:
1. Point to the ubuntu config you started from (version X of package X, or the source)
2. Write down a `diff` command to display the diff
3. Describe the diff with the config you started from in words

We can still update this when we harden the kernel

fortanix-vme/fortanix-vme-runner/src/platforms/amdsevsnp.rs
tvsfx: TODO after discussion: 
1. check to the end of the range like libvirt does - no need to restrict ourselves to 64 VMs
2. create a `const` for the starting number

fortanix-vme/fortanix-vme-runner/src/platforms/amdsevsnp.rs
tvsfx: I would rename `VmRuntime` to `VsockConfig` or similar then, and just return that; we can still wrap it in a `VmRuntime` object if the need ever arises. I think it's nicer to return a struct with named fields than a tuple where we might misuse the fields.

fortanix-vme/fortanix-vme-runner/src/lib.rs
tvsfx: I would give it a more specific name then, and instead of taking `Cow<'static, str>` just bake the error message into the error variant. Some of the `Io` errors we currently create are also `Runtime` errors; its just a catch-all error type that will need to be refined later.

sardok: I think using `Io` error is not a good idea because there is no `Io` error, it can be misleading. Do you want me to change `Runtime` to something like `NoAvailableCidFound`?

fortanix-vme/fortanix-vme-runner/src/platforms/amdsevsnp.rs
```diff
@@ -0,0 +1,221 @@
...
+                if err_code != nix::Error::EADDRINUSE {
+                    return Err(map_nix_error(
+                        format!("Unable to ioctl vhost-vsock device: {}", VHOST_VSOCK_DEV),
+                        err_code,
+                    ));
+                }
+            }
+        }
+
+        cid = cid.checked_add(1).ok_or(RunnerError::NoAvailableCidFound)?;
```
tvsfx: Another reason to do 
```rust
let random_start: u32 = rand::thread_rng().random_range(CID_START..=u32::MAX);
for cid in (random_start..=u32::MAX).chain(CID_START..random_start) {...}
```
is that we wont need this `checked_add` and we can just throw 1 error after the loop

fortanix-vme/fortanix-vme-tools/src/bin/ftxvme-runner-cargo.rs
```diff
@@ -152,107 +90,243 @@ struct CargoTomlMetadata {
     fortanix_vme: FortanixVmeConfig,
 }
 
-#[derive(Serialize, Deserialize, Debug)]
+#[derive(Serialize, Deserialize, Debug, Default)]
 #[serde(rename_all = "kebab-case", default, deny_unknown_fields)]
 #[rustfmt::skip] // contains long lines because of links and they may wrapped by mistake
```
Taowyoo: After removed this line and run `cargo fmt` in crate root. There is no change around this struct.
I think we can get rid of it.

And for rustdoc, raw url is not suggested but `<https://xxx>`

fortanix-vme/fortanix-vme-runner/doc/amd-sevsnp-host.md
```diff
@@ -0,0 +1,72 @@
+Host Configuration for running Amd SEV/SNP VM
```
Taowyoo: nit: use level 1 header for title?

fortanix-vme/fortanix-vme-runner/src/platforms/amdsevsnp.rs
```diff
@@ -0,0 +1,221 @@
...
+}
+
+fn build_qemu_command(
+    run_mode: RunMode,
+    vm_run_args: VmRunArgs,
+    vsock_config: &VsockConfig,
+) -> Command {
+    const QEMU_EXECUTABLE: &str = "qemu-system-x86_64";
+    const QEMU_MACHINE: &str = "q35";
+    const AMD_PROCESSOR: &str = "EPYC-v4";
```
jethrogb, line 118: Are we choosing this as the base model for VME-SNP in perpetuity?

tvsfx: No, I hard-coded this for now since this is the model we had used for testing so far. Should this go through a design meeting/Is there a reason to make this a parameter right now?

fortanix-vme/tools/confidential-vm-blobs/blobs/README.md
sardok: Ok, blobs are gone now, so this should resolve this conversation. In future, we may add a README about how to get `fortanix-confidential-vm-blobs` deb package.

fortanix-vme/fortanix-vme-runner/Cargo.toml
```diff
@@ -9,16 +9,19 @@ authors = ["Fortanix, Inc."]
 fortanix-vme-abi = { path = "../fortanix-vme-abi", features = ["std"] }
 fortanix-vme-eif = { path = "../fortanix-vme-eif" }
 # External dependencies
+anyhow = "1.0"
 clap = { version = "4.0", features = ["derive"] }
+clap-verbosity-flag = "3.0.3"
+confidential-vm-blobs = { path = "../tools/confidential-vm-blobs/" }
```
Taowyoo: could you please move this to line 9 after `# Workspace dependencies`

fortanix-vme/fortanix-vme-runner/src/main.rs
```diff
@@ -1,54 +1,107 @@
...
+use std::io::{Error as IoError, Write};
 use std::os::unix::fs::OpenOptionsExt;
 use std::path::PathBuf;
 
+type DefaultLogLevel = WarnLevel;
+
 #[derive(Parser, Debug)]
-#[command(author, version, about, long_about = None)]
+#[command(author, version, about = "Run the given enclave image file", long_about = None)]
 struct Cli {
```
Taowyoo, line 20: One hight level question: are we fine to break backcompat for cli?

Taowyoo: Maybe you make nirto as default subcommand so roche side do not need to update existing CI script

Taowyoo (commented):
gave a pass:

more feedbacks are:
1. could we use VMADDR_CID_ANY to simplfy logic
2. Should we ensure cli of VME runner to be backcompat

fortanix-vme/fortanix-vme-tools/src/bin/ftxvme-runner-cargo.rs
Taowyoo: so could we remove this line?

```suggestion
```

fortanix-vme/fortanix-vme-runner/doc/amd-sevsnp-host.md
Taowyoo:
```suggestion
# Host Configuration for running Amd SEV/SNP VM
```

fortanix-vme/fortanix-vme-runner/src/platforms/amdsevsnp.rs
```diff
@@ -0,0 +1,220 @@
...
+        firmware_image_path,
+        memory_mib,
+        cpu_count,
+    } = vm_run_args;
+    let memory_size = format!("{}M", memory_mib);
+
+    // TODO (RTE-740): id-block
+    let mut command = match run_mode {
+        RunMode::AmdSevVm => {
+            let mut command = Command::new("sudo");
```
jethrogb: It's not acceptable to elevate privileges inside the runner.

tvsfx: fixed.

fortanix-vme/fortanix-vme-tools/src/bin/ftxvme-runner-cargo.rs
tvsfx: Removed the attribute and wrapped URLs in `<..>`

fortanix-vme/fortanix-vme-runner/src/main.rs
tvsfx: Note that breakage happens in both the `cargo-runner` and the regular `enclave-runner`. 

On a technical level, it will require quite a bit of refactoring to implement a default subcommand: you can make your command optional using `Option<Command>` but then you need to include the flattened arguments of your default subcommand again at the top-level, and make sure those can never be present together with a concrete subcommand.

Do you foresee problematic scenarios? Are there places where runner and backend are shipped separately? If the only issue is roche references I don't think its worth fixing this and we could just do a semver breaking change for the runner? If not I'll refactor.

cc: @raoulstrackx

sardok: About the mono repo side, we publish a `fortanix-vme-tools` package with toolchain and CI scripts uses a known version. And versions are append only. So when we make new package out of this changes, we update the CI too.

Taowyoo: It seems we always need to update CI. They I am fine with current change.
Thanks for detailed explanation here.

raoulstrackx: It's ok to not be backwards compatible for Nitro. Please discuss Fortanix private business on Slack

PR #868, opened 2026-02-04
[RTE-751]  Qemu vsock vhost

sardok:
This PR  vsock related configuration to the guest vm with qemu parameters. Although it is not possible to run two instances of fortanix-vme runner (due to port binding), this PR adds support for dynamic cid assignment.


fortanix-vme/fortanix-vme-runner/src/platforms/amdsevsnp.rs
sardok, line 126: I added a README to circumvent `root` perms for vsock. I guess we need to take similar steps for the rest (if there is any) and get rid of this `sudo`. cc @tvsfx

sardok:
Included in: https://github.com/fortanix/rust-sgx/pull/863.

PR #875, opened 2026-02-07
[RTE-711] VME runner: integrate enclave-runner StreamRouter + EnclavePlatform builder, refactor server

Taowyoo:
**Summary**
This change migrates `fortanix-vme-runner` to the `enclave-runner` API (builder + StreamRouter), refactors the command server/connection flow, and tightens error and address handling. It also updates `enclave-runner` trait bounds to be `Send + Sync` for routers and adds `Addr::port()` in the VME ABI.

**Key Changes**
1. Implemented `EnclavePlatform<Command>` via a new `EnclaveBuilder` in `fortanix-vme-runner`, replacing the old `EnclaveRunner` API.
2. Refactored server state to use `StreamRouter` for connect/bind/accept.
3. Added `RunnerError` variants (IO, VME ABI, Join, Nix, connection not found), plus better error conversion.
4. Switched `fortanix-vme-runner` CLI to non-async `main` and `enclave_runner::EnclaveBuilder` to run enclaves.
5. Updated `enclave-runner` stream router interfaces and default `OsStreamRouter` to be `Send + Sync`, and moved to `.boxed()` futures.

**Behavioral Notes / Migration**
- `EnclaveRunner` removed in favor of `EnclaveBuilder` (via `enclave_runner` traits). Callers should construct and run using `EnclaveBuilder`.
- `StreamRouter` must now be `Send + Sync` to satisfy new bounds.

**Testing**

Tested locally with simulate
```bash
CARGO_MANIFEST_DIR="$(pwd)" PATH="$(pwd):$PATH" ftxvme-runner-cargo --verbose --simulate amd-sev-snp ~/github_ws/std-unit-test
```

### DIFF
--- a/fortanix-vme/fortanix-vme-runner/src/lib.rs
+++ b/fortanix-vme/fortanix-vme-runner/src/lib.rs
@@ -801,7 +801,7 @@ async fn accept_stream(
     .await
 }
 
-/// An type that implements [`enclave_runner::EnclavePlatform<enclave_runner::Command>`]. So user
+/// An type that implements [`enclave_runner::platform::EnclavePlatform<enclave_runner::Command>`]. So user
 /// can use enclave_runner API to create vme enclaves.
 ///
 /// ```ignore


fortanix-vme/fortanix-vme-runner/src/lib.rs
```diff
@@ -658,3 +680,65 @@ impl<P: Platform + 'static> Server<P> {
...
+
+    fn build(
+        self,
+        _loader: Self::Loader,
+        configuration: EnclaveConfiguration,
+        mut cmd_configuration: CommandConfiguration
+    ) -> Result<enclave_runner::Command, anyhow::Error>
+    {
+        // cmd_args by default have an b"enclave" in it which is not needed
+        cmd_configuration.cmd_args.clear();
```
jethrogb: This is invalid.

Based on the comment, I think you mean `remove(0)`. But even that seems incorrect. `std::env::args().nth(0)` is always the process name, which we set to `"enclave"`.

fortanix-vme/fortanix-vme-runner/src/lib.rs
jethrogb: The API only allows running one enclave at a time. What's the point of keeping `servers: Vec<_>` and a separate `wait` function?

fortanix-vme/fortanix-vme-runner/src/lib.rs
```diff
@@ -658,3 +680,65 @@ impl<P: Platform + 'static> Server<P> {
...
+
+    pub struct Command {
+        _private: (),
+    }
+
+    impl Command {
+        pub(crate) fn internal_new<P: Platform + 'static, Args: Into<P::RunArgs> + 'static + Send>(
+            enclave_builder: EnclaveBuilder<P, Args>,
+            _stream_router: Box<dyn StreamRouter>,
+            _forward_panics: bool,
```
jethrogb: `forward_panics` should be implemented

fortanix-vme/fortanix-vme-runner/src/lib.rs
Taowyoo: Hi @jethrogb , you are right. But remove(0) should be valid as it keeps the current runner logic:
For vme runner, the original implementation use name of image as process name, see line 669:
```rust
            EnclaveState::Null => {
                enclave_args.insert(0, self.name.clone());
                // Assume Platform::run will do some blocking logic
                let handle = tokio::task::spawn_blocking(|| P::run(run_args));
```
and
https://github.com/fortanix/rust-sgx/blob/de27d22e986f70096573bfafbc6c3889a9cf5040/fortanix-vme/fortanix-vme-runner/src/main.rs#L145-L150

Do you suggest to change this behavior to hide image name to VM ?

sardok: > Based on the comment, I think you mean `remove(0)`. But even that seems incorrect. `std::env::args().nth(0)` is always the process name, which we set to `"enclave"`

Is this what we should do for the fortanixvme targets too? Unlike SGX, we're executing in unix environment and in unix environment `argv[0]` is process name while `argc > 0`. @tvsfx and I were trying to decide what to do for the sev/snp target. cc @jethrogb

fortanix-vme/fortanix-vme-runner/src/lib.rs
Taowyoo: added, please check

fortanix-vme/fortanix-vme-runner/src/lib.rs
Taowyoo: removed vec and unnecessary wait function

fortanix-vme/fortanix-vme-runner/src/lib.rs
```diff
@@ -641,15 +661,16 @@ impl<P: Platform + 'static> Server<P> {
...
         let mut state = self.state.enclave.write().await;
         match *state {
             EnclaveState::Running { .. } => panic!("Enclave already exists"),
             EnclaveState::Null => {
                 enclave_args.insert(0, self.name.clone());
                 // Assume Platform::run will do some blocking logic
                 let handle = tokio::task::spawn_blocking(|| P::run(run_args));
-                let enclave = handle.await.expect("Failed to run enclave")?;
+                let ret = handle.await?;
+                let enclave = ret?;
```
raoulstrackx: Nit: can't you just do `let = handle.await??`

fortanix-vme/fortanix-vme-runner/src/main.rs
```diff
@@ -1,12 +1,14 @@
+use anyhow::Context;
```
raoulstrackx: Can this application be renamed from `fortanix-vme-runner` to `enclave-runner-nitro`

Taowyoo: Em @raoulstrackx , are you sure? With @tvsfx PR https://github.com/fortanix/rust-sgx/pull/863 , this binary will be used for both nitro and amd sev.

fortanix-vme/fortanix-vme-runner/src/lib.rs
Taowyoo: done

fortanix-vme/fortanix-vme-runner/src/main.rs
raoulstrackx: Ah ok yes that's also an option. Then `fortanix-vme-runner` makes sense

fortanix-vme/fortanix-vme-runner/src/lib.rs
raoulstrackx: If I remember correctly, using a custom name will show up nicely when inspecting the running enclaves through the Nitro tools. Obviously, that's only a small advantage.
cc: @jethrogb This is the only comment blocking this PR. How strong is your opinion on this? For this PR it's limited to Nitro enclaves only.

jethrogb: @sardok The UNIX environment doesn't matter. We ignore what's passed in by the kernel and replace it with the args that we set here. See https://github.com/fortanix/rust/blob/a69f365d2349d101c677b52cf52f3831ca33e322/library/std/src/sys/unix/args.rs#L192

jethrogb: > using a custom name will show up nicely when inspecting the running enclaves through the Nitro tools

I don't understand what you mean by this. I don't see any relation to the code that's being discussed. What's at issue here is what an EDP program gets when calling `std::env::args().nth(0)`. In SGX this is `"enclave"`. Apparently, in VME-NE it's currently some external name. I don't think it's correct that different enclave platforms behave differently in this regard.

fortanix-vme/fortanix-vme-runner/src/lib.rs
```diff
@@ -89,22 +94,68 @@ impl StreamConnection for VsockStream {
...
+
+fn addr_from_string(err_msg_label: &str, addr: &str) -> Result<Addr, VmeError> {
+    if addr.is_empty() || addr == "error" {
+        error!("Missing {} address from stream router", err_msg_label);
+        return Err(VmeError::Command(VmeErrorKind::InvalidData));
+    }
+    let socket: SocketAddr = addr.parse().map_err(|err| {
+        error!("Failed to parse {} address '{}' : {}", err_msg_label, addr, err);
+        VmeError::Command(VmeErrorKind::InvalidInput)
+    })?;
```
jethrogb: It's not acceptable to error here. It's a bug in the current VME protocol that it can't properly represent all possible addresses that a StreamRouter may return.

#874 will address this. In the mean time, this function needs to always succeed.

fortanix-vme/fortanix-vme-runner/src/lib.rs
```diff
@@ -250,32 +302,29 @@ impl ClientConnection {
...
-            servers: Vec::new(),
-        })
+    pub fn new() -> Self {
+        Self { platform: PhantomData, stream_router: Arc::from(OsStreamRouter::new()) }
+    }
+
+    pub fn with_stream_router(&mut self, router: Box<dyn StreamRouter + Send + Sync>) -> &mut Self {
+        self.stream_router = Arc::from(router);
+        self
     }
```
jethrogb: This function is never used. Also, I don't think you need to convert to `SharedStreamRouter`, it's only used once (stored in `ServerState`).

Looking at it in more detail, I think the whole `EnclaveRunner` type is obsolete. Everything in `run_enclave` can be moved to `EnclaveBuilder`/`Command::internal_new`

fortanix-vme/fortanix-vme-runner/src/lib.rs
```diff
@@ -659,3 +752,68 @@ impl<P: Platform + 'static> Server<P> {
...
+        Ok(Self { runner, runner_args, enclave_name })
+    }
+}
+
+impl<P: Platform + 'static, Args: Into<P::RunArgs> + 'static + Send> EnclavePlatform<enclave_runner::Command> for EnclaveBuilder<P, Args> {
+    type Loader = ();
+
+    fn build(
+        self,
+        _loader: Self::Loader,
```
jethrogb, line 901: Is loader an SGX-specific concept? Should the EnclavePlatform API be changed so that the loader is passed in earlier, during the creation of `self`?

fortanix-vme/fortanix-vme-runner/src/lib.rs
sardok: > If I remember correctly, using a custom name will show up nicely when inspecting the running enclaves through the Nitro tools. Obviously, that's only a small advantage.

Yes AFAIK its called as image name in nitro lib. Thats fine, we can pass a nice image name to nitro lib while using `enclave` as `std::env::args().nth(0)`. So it should be 
`enclave_args.insert(0, "enclave");`.

fortanix-vme/fortanix-vme-runner/src/lib.rs
jethrogb: Thanks, looks good.

fortanix-vme/fortanix-vme-runner/src/lib.rs
Taowyoo: Got it, then here, I guess I can just assume it succeeded and unwrap here?

Taowyoo: Got it, then here, I guess I can just assume it succeeded and unwrap here?

jethrogb: Uh, no, panicking is the opposite of succeeding.

fortanix-vme/fortanix-vme-runner/src/lib.rs
Taowyoo: Removed EnclaveRunner and updated SharedStreamRouter to just
```
type BoxedStreamRouter = Box<dyn StreamRouter + Send + Sync>;
```

fortanix-vme/fortanix-vme-runner/src/lib.rs
Taowyoo: I think so. I cannot map sgx loader to anything in vme workflow,
For example, in SGX, loader in practice is `IsgxDevice`  represent device such as `/dev/isgx` ...
But in VME, different platform have different behavoir. For nitro, there is no such device.

In future, e.g. AMD SEV, `/dev/vsock*` is used but it is wrapped in platform implementation instead of exposed in runner level.

fortanix-vme/fortanix-vme-runner/src/lib.rs
Taowyoo: Updated code to return unspecific ip address `0.0.0.0:0` when given string cannot be parsed, please check

enclave-runner/src/stream_router/mod.rs
```diff
@@ -91,7 +91,7 @@ pub trait StreamRouter {
         addr: &'future str,
         local_addr: Option<&'future mut String>,
         peer_addr: Option<&'future mut String>,
-    ) -> std::pin::Pin<Box<dyn Future<Output = IoResult<Box<dyn AsyncStream>>> +'future>>;
+    ) -> std::pin::Pin<Box<dyn Future<Output = IoResult<Box<dyn AsyncStream>>> + Send +'future>>;
```
NicholasO-codes, line 94: (mega) nit: -> `Send +'future` --> `Send + 'future`

fortanix-vme/fortanix-vme-runner/src/main.rs
```diff
@@ -12,6 +12,8 @@ use std::fs::File;
 use std::io::{Error as IoError, Write};
 use std::os::unix::fs::OpenOptionsExt;
 use std::path::PathBuf;
+use fortanix_vme_runner::EnclaveBuilder as EnclaveBuilderVme;
```
NicholasO-codes: nits: 
- this `EnclaveBuilder as EnclaveBuilderVme` can be included in the imports on line 5
- double check the import sort order (I might be mistaken, but `enclave_runner` seems like it should be further to the top of the file).

fortanix-vme/fortanix-vme-runner/src/main.rs
```diff
@@ -238,17 +239,23 @@ async fn run_nitro_enclave(nitro_cli: AwsNitroCli) -> Result<()> {
...
-async fn run_to_completion<P: Platform + 'static>(
+fn run_to_completion<P: Platform + 'static>(
     run_args: P::RunArgs,
     enclave_name: String,
     enclave_args: Vec<String>,
 ) -> Result<(), anyhow::Error> where <P as Platform>::RunArgs : Send + Sync{
-    EnclaveRunner::<P>::run_to_completion(run_args, enclave_name, enclave_args).await
-        .context("failed to run enclave")
+    // EnclaveRunner::<P>::run_to_completion(run_args, enclave_name, enclave_args).await
+    //     .context("failed to run enclave")
```
NicholasO-codes: Should we remove these commented out lines?

fortanix-vme/fortanix-vme-runner/src/lib.rs
```diff
@@ -92,22 +117,65 @@ impl StreamConnection for VsockStream {
...
+#[derive(Clone, Copy, Debug)]
+struct ConnLogInfo {
+    local_port: u32,
+    peer_port: u32,
+    protocol: &'static str,
+}
+
+impl Display for ConnLogInfo {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.write_fmt(format_args!("[{} ({} -> {})]", self.protocol, self.local_port, self.peer_port))
```
s-arash: Nit: you could use `write!` instead.

fortanix-vme/fortanix-vme-runner/src/lib.rs
```diff
@@ -634,7 +619,11 @@ impl<P: Platform + 'static> ServerState<P> {
     }
 
     fn handle_request_exit(self: &Self, exit_code: i32) -> Result<(), VmeError> {
-        std::process::exit(exit_code);
+        if self.forward_panics && exit_code != 0 {
+            panic!("enclave panic with exit code: {}", exit_code);
```
NicholasO-codes: nit: "panic" --> "panicked"

fortanix-vme/fortanix-vme-runner/src/lib.rs
```diff
@@ -659,32 +648,55 @@ impl<P: Platform + 'static> ServerState<P> {
...
+    mut peer_addr: Option<&mut String>,
+) -> io::Result<Box<dyn AsyncStream>> {
+    poll_fn(|cx| {
+        let local = local_addr.as_mut().map(|addr| &mut **addr);
+        let peer = peer_addr.as_mut().map(|addr| &mut **addr);
+        listener.as_mut().poll_accept(cx, local, peer)
+    }).await
+}
+
+pub struct EnclaveBuilder<P: Platform, Args: Into<P::RunArgs>> {
```
NicholasO-codes, line 816: nit: Could we add some doc comments to this struct to provide some brief context on it's existence?

fortanix-vme/fortanix-vme-runner/src/lib.rs
```diff
@@ -698,30 +710,64 @@ impl<P: Platform + 'static> Server<P> {
...
                             }
-                        }; 
+                        };
                     });
                 }
             }
         );
-        Ok(handle)
+
+        let _encalve_descripter = tokio::task::spawn_blocking(|| P::run(runner_args)).await??;
```
NicholasO-codes: (mega nit?) typo: `_encalve_descripter` ---> `_enclave_descriptor`

fortanix-vme/fortanix-vme-runner/src/lib.rs
```diff
@@ -698,30 +710,64 @@ impl<P: Platform + 'static> Server<P> {
...
+        pub(crate) fn internal_new<P: Platform + 'static, Args: Into<P::RunArgs> + 'static + Send>(
+            enclave_builder: EnclaveBuilder<P, Args>,
+            stream_router: Box<dyn StreamRouter + Send + Sync>,
+            forward_panics: bool,
+            cmd_configuration: CommandConfiguration,
+        ) -> enclave_runner::Command {
+            (Box::new(move || -> Result<(), anyhow::Error> {
+                tokio::runtime::Builder::new_multi_thread()
+                    .enable_all()
+                    .build()?.
```
NicholasO-codes: nit: move the period to the next line so it's beside the `block_on(...)`

NicholasO-codes (dismissed):
I've noted a few things that should probably be checked, but I'm approving based on:
- the virtual code walk through we had together (explaining the changes, why it was needed, how it's integrating with the other pieces, how it's been tested thus far, etc.)
- the solo code review (hence the comments from me)

If there are outstanding improvements, a follow up PR should be sufficient. Thanks!

fortanix-vme/fortanix-vme-runner/src/lib.rs
```diff
@@ -659,32 +648,55 @@ impl<P: Platform + 'static> ServerState<P> {
...
-impl<P: Platform + 'static> Server<P> {
-    fn bind(enclave_name: String, port: u32) -> io::Result<Self> {
-        let command_listener = VsockListener::bind(VsockAddr::new(VMADDR_CID_ANY, port))?;
+async fn accept_stream(
+    listener: &mut Pin<Box<dyn AsyncListener>>,
+    mut local_addr: Option<&mut String>,
+    mut peer_addr: Option<&mut String>,
+) -> io::Result<Box<dyn AsyncStream>> {
+    poll_fn(|cx| {
+        let local = local_addr.as_mut().map(|addr| &mut **addr);
```
s-arash: Nit: you could do `local_addr.as_deref_mut()` instead and avoid having to introduce the `local` local! 
Also for `peer_addr`.

fortanix-vme/fortanix-vme-runner/src/lib.rs
```diff
@@ -698,30 +710,64 @@ impl<P: Platform + 'static> Server<P> {
...
+    }
+}
+
+/// Module of helper types for implementing [`enclave_runner`] traits.
+mod command {
+    use enclave_runner::stream_router::StreamRouter;
+
+    use super::*;
+
+    pub struct Command {
```
s-arash: Why do we need this struct? Couldn't `internal_new` be a fn on the `command` module?

fortanix-vme/fortanix-vme-runner/src/lib.rs
Taowyoo: nice catch

fortanix-vme/fortanix-vme-runner/src/lib.rs
Taowyoo: the `build` fn above expect to return a var that can be convertted to enclave_runner::Command.
Fyi
```
impl From<CommandFn> for Command {
    fn from(f: CommandFn) -> Self {
        Command {
            f
        }
    }
}
```

And you are right, i think we can get rid of `Command` struct just use a free function.

fortanix-vme/fortanix-vme-runner/src/lib.rs
```diff
@@ -634,94 +727,220 @@ impl<P: Platform + 'static> ServerState<P> {
...
-                enclave_args.insert(0, self.name.clone());
-                // Assume Platform::run will do some blocking logic
-                let handle = tokio::task::spawn_blocking(|| P::run(run_args));
-                let enclave = handle.await.expect("Failed to run enclave")?;
-                *state = EnclaveState::Running {
-                    enclave,
-                    args: enclave_args,
-                };
+        // This line is critical to ensure state keeps live otherwise it's moved into the future
+        let command_server_state = state.clone();
```
jethrogb, line 864: This doesn't make sense to me. It's fine if it's moved into the future. That should keep the state alive until the end of the `loop {}`.

fortanix-vme/fortanix-vme-runner/src/lib.rs
jethrogb: `/dev/vsock` is not a loader. But you could do `qemu-system-x86_64 -S`. The Nitro kernel API also separates loading and running, but nitro-enclaves-cli doesn't expose this.

Start QEMU like so `-S -qmp chardev:qmp -chardev socket,id=qmp,fd=X`
You can do this from Rust like so:
```
    let (control_stream, child_stream) = std::os::unix::net::UnixStream::pair()?;
    // unset CLOEXEC
    nix::fcntl::fcntl(&child_stream, nix::fcntl::FcntlArg::F_SETFD(nix::fcntl::FdFlag::empty()))?;

    std::process::Command::new("qemu-system-x86_64")
        ...

    let control_stream2 = control_stream.try_clone()?;
    let mut qmp = qapi::Qmp::new(qapi::Stream::new(BufReader::new(control_stream), control_stream2));
    qmp.handshake()?;
    qmp.execute(&qapi::qmp::cont {})?; 
```

This can be fixed in a follow-on PR

fortanix-vme/fortanix-vme-runner/src/lib.rs
Taowyoo: You are right. During local testing, i found this fix stuck issue. But then I found this change actually does not matter. Maybe I meet some issue with outdated compile target.

Created one PR for this , please review: https://github.com/fortanix/rust-sgx/pull/885

fortanix-vme/fortanix-vme-runner/src/lib.rs
Taowyoo: Hi folks, I create an issue as summary discussion so far: https://github.com/fortanix/rust-sgx/issues/883

Please take a look so that everyone is on same page
I also proposed an idea for fix , hope it makes sense to u.
@sardok @jethrogb @raoulstrackx

PR #911, opened 2026-03-18
RTE-839, RTE-840, RTE-841, RTE-842: Initial refactor of dcap-artifact-retrieval-tool

rui-alm:
Some experimental code which might make sense to use as the initial step for the refactoring of `dcap-artifact-retrieval` tool. It includes:

- Simplification of generics affecting `Client`, `BackoffService`, `PcsService`.
- API services are now stateless and therefore their functions are now static.
- Attestation specific clients that expose only their supported APIs. Azure behavior remains the same: only `pckcert` is obtained while other APIs still going via Intel PCS.
- Similar Rest APIs are implemented on internal methods which accept a superset of parameters to cover the possible combinations.
- Some lifetime reference annotations have been removed or moved to the functions that actually need them.
- Replacement of `build_input` with explicit initialization of input types.

RTE-785 contains the design spec and links to several tickets with the expected following steps, which also include some of the unfinished work from this experimental branch.

### DIFF
--- a/intel-sgx/dcap-artifact-retrieval/src/provisioning_client/azure.rs
+++ b/intel-sgx/dcap-artifact-retrieval/src/provisioning_client/azure.rs
@@ -17,7 +17,7 @@ use super::{
     StatusCode,
 };
 use crate::intel::IntelPCSClient;
-use crate::provisioning_client::{BackoffService, CachedService, PcsService};
+use crate::provisioning_client::{BackoffService, CachedService};
 use crate::{Error, IntelProvisioningClientBuilder, ProvisioningClient, WithApiVersion};
 
 /// A Provisioning Certificate client builder for Azure. It is based on the internal logic of the Azure DCAP
@@ -26,14 +26,12 @@ use crate::{Error, IntelProvisioningClientBuilder, ProvisioningClient, WithApiVe
 /// field in the past (see PROD-5800).
 /// For info on the Azure DCAP provider: <https://github.com/microsoft/Azure-DCAP-Client>
 pub struct AzureProvisioningClientBuilder {
-    api_version: PcsVersion,
     client_builder: IntelProvisioningClientBuilder,
 }
 
 impl AzureProvisioningClientBuilder {
     pub fn new(api_version: PcsVersion) -> Self {
         Self {
-            api_version,
             client_builder: IntelProvisioningClientBuilder::new(api_version),
         }
     }
@@ -44,7 +42,6 @@ impl AzureProvisioningClientBuilder {
     }
 
     pub fn build<F: for<'a> Fetcher<'a>>(&self, fetcher: F) -> AzureClient<F> {
-        let pckcert_service = PckCertService;
         let client = self.client_builder.build(
             fetcher,
         );
@@ -54,7 +51,6 @@ impl AzureProvisioningClientBuilder {
             azure_api_id: PckCertApi::API_VERSION_07_2021.to_owned(),
             pckcert_service: CachedService::new(
                 BackoffService::new(
-                    PcsService::new(pckcert_service),
                     self.client_builder.client_builder.retry_timeout.clone(),
                 ),
                 self.client_builder.client_builder.cache_capacity.clone(),
@@ -201,7 +197,7 @@ impl ProvisioningServiceApi for PckCertService {
     type Input<'a> = PckCertIn<'a>;
     type Output = PckCert<Unverified>;
 
-    fn build_request(&self, base_url: &str, input: &Self::Input<'_>) -> Result<(String, Vec<(String, String)>), Error> {
+    fn build_request(base_url: &str, input: &Self::Input<'_>) -> Result<(String, Vec<(String, String)>), Error> {
         // Re-implements `build_pck_cert_url` from Azure's DCAP Client
         // https://github.com/microsoft/Azure-DCAP-Client/blob/master/src/dcap_provider.cpp#L677
         // Constants from the Azure DCAP client:
@@ -217,7 +213,7 @@ impl ProvisioningServiceApi for PckCertService {
         Ok((url, Vec::new()))
     }
 
-    fn validate_response(&self, status_code: StatusCode) -> Result<(), Error> {
+    fn validate_response(status_code: StatusCode) -> Result<(), Error> {
         match &status_code {
             StatusCode::Ok => Ok(()),
             StatusCode::BadRequest => Err(Error::PCSError(status_code, "Invalid parameter")),
@@ -249,7 +245,6 @@ impl ProvisioningServiceApi for PckCertService {
     }
 
     fn parse_response(
-        &self,
         response_body: String,
         _response_headers: Vec<(String, String)>,
         _api_version: PcsVersion,
--- a/intel-sgx/dcap-artifact-retrieval/src/provisioning_client/azure.rs
+++ b/intel-sgx/dcap-artifact-retrieval/src/provisioning_client/azure.rs
@@ -10,16 +10,15 @@ use rustc_serialize::hex::ToHex;
 use serde::Deserialize;
 use std::collections::BTreeMap;
 use std::convert::TryInto;
-use std::marker::PhantomData;
 use std::time::Duration;
 
 use super::{
-    Client, ClientBuilder, Fetcher, PcsVersion, ProvisioningServiceApi,
+    Fetcher, PcsVersion, ProvisioningServiceApi,
     StatusCode,
 };
 use crate::intel::IntelPCSClient;
-use crate::provisioning_client::{BackoffService, CachedService, PcsService, QeIdService, TcbEvaluationDataNumbersService, TcbInfoService};
-use crate::{Error, IntelProvisioningClientBuilder, WithApiVersion};
+use crate::provisioning_client::{BackoffService, CachedService, PcsService};
+use crate::{Error, IntelProvisioningClientBuilder, ProvisioningClient, WithApiVersion};
 
 /// A Provisioning Certificate client builder for Azure. It is based on the internal logic of the Azure DCAP
 /// provider. Only the PCK certificates are downloaded from Azure. For others Intel is contacted.
@@ -46,12 +45,6 @@ impl AzureProvisioningClientBuilder {
 
     pub fn build<F: for<'a> Fetcher<'a>>(&self, fetcher: F) -> AzureClient<F> {
         let pckcert_service = PckCertService;
-        let qeid = QeIdService;
-        let sgx_tcbinfo = TcbInfoService::<platform::SGX> {_type: PhantomData};
-        let tdx_tcbinfo = TcbInfoService::<platform::TDX> {_type: PhantomData};
-        let sgx_evaluation_data_numbers = TcbEvaluationDataNumbersService::<platform::SGX> { _type: PhantomData };
-        let tdx_evaluation_data_numbers = TcbEvaluationDataNumbersService::<platform::TDX> { _type: PhantomData };
-
         let client = self.client_builder.build(
             fetcher,
         );
@@ -72,7 +65,7 @@ impl AzureProvisioningClientBuilder {
     }
 }
 
-struct AzureClient<F: for<'b> Fetcher<'b>> {
+pub struct AzureClient<F: for<'b> Fetcher<'b>> {
     azure_base_url: String,
     azure_client_id: String,
     azure_api_id: String,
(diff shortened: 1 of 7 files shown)


intel-sgx/dcap-artifact-retrieval/src/provisioning_client/azure.rs
```diff
@@ -41,34 +41,130 @@ impl AzureProvisioningClientBuilder {
...
+    pub fn build<F: for<'a> Fetcher<'a>>(&self, fetcher: F) -> AzureClient<F> {
+        let client = self.client_builder.build(
             fetcher,
-        )
+        );
+        AzureClient {
+            azure_base_url: PckCertApi::SECONDARY_CERT_URL.to_owned(),
+            azure_client_id: PckCertApi::DEFAULT_CLIENT_ID.to_owned(),
+            azure_api_id: PckCertApi::API_VERSION_07_2021.to_owned(),
+            pckcert_service: CachedService::new(
```
tvsfx, line 52: We don't actually seem to care about configuring `cache_capacity` and `cache_self_time`, right? We should probably introduce a shorthand to create a wrapped service. Like a function that can produce any `CachedService<T: ProvisioningServiceApi>`

intel-sgx/dcap-artifact-retrieval/src/provisioning_client/pccs.rs
```diff
@@ -257,182 +229,11 @@ impl<'inp> ProvisioningServiceApi<'inp> for PckCrlApi {
...
-        } else {
-            format!(
-                "{}/sgx/certification/v{}/tcb?fmspc={}&update=early",
-                self.base_url, api_version, fmspc,
-            )
-        };
-        Ok((url, Vec::new()))
-    }
-
-    fn validate_response(&self, status_code: StatusCode) -> Result<(), Error> {
```
tvsfx, line 313: Easy refactor; `validate_response` can just return a statically defined map of error codes and corresponding output strings (also, errors shouldnt be capitalized). We can do the boilerplate in common code.

intel-sgx/dcap-artifact-retrieval/src/provisioning_client/mod.rs
```diff
@@ -422,41 +639,36 @@ impl<O: Clone, T: for<'a> ProvisioningServiceApi<'a, Output = O> + Sync + ?Sized
...
-        let value = self.service.call_service::<F>(fetcher, input)?;
+        let value = self.service.call_service::<F, T>(fetcher, base_url, input)?;
         cache.insert(key, (value.clone(), SystemTime::now()));
         Ok(value)
     }
 }
 
-struct BackoffService<T: for<'a> ProvisioningServiceApi<'a> + Sync + ?Sized> {
-    service: PcsService<T>,
+struct BackoffService {
```
tvsfx, line 648: This might be nitpicky, but if you want to structure your services like this, it feels wrong to me to not bind a `BackOffService` to a specific `T`, same for `PcsService`.

The fact that the `call_service` function for `BackoffService` now has a different signature for the one for `CachedService` is a symptom of this; this asymmetry is confusing to users and makes it so you could not define a `CallableService` trait and define `CachedService` using another generic `C: CallableService` with field `service: C`. Not saying this design is better but the fact that you cannot do this seems odd.

intel-sgx/dcap-artifact-retrieval/src/provisioning_client/mod.rs
```diff
@@ -422,41 +639,36 @@ impl<O: Clone, T: for<'a> ProvisioningServiceApi<'a, Output = O> + Sync + ?Sized
...
+        &self,
         fetcher: &'a F,
-        input: &<T as ProvisioningServiceApi<'a>>::Input,
-    ) -> Result<<T as ProvisioningServiceApi<'a>>::Output, Error> {
+        base_url: &str,
+        input: &T::Input<'_>,
+    ) -> Result<T::Output, Error> {
         if let Some(retry_timeout) = self.retry_timeout {
-            let op = || match self.service.call_service::<F>(fetcher, input) {
+            let op = || match PcsService::call_service::<F, T>(fetcher, base_url, input) {
```
tvsfx, line 671: cfr. my comment above: note how you're forced to specify `T` here, even though we know which service we want to call at the time when our top-level cached service is created. Technically, the reason this doesnt work is because the compiler doesnt reason backwards from `T::Input` to the `ProvisioningServiceApi` we need. That's why its more convenient to bind this instance into the struct when we create it.

intel-sgx/dcap-artifact-retrieval/src/provisioning_client/mod.rs
```diff
@@ -341,53 +575,39 @@ impl ClientBuilder {
     }
 }
 
-struct PcsService<T: for<'a> ProvisioningServiceApi<'a> + Sync + ?Sized> {
-    service: Box<T>,
```
tvsfx, line 345: Now that we don't need trait objects anymore, we can get rid of the `?Sized` (and maybe `Sync`? not sure why thats here) bounds here, removing them from all places that depend on this.

intel-sgx/dcap-artifact-retrieval/src/provisioning_client/azure.rs
```diff
@@ -41,34 +41,130 @@ impl AzureProvisioningClientBuilder {
...
+    fn pckcert(
+        &self,
+        api_key: &Option<String>,
+        encrypted_ppid: Option<&EncPpid>,
+        pce_id: &PceId,
+        cpu_svn: &CpuSvn,
+        pce_isvsvn: PceIsvsvn,
+        qe_id: Option<&QeId>,
+    ) -> Result<PckCert<Unverified>, Error> {
+        let input = PckCertIn { encrypted_ppid, pce_id, cpu_svn, pce_isvsvn, qe_id, api_version: self.client.client.api_version, api_key, azure_client_id: &self.azure_client_id, azure_api_version: &self.azure_api_id };
```
tvsfx, line 82: I think this is a good change; this feels like the right place to massage input parameters. Made possible by the fact that you got rid of the gnarly trait objects :)

intel-sgx/dcap-artifact-retrieval/src/provisioning_client/azure.rs
```diff
@@ -41,34 +41,130 @@ impl AzureProvisioningClientBuilder {
...
+        self.pckcert_service.call_service(&self.client.client.fetcher, &self.azure_base_url, &input)
+    }
+    fn pckcerts(&self, pck_id: &PckID) -> Result<PckCerts, Error> {
```
tvsfx, line 85: Do I understand correctly that the old `pckcerts_with_fallback` implementation was basically a way for the Azure client to emulate the `pckcerts` API, meaning we can just use it as the `pckcerts` implementation now that the clients have been split?

intel-sgx/dcap-artifact-retrieval/src/provisioning_client/azure.rs
```diff
@@ -41,34 +41,130 @@ impl AzureProvisioningClientBuilder {
...
+
+        // BTreeMap by default is Ascending
+        let pck_certs: Vec<_> = pckcerts_map.into_iter().rev().map(|(_, v)| v).collect();
+        pck_certs
+            .try_into()
+            .map_err(|e| Error::PCSDecodeError(format!("{}", e).into()))
+    }
+
+    fn sgx_tcbinfo(&self, fmspc: &pcs::Fmspc, evaluation_data_number: Option<u16>) -> Result<pcs::TcbInfo<platform::SGX>, Error> {
+        self.client.sgx_tcbinfo(fmspc, evaluation_data_number)
```
tvsfx, line 143: No need to refactor this now, but FYI, there's a crate to make this type of boilerplate more concise if you have a lot of it: https://crates.io/crates/delegate

tvsfx: Since the azure client just contains a regular `client`, I don't think you need this though; you can just have e.g. `AsRef<Client>` bounds where you need `Client` functionality and implement `AsRef<Client>` for `AzureClient`

intel-sgx/dcap-artifact-retrieval/src/provisioning_client/common.rs
```diff
@@ -19,48 +17,6 @@ pub const TCB_INFO_ISSUER_CHAIN_HEADER_V4: &'static str = "TCB-Info-Issuer-Chain
 pub const ENCLAVE_ID_ISSUER_CHAIN_HEADER: &'static str = "SGX-Enclave-Identity-Issuer-Chain";
 pub const TCB_EVALUATION_DATA_NUMBERS_ISSUER_CHAIN: &'static str = "TCB-Evaluation-Data-Numbers-Issuer-Chain";
 
-pub struct PckCertsApiNotSupported;
```
tvsfx, line 22: yay for not introducing spurious abstractions :)

intel-sgx/dcap-artifact-retrieval/src/provisioning_client/azure.rs
```diff
@@ -41,34 +41,130 @@ impl AzureProvisioningClientBuilder {
...
+    azure_client_id: String,
+    azure_api_id: String,
+    pckcert_service: CachedService<PckCertService>,
+    client: IntelPCSClient<F>,
 }
 
+impl<F: for<'a> Fetcher<'a>> ProvisioningClient for AzureClient<F> {
+    fn pckcert(
+        &self,
+        api_key: &Option<String>,
```
tvsfx, line 75: I do feel it might not be necessary to have the `api_key` as an argument here; if we dont expect to use different keys, I think it was fine being part of the client's state?

intel-sgx/dcap-artifact-retrieval/src/provisioning_client/mod.rs
rui-alm: The reason I've gone with this approach is because `PcsService` is now stateless, which meant `BackoffService` no longer needed to have a `service` member. This means none of `BackoffService` members need the type `T` anymore. Only the function `call_service` needs to be aware of the type to call the corresponding (static) `T::call_service`.

I see `CachedService` differently because it actually has a member `cache` which requires the type `T`, has the cache is returning a value of type `T::Output`.

Maybe `BackoffService` is an over-simplistic example, but if there were other functions, including static ones, that don't required any use of type `T`, their invocation would still require one to be passed. With this approach, only `call_service` needs to be parameterized with `T` and other functions wouldn't.

In some quick sandboxed test I've done, the compiler even warns me that type `T` is unused if I don't have a member variable that uses it, even if there a member function that does.

intel-sgx/dcap-artifact-retrieval/src/provisioning_client/azure.rs
rui-alm: Yes. It also applies to PCCS. In fact this branch has introduced the duplication of the fallback code, but I have a ticket to deal with it already.

Rather than having a `pckcerts_with_fallback` that was implementing 2 behaviours, 
- the call to the actual PCS `pckcerts` and
- the fallback behavior (based on `pckcert`) that would only be used for Azure and PCCS
 
I've split the logic so that PCS implements the supported `pckcerts` REST API call, and Azure and PCCS implement only the "fallback" behavior.

The fallback `pckcerts` is slightly different from the PCS `pckcerts`, because we use the `PckId` instead of the `EncPpid` and `PceId`.

intel-sgx/dcap-artifact-retrieval/src/provisioning_client/azure.rs
rui-alm: The reason I've put the `api_key` here is because it's only necessary for `pckcert` (and PCS `pckcerts`). Other REST API calls don't have it. When I say necessary, it was only necessary for V3 and it's optional for V4. Given the nature of this field, I'd assume it's a leftover from V3, perhaps a way to not break V3 compatible code, but probably makes no difference to pass an `api_key` or not when using V4, as usually this is used to authenticate/authorize an API call. Perhaps there are some statistics on Intel's portal, but I doubt it would make difference on the API response.

Perhaps as part of the "Remove V3 support" ticket, we can discuss whether we still need to have this field or should we get rid of it altogether.

intel-sgx/dcap-artifact-retrieval/src/provisioning_client/mod.rs
rui-alm: I think I've gotten rid of all of those, but if you've found some left over, let me know. This comment is on the removed code and I haven't found any more of those on the remaining code.

tvsfx: Yes, I wrote this assuming we might resurrect the `T` parameter on the other service types. If we don't then this comment might not be relevant.

intel-sgx/dcap-artifact-retrieval/src/provisioning_client/mod.rs
tvsfx: I understand why you want to do this from a technical perspective :) I just don't think it's the correct thing to do from a domain modeling perspective. See my other comment for why modeling things this way can get awkward, especially if you have multiple methods that rely on your generic (you'll have to specify what the generic is each time): https://github.com/fortanix/rust-sgx/pull/911/changes#r2991883027

> but if there were other functions, including static ones, that don't required any use of type T, their invocation would still require one to be passed.

So what I'm arguing is that the concept of a `Service` doesn't make sense without it being tied to a specific API; since we say that a `CachedService` needs to be tied to a specific API, I think it's weird to then also say that a `BackoffService` does not need to be tied to a specific API, even though it's also a service. I'm hence implicitly also saying that I don't want to model such static functions as part of the `Service` API. 

> In some quick sandboxed test I've done, the compiler even warns me that type T is unused if I don't have a member variable that uses it, even if there a member function that does.

Did you not use `PhantomData`? That's what that's for.

PR #768, opened 2025-06-04
 AEX notify detection support

jethrogb:

PR #912, opened 2026-03-30
sgxs-tools: Support AEX-Notify feature

jovanbulck:
Changes:

- sgxs-sign: support signing enclaves with the AEX-Notify feature enabled
- sgxs-info: print whether a given TCS has AEX-Notify enabled

AEX-Notify is currently not supported by EDP, but this commit allows other runtimes building on sgxs-tools (e.g., [bare-sgx](https://github.com/jovanbulck/bare-sgx/tree/sgxs)) to support AEX-Notify.


jovanbulck:
I noticed now this seems to be related to #768 

@jethrogb any updates on that? Has there been any plans to support AEX-Notify for `sgxs-tools` and/or Rust-EDP?

PR #914, opened 2026-03-31
[RTE-894] Correcting SGX-ISA types and functions related to TDX support

Taowyoo:
Since REPORTMACSTRUCT is used by both SGX and TDX.
The naming should not strict to TDX.

This PR:
1. Renames the type to `ReportMacStruct` and move to the main crate level instead of under `tdx`  module namespace.
2. Keeps backward compatibility by adding a type alias using old name.
3. `verify` function now is part of `ReportMac` instead of  `TdxReportV1`, although maintaining compatibility that the `verify`  function still exists in `TdxReportV1`
4. Move the internal `tdx_arch` module to be part of entire `arch`  module in the crate.
5. Refactor the `arch` module to allow non-upstreamed new instruction sets to be used without breaking the entire module namespace in the main `lib.rs` file.
7. Move all `TdxError` crates out as it is more relevant to `tdx-ql`  crate. The error from SGX-ISA should be only the `ErrorCode` types.  `TdxError` is not part of the ISA.
8. Add proper enum to `ReportType` possible known constants, instead of a singular constants.

### DIFF
--- a/intel-sgx/sgx-isa/src/arch.rs
+++ b/intel-sgx/sgx-isa/src/arch.rs
@@ -15,6 +15,10 @@ pub struct Align16<T>(pub T);
 #[repr(align(128))]
 pub struct Align128<T>(pub T);
 
+/// Wrapper struct to force 256-byte alignment.
+#[repr(align(256))]
+pub struct Align256<T>(pub T);
+
 /// Wrapper struct to force 512-byte alignment.
 #[repr(align(512))]
 pub struct Align512<T>(pub T);
@@ -68,5 +72,28 @@ pub fn ereport(
         );
 
         report.assume_init()
+    }
+}
+
+/// Call the `EVERIFYREPORT2` instruction to verify a REPORT MAC struct.
+/// The concrete type is [`crate::ReportMac`].
+pub fn everifyreport2(tdx_report_mac: &Align256<[u8; 256]>) -> Result<(), u32> {
+    unsafe {
+        let error: u32;
+        asm!(
+            "xchg %rbx, {0}",
+            "enclu",
+            "mov {0}, %rbx",
+            "jz 1f",
+            "xor %eax, %eax",
+            "1:",
+            inout(reg) tdx_report_mac => _,
+            inlateout("eax") Enclu::EVerifyReport2 as u32 => error,
+            options(att_syntax, nostack),
+        );
+        match error {
+            0 => Ok(()),
+            err => Err(err),
+        }
     }
 }
--- a/intel-sgx/sgx-isa/src/arch.rs
+++ b/intel-sgx/sgx-isa/src/arch.rs
@@ -5,79 +5,96 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 use super::Enclu;
 use core::arch::asm;
-use core::mem::MaybeUninit;
 
-/// Wrapper struct to force 16-byte alignment.
-#[repr(align(16))]
-pub struct Align16<T>(pub T);
+/// Group all functions and types that is already upstreamed in the sgxstd
+#[cfg(all(target_env = "sgx", not(feature = "sgxstd")))]
+mod upstream {
+    use super::*;
+    use core::mem::MaybeUninit;
 
-/// Wrapper struct to force 128-byte alignment.
-#[repr(align(128))]
-pub struct Align128<T>(pub T);
+    /// Wrapper struct to force 16-byte alignment.
+    #[repr(align(16))]
+    pub struct Align16<T>(pub T);
+
+    /// Wrapper struct to force 128-byte alignment.
+    #[repr(align(128))]
+    pub struct Align128<T>(pub T);
+
+    /// Wrapper struct to force 512-byte alignment.
+    #[repr(align(512))]
+    pub struct Align512<T>(pub T);
+
+    /// Call the `EGETKEY` instruction to obtain a 128-bit secret key.
+    pub fn egetkey(request: &Align512<[u8; 512]>) -> Result<Align16<[u8; 16]>, u32> {
+        unsafe {
+            let mut out = MaybeUninit::uninit();
+            let error;
+
+            asm!(
+                // rbx is reserved by LLVM
+                "xchg %rbx, {0}",
+                "enclu",
+                "mov {0}, %rbx",
+                inout(reg) request => _,
+                inlateout("eax") Enclu::EGetkey as u32 => error,
+                in("rcx") out.as_mut_ptr(),
+                options(att_syntax, nostack),
+            );
+
+            match error {
+                0 => Ok(out.assume_init()),
+                err => Err(err),
+            }
+        }
+    }
+
+    /// Call the `EREPORT` instruction.
+    ///
+    /// This creates a cryptographic report describing the contents of the current
+    /// enclave. The report may be verified by the enclave described in
+    /// `targetinfo`.
+    pub fn ereport(
+        targetinfo: &Align512<[u8; 512]>,
+        reportdata: &Align128<[u8; 64]>,
+    ) -> Align512<[u8; 432]> {
+        unsafe {
+            let mut report = MaybeUninit::uninit();
+
+            asm!(
+                // rbx is reserved by LLVM
+                "xchg %rbx, {0}",
+                "enclu",
+                "mov {0}, %rbx",
+                inout(reg) targetinfo => _,
+                in("eax") Enclu::EReport as u32,
+                in("rcx") reportdata,
+                in("rdx") report.as_mut_ptr(),
+                options(att_syntax, preserves_flags, nostack),
+            );
+
+            report.assume_init()
+        }
+    }
+}
+
+// Export the function in the `upstream` group if not using `sgxstd`
+#[cfg(all(target_env = "sgx", not(feature = "sgxstd")))]
+pub use self::upstream::*;
+
+// Export function in the `fortanix_sgx::arch` namespace if using `sgxstd`
+#[cfg(all(target_env = "sgx", feature = "sgxstd"))]
+pub use std::os::fortanix_sgx::arch::*;
+
+// Functions and types below is not yet upstreamed and will be added to the
+// upstream in the future.
 
 /// Wrapper struct to force 256-byte alignment.
 #[repr(align(256))]
 pub struct Align256<T>(pub T);
 
-/// Wrapper struct to force 512-byte alignment.
-#[repr(align(512))]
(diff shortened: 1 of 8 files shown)


intel-sgx/sgx-isa/src/tdx.rs
```diff
@@ -101,12 +104,12 @@ struct_def! {
     }
 }
 
-impl TdxReportMac {
+impl ReportMac {
```
gilanghamidy: This has to be put outside `tdx` module, in the main `lib.rs`

gilanghamidy (dismissed):

intel-sgx/sgx-isa/src/tdx.rs
gilanghamidy: This method has to be moved into in `impl TeeReport` instead as the `EVERIFYREPORT2` is also TEE generic.

gilanghamidy:
@Taowyoo I will help taking over the further revision, so you don't have to worry about it. :)

intel-sgx/sgx-isa/src/arch.rs
```diff
@@ -70,3 +74,26 @@ pub fn ereport(
         report.assume_init()
     }
 }
+
+/// Call the `EVERIFYREPORT2` instruction to verify a REPORT MAC struct.
+/// The concrete type is [`crate::ReportMac`].
+pub fn everifyreport2(tdx_report_mac: &Align256<[u8; 256]>) -> Result<(), u32> {
```
jethrogb: This still says tdx

intel-sgx/sgx-isa/src/lib.rs
```diff
@@ -808,6 +821,100 @@ impl Default for Keypolicy {
...
+/// Ref: Intel® Trust Domain CPU Architectural Extensions, table 2-5.
+/// Version: 343754-002US, MAY 2021
+/// Link TDX: <https://cdrdv2.intel.com/v1/dl/getContent/733582>
+/// Link 256BITSGX: <https://cdrdv2-public.intel.com/851355/319433-057-architecture-instruction-set-extensions-programming-reference.pdf>
+#[repr(C, align(256))]
+#[cfg_attr(
+    feature = "large_array_derive",
+    derive(Clone, Debug, Eq, PartialEq)
+)]
+pub struct ReportMac {
```
jethrogb: Every other type in this crate uses the name from the SDM.

intel-sgx/sgx-isa/src/lib.rs
```diff
@@ -808,6 +821,100 @@ impl Default for Keypolicy {
...
+/// and the future 256BITSGX
+///
+/// Ref: Intel® Trust Domain CPU Architectural Extensions, table 2-5.
+/// Version: 343754-002US, MAY 2021
+/// Link TDX: <https://cdrdv2.intel.com/v1/dl/getContent/733582>
+/// Link 256BITSGX: <https://cdrdv2-public.intel.com/851355/319433-057-architecture-instruction-set-extensions-programming-reference.pdf>
+#[repr(C, align(256))]
+#[cfg_attr(
+    feature = "large_array_derive",
+    derive(Clone, Debug, Eq, PartialEq)
```
jethrogb: Missing `Default`

intel-sgx/sgx-isa/src/lib.rs
```diff
@@ -808,6 +821,100 @@ impl Default for Keypolicy {
...
+)]
+pub struct ReportMac {
+    /// (  0) TEE Report type
+    pub report_type: TeeReportType,
+    /// (  4) Reserved, must be zero
+    pub reserved1: [u8; REPORT_MAC_STRUCT_RESERVED1_BYTES],
```
jethrogb: Reserved fields should be prefixed with `_`

intel-sgx/sgx-isa/src/lib.rs
```diff
@@ -808,6 +821,100 @@ impl Default for Keypolicy {
...
+    pub report_data: [u8; REPORT_DATA_SIZE],
+    /// (192) Reserved, must be zero
+    pub reserved2: [u8; REPORT_MAC_STRUCT_RESERVED2_BYTES],
+    /// (224) The Message Authentication Code over this structure
+    pub mac: [u8; TEE_MAC_SIZE],
+}
+}
+
+impl ReportMac {
+    pub const UNPADDED_SIZE: usize = 256;
```
jethrogb, line 903: Missing `TRUNCATED_SIZE` const

intel-sgx/sgx-isa/src/lib.rs
```diff
@@ -808,6 +821,100 @@ impl Default for Keypolicy {
...
+    #[cfg(target_env = "sgx")]
+    pub fn verify(&self) -> Result<(), ErrorCode> {
+        everifyreport2(self.as_ref())
+            // Same as `egetkey` reasoning: unwrap is okay here
+            .map_err(|e| ErrorCode::try_from(e).unwrap())
+    }
+}
+
+#[cfg(target_env = "sgx")]
+impl AsRef<Align256<[u8; ReportMac::UNPADDED_SIZE]>> for ReportMac {
```
jethrogb: I don't understand why this is necessary.  `struct_def` should take care of this impl.

jethrogb (commented):
A lot of my comments are simply about consistency with how other types were already being defined in this crate. Please review all types that were recently added for consistency.

intel-sgx/sgx-isa/src/lib.rs
gilanghamidy: I was following initial @Taowyoo implementation when moving it here. I think it was for the convertibility from `ReportMac` type when being passed into `everifyreport2` function that has `Align256` as parameter.

intel-sgx/sgx-isa/src/lib.rs
gilanghamidy: Do we need that? Because the MAC itself is verified by `everifyreport2`, and software basically has no means to verify that. So we don't need the implementation like in the `sgx_isa::Report` type where there is `verify` function that uses a closure that splits the MAC and the rest of the data.

intel-sgx/sgx-isa/src/lib.rs
gilanghamidy: Refactored and addressed this in a more proper way to cater upstreamed and non-upstreamed types.

intel-sgx/sgx-isa/src/lib.rs
gilanghamidy: I assume this is closed as we don't need MAC truncation as in the legacy SGX report ISA which the enclave software verifies the MAC manually.

intel-sgx/sgx-isa/src/tdx.rs
```diff
@@ -12,105 +12,31 @@ use crate::arch;
...
 }
 
-#[cfg(target_env = "sgx")]
-impl AsRef<tdx_arch::Align256<[u8; TdxReportMac::UNPADDED_SIZE]>> for TdxReportMac {
-    fn as_ref(&self) -> &tdx_arch::Align256<[u8; Self::UNPADDED_SIZE]> {
-        unsafe { &*(self as *const _ as *const _) }
-    }
-}
+// Ensure new type name is backward compatibe
+pub type TdxReportMac = super::ReportMacStruct;
```
tvsfx: Note: we re-expose `TdxReportMac` but not `TeeReportType` or any of the constants. We should do all or nothing imo.

intel-sgx/sgx-isa/src/tdx.rs
```diff
@@ -267,75 +192,8 @@ impl TdxReportV1 {
...
-            TdxError::OutOfMemory => f.write_str("Not enough memory is available to complete this operation"),
-            TdxError::QuoteFailure => f.write_str("Failed to get the TD Quote"),
-            TdxError::ReportFailure => f.write_str("Failed to get the TD Report"),
-            TdxError::TdcallFailure => f.write_str("TD call failed"),
-            TdxError::UnsupportedAttKeyId => f.write_str("The platform Quoting infrastructure does not support any of the keys described in att_key_id_list"),
-            TdxError::VsockFailure => f.write_str("vsock related failure"),
-            TdxError::WrongDevice => f.write_str("Unsupported device"),
-            TdxError::Unexpected(num) => f.write_fmt(format_args!("Unexpected errno: {}", num)),
-        }
+    pub fn verify(&self) -> Result<(), ErrorCode> {
```
tvsfx: misleading implementation cause it only verifies the mac, not the whole struct

intel-sgx/sgx-isa/src/lib.rs
```diff
@@ -808,6 +810,106 @@ impl Default for Keypolicy {
...
+    }
+}
+
+impl ReportType {
+    pub const UNPADDED_SIZE: usize = 4;
+}
+
+// All variants of ReportVersion that is valid for TDX report
+enum_def! {
+#[derive(Clone,Copy,Debug,PartialEq,Eq)]
```
tvsfx, line 843: nit: format this

intel-sgx/sgx-isa/src/tdx.rs
```diff
@@ -12,105 +12,31 @@ use crate::arch;
...
-        /// (128) Data provided by the user
-        pub report_data: [u8; TDX_REPORT_DATA_SIZE],
-        /// (192) Reserved, must be zero
-        pub reserved2: [u8; TDX_REPORT_MAC_STRUCT_RESERVED2_BYTES],
-        /// (224) The Message Authentication Code over this structure
-        pub mac: [u8; TEE_MAC_SIZE],
-    }
+// All variants of REPORTTYPE.VERSION that is valid for TDX report
+enum_def! {
+#[derive(Clone,Copy,Debug,PartialEq,Eq)]
```
tvsfx: nit: format this

intel-sgx/sgx-isa/src/tdx.rs
gilanghamidy: We decide to do nothing as it is not pushed to crates.io yet and we will bump minor version anyway so we can make breaking changes.

intel-sgx/sgx-isa/src/tdx.rs
gilanghamidy: Removing it

intel-sgx/sgx-isa/src/lib.rs
gilanghamidy: Yeah I was copy pasting code above and it was not formatted. I'll format it but leave the other as is for future KTLO.

tvsfx (approved):

PR #915, opened 2026-04-07
Don't force nodes. prefix on all api paths and start using new paths from em-client

adrian-brabete:
This PR also enables the use of the new api paths from the em-client

### DIFF
--- a/em-app/src/utils.rs
+++ b/em-app/src/utils.rs
@@ -62,15 +62,15 @@ pub fn get_runtime_configuration(
     
     config.push_cert(cert, key).map_err(|e| format!("TLS configuration failed: {:?}", e))?;
 
-    let base_path = if server.starts_with("ccm.") && server.ends_with("fortanix.com") {
-        format!("nodes.{}", server)
+   let tls_client = if (server.starts_with("ccm.") || server.starts_with("em.")) && server.ends_with("fortanix.com") {
+        MbedSSLClient::new_with_sni(Arc::new(config), true, Some(format!("nodes.{}", server)))
     } else {
-        server.to_owned()
+        MbedSSLClient::new(Arc::new(config), true)
     };
 
-    let ssl = MbedSSLClient::new_with_sni(Arc::new(config), true, Some(base_path.clone()));
-    let connector = HttpsConnector::new(ssl);
-    let mut client = Client::try_new_with_connector(&format!("https://{}:{}", base_path, port), None, connector).map_err(|e| format!("EM SaaS request failed: {:?}", e))?;
+    let connector = HttpsConnector::new(tls_client);
+    let mut client = Client::try_new_with_connector(&format!("https://{}:{}", server, port), None, connector)
+        .map_err(|e| format!("CCM client construction failed: {:?}", e))?;
     client.set_use_new_paths(true);
     let response = client.get_runtime_application_config(expected_hash).map_err(|e| format!("Failed requesting workflow config response: {:?}", e))?;
 
--- a/em-app/src/utils.rs
+++ b/em-app/src/utils.rs
@@ -61,10 +61,17 @@ pub fn get_runtime_configuration(
     }
     
     config.push_cert(cert, key).map_err(|e| format!("TLS configuration failed: {:?}", e))?;
-    
-    let ssl = MbedSSLClient::new_with_sni(Arc::new(config), true, Some(format!("nodes.{}", server)));
+
+    let base_path = if server.starts_with("ccm.") && server.ends_with("fortanix.com") {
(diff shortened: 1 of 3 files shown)


em-app/src/utils.rs
```diff
@@ -61,10 +61,18 @@ pub fn get_runtime_configuration(
     }
     
     config.push_cert(cert, key).map_err(|e| format!("TLS configuration failed: {:?}", e))?;
-    
-    let ssl = MbedSSLClient::new_with_sni(Arc::new(config), true, Some(format!("nodes.{}", server)));
+
+    let base_path = if sever.starts_with("ccm.") && server.ends_with("fortanix.com") {
+        format!("nodes.{}", server)
+    } else {
+        server.to_owned()
```
Taowyoo: I understand first branch is for CCM, what the case for this branch?
Other armor services?

adrian-brabete: Yes indeed, its for the armor platform.

Pagten:
Let's edit the title of the PR to make clear that this PR also enables the use of the new CCM-on-Armor API paths.

em-app/src/utils.rs
```diff
@@ -61,10 +61,17 @@ pub fn get_runtime_configuration(
     }
     
     config.push_cert(cert, key).map_err(|e| format!("TLS configuration failed: {:?}", e))?;
-    
-    let ssl = MbedSSLClient::new_with_sni(Arc::new(config), true, Some(format!("nodes.{}", server)));
+
+    let base_path = if server.starts_with("ccm.") && server.ends_with("fortanix.com") {
```
Pagten: `base_path` doesn't really seem appropriate for this variable. We could use `hostname` or stick with `server`.

em-app/src/utils.rs
```diff
@@ -61,10 +61,17 @@ pub fn get_runtime_configuration(
...
+        format!("nodes.{}", server)
```
Pagten: This branch is not sufficient for backwards compatibility. Previously we had:
```
let ssl = MbedSSLClient::new_with_sni(Arc::new(config), true, Some(format!("nodes.{}", server)));
let connector = HttpsConnector::new(ssl);
let client = Client::try_new_with_connector(&format!("https://{}:{}/v1/runtime/app_configs", server, port), None, connector).map_err(|e| format!("EM SaaS request failed: {:?}", e))?;
```
So we're setting up a TCP connection to, for example, `ccm.fortanix.com`, but we're using an SNI value of `nodes.ccm.fortanix.com`.

With these changes, we would be setting up a TCP connection to `nodes.ccm.fortanix.com`. I think this would work, but for backwards compatibility it is better not to change the approach.

Also, it would be good to also take the backcompat code path for `server` names starting with `em.` (for enclave manager).

em-app/src/utils.rs
```diff
@@ -61,10 +61,17 @@ pub fn get_runtime_configuration(
...
+    } else {
+        server.to_owned()
+    };
+
+    let ssl = MbedSSLClient::new_with_sni(Arc::new(config), true, Some(base_path.clone()));
     let connector = HttpsConnector::new(ssl);
-    let client = Client::try_new_with_connector(&format!("https://{}:{}/v1/runtime/app_configs", server, port), None, connector).map_err(|e| format!("EM SaaS request failed: {:?}", e))?;
+    let mut client = Client::try_new_with_connector(&format!("https://{}:{}", base_path, port), None, connector).map_err(|e| format!("EM SaaS request failed: {:?}", e))?;
```
Pagten: Dropping the `v1/runtime/app_configs` path here is intentional, right?

em-app/src/utils.rs
Pagten: This seems closer to what we need:
```rust
let tls_client = if (server.starts_with("ccm.") || server.starts_with("em.")) && server.ends_with("fortanix.com") {
    MbedSSLClient::new_with_sni(Arc::new(config), true, Some(format!("nodes.{}", server)))
} else {
    MbedSSLClient::new(Arc::new(config), true)
};

let connector = HttpsConnector::new(tls_client);
let mut client = Client::try_new_with_connector(&format!("https://{}:{}", server, port), None, connector)
    .map_err(|e| format!("CCM client construction failed: {:?}", e))?;
```

Pagten (changes requested):

em-app/src/utils.rs
adrian-brabete: Yes, it was aleady going to be dropped as part of the `try_new_with_connector` so setting it here wasn't doing anything.

Pagten (approved):

PR #923, opened 2026-04-17
[RTE-924] Fix `dcap-artifact-retrieval` retrieving incorrect TCB Info for TDX from SGX URL

gilanghamidy:
The error is also uncaught because the unit testing for PCCS wasn't added before.


rui-alm (approved):

gilanghamidy:
Work is tracked as RTE-924

PR #924, opened 2026-04-22
Improve insecure-time frequency estimation

s-arash:
This PR makes the following improvements:

- Fix the bug in calculating next sync time in `fn resync_clocks()`
  The current buggy behavior is to set next sync time to a long time in the future the first time, and to practically infinity the second time. This means that in the lifetime of a program, the clock is only resynced twice (at most).

- Estimate recent frequency instead of average freq over program lifetime
   The TSC freq can change slightly over time, we want the recent frequency to accurately estimate passed time, not
       the average frequency.
- Keep track of latest system time to have a more accurate estimate of current time
  instead of estimating current time from TSC ticks since program start time.
- Cleanup/simplify the code.
- Make resyncs more robust to unreliable reads of TSC, TSC getting reset, etc.
- Remove `initial_frequency` from `LearningFreqTscBuilder` as it was ignored

### DIFF
--- a/intel-sgx/insecure-time/Cargo.toml
+++ b/intel-sgx/insecure-time/Cargo.toml
@@ -23,6 +23,7 @@ spin = { version = "0.10", default-features = false, features = ["rwlock", "mute
 [dev-dependencies]
 rand = "0.8"
 clap = { version = "4.4", features = ["derive"] }
+crossterm = "0.29"
 
 [features]
 default = ["std"]
--- a/intel-sgx/insecure-time/Cargo.toml
+++ b/intel-sgx/insecure-time/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "insecure-time"
-version = "0.2.1"
+version = "0.3.0"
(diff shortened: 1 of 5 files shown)


Goirad (dismissed):
LGTM pending minor fix to weighing formula

Goirad (approved):
LGTM

intel-sgx/insecure-time/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "insecure-time"
-version = "0.2.1"
```
jethrogb, line 3: Why? Did the API change?

s-arash: Yes, the unused `fn set_initial_frequency` was removed.

jethrogb: If a function is removed, then the old version should still work. So dependents of insecure-time should be updated so they accept 0.2 or 0.3.

PR #925, opened 2026-04-24
[RTE-871] Add Root CA CRL in collateral certificate chain verification

gilanghamidy:
Previously, `pcs` crate disregards the existence of the Root CA CRL and we did not take it as a part of collateral verification process. This PR adds the functionality to allow all verification functions also accepts root CA CRL in case it is available.

This PR refactors some of the duplication on the certificate chain verification function, and centralize most of them in a single function. At the same time, it adds parameters to accept root CA CRL if it is available. A new function `verify_with_rootcacrl` is introduced to allow backward compatibility for older `verify` functions that does not provide such CRL. As such, passing empty Root CA CRL does not fail verification.

A new type `RootCaCrl` is introduced to act as a container of a verified CRL that can be passed to the function. It behaves similarly to `PckCrl`, but ultimately, it is the object that is accepted by the `verify_with_rootcacrl` parameter to ensure validation of the CRL has been done before passing it.

---
This is **part 1 of 2 in a stack** made with GitButler:
- <kbd>&nbsp;2&nbsp;</kbd> #926 
- <kbd>&nbsp;1&nbsp;</kbd> #925 👈


intel-sgx/pcs/src/root_ca_crl.rs
Taowyoo, line 1: could we rename this to `root_ca_crl.rs‎`

intel-sgx/pcs/src/lib.rs
```diff
@@ -281,6 +284,65 @@ fn create_cert_chain(certs: &Vec<String>) -> Result<(Vec<MbedtlsBox<Certificate>
     }
 }
 
+/// Function to verify the PCS certificate chain that is attached in the response
+/// from the PCS server. Previously it was replicated in each `verify` function. But
+/// we now centralize this in here and shared among all PCS types. Also, we incorporate
+/// passing the root CA CRL in here if available.
+#[cfg(feature = "verify")]
+fn build_and_verify_cert_chain<B: Deref<Target = [u8]>>(
+    ca_chain: &Vec<String>,
```
Taowyoo: Understand that this function is refactored from duplicated code.
But need to notice that this function is only used for verify CA certificate chain.

Could you clarify this in rustdoc above?
This is needed since the error throw by this function is related to CA.

intel-sgx/pcs/src/lib.rs
```diff
@@ -359,3 +421,11 @@ impl fmt::Display for TcbStatus {
         }
     }
 }
+
+#[cfg(feature = "verify")]
+pub(crate) fn as_mbedtls_crl(crl_pem: &String) -> Result<Crl, Error> {
```
Taowyoo: nit

```suggestion
pub(crate) fn as_mbedtls_crl(crl_pem: &str) -> Result<Crl, Error> {
```

intel-sgx/pcs/src/lib.rs
gilanghamidy: I wrote in the docs above that this function is only used for verifying certificate chain that is a part of PCS server response, and indeed all the errors are related to CA and CRL. What else do you expect to add in the information?

intel-sgx/pcs/src/lib.rs
```diff
@@ -281,6 +284,65 @@ fn create_cert_chain(certs: &Vec<String>) -> Result<(Vec<MbedtlsBox<Certificate>
...
+        } else {
+            None
+        };
+
+        let root_crl_ref = if let Some(crl) = &mut root_crl {
+            Some(crl)
+        } else {
+            None
+        };
+        Certificate::verify(&trust_ca, &root_list, root_crl_ref, None).map_err(|e| Error::InvalidQe3Id(e))?;
```
rui-alm, line 325: Are the `trust_ca` and `root_list` being passed in the right order to `verify`?

gilanghamidy: No it does not have to. I checked on the `mbedtls` code, the verify function iterates the CRL list ignoring the order of the trusted certificates that is passed there. They are initially matching the same DN before continuing verifying the CRL integrity and authenticity. So it can be in any order.

intel-sgx/pcs/src/lib.rs
```diff
@@ -281,6 +284,64 @@ fn create_cert_chain(certs: &Vec<String>) -> Result<(Vec<MbedtlsBox<Certificate>
     }
 }
 
+/// Function to verify the PCS certificate chain that is attached in the response
+/// from the PCS server. Previously it was replicated in each `verify` function. But
+/// we now centralize this in here and shared among all PCS types. Also, we incorporate
+/// passing the root CA CRL in here if available
+#[cfg(feature = "verify")]
+fn build_and_verify_cert_chain<B: Deref<Target = [u8]>>(
+    ca_chain: &Vec<String>,
```
Taowyoo: nit
```suggestion
    ca_chain: &[String],
```

Taowyoo (dismissed):

Taowyoo (dismissed):

Taowyoo (approved):

PR #929, opened 2026-05-01
Patch bump `pcs` and `dcap-artifact-retrieval`

gilanghamidy:
We are bumping the `pcs` and `dcap-artifact-retrieval` for the implementation of the Root CA CRL.

### DIFF
--- a/intel-sgx/pcs/src/lib.rs
+++ b/intel-sgx/pcs/src/lib.rs
@@ -68,7 +68,7 @@ pub trait PlatformType : Clone + Default {
 
 ///Function to attempt deserialize [PlatformType] instance based on the [PlatformType::platform_id] value.
 pub fn deserialize_platform_id<'de, D: Deserializer<'de>, T: PlatformType>(deserializer: D) -> Result<T, D::Error> {
-    let platform_str = <&str>::deserialize(deserializer)?;
+    let platform_str = String::deserialize(deserializer)?;
     if platform_str == T::platform_id() {
         Ok(T::default())
     } else {
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1252,7 +1252,7 @@ dependencies = [
 
 [[package]]
 name = "dcap-artifact-retrieval"
-version = "0.5.1"
+version = "0.5.2"
 dependencies = [
  "assert_matches",
  "backoff",
@@ -3464,7 +3464,7 @@ dependencies = [
 
 [[package]]
 name = "pcs"
-version = "0.8.2"
+version = "0.8.3"
 dependencies = [
  "anyhow",
  "b64-ct",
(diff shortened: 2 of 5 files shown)


intel-sgx/pcs/src/lib.rs
```diff
@@ -68,7 +68,7 @@ pub trait PlatformType : Clone + Default {
 
 ///Function to attempt deserialize [PlatformType] instance based on the [PlatformType::platform_id] value.
 pub fn deserialize_platform_id<'de, D: Deserializer<'de>, T: PlatformType>(deserializer: D) -> Result<T, D::Error> {
-    let platform_str = <&str>::deserialize(deserializer)?;
+    let platform_str = String::deserialize(deserializer)?;
```
Taowyoo, line 71: why change back again?

Taowyoo (approved):

Taowyoo:
please do `cargo publish --dry-run` locally before merge this PR
@gilanghamidy

Taowyoo:
I guess this PR is related to internal PR 13597
Do we also need to upgrade `sgx-isa` ?

intel-sgx/pcs/src/lib.rs
gilanghamidy: @Pagten said the `&str` mah have stronger restriction and may cause deserialization issue. I am not entirely sure here, but I am following his advice.

gilanghamidy:
> I guess this PR is related to internal PR 13597 Do we also need to upgrade `sgx-isa` ?

No, we don't make any changes to `sgx-isa` here and it also does not depend to `pcs`. The current internal PR build fail causes by crate patch that causes dependency duplication, which should be gone when we update `pcs` upstream.

intel-sgx/pcs/src/lib.rs
Pagten: I talked to @gilanghamidy about this. Deserializing to `&str` causes issues when the input to be deserialized does not literally contain the part to be deserialized as a subslice. See for instance https://github.com/serde-rs/json/issues/742.

I did not have time to check whether that could be the case for where this specific method is used, so I advised to revert back to `String` to be safe.

gilanghamidy:
Dry-run check is working. Crate must be pushed in the order of `pcs` then `dcap-artifact-retrieval`.

intel-sgx/pcs/src/lib.rs
gilanghamidy: Ah okay now I understand from this https://github.com/serde-rs/json/issues/742 example. In our case actually, if the input JSON has an escape sequence, it must not be accepted altogether since the expected valid input is only `SGX` and `TDX`. That function is basically kind of like enum deserializer, and should reject other string values that appears in that field type.

So I believe this should actually be safe.

Taowyoo: good to learn these

PR #932, opened 2026-05-18
Fix issues with rustc-dep-of-std in `insecure-time`

s-arash:
Fixed issues:
- update to a version of `spin` that supports `rustc-dep-of-std` (see https://github.com/zesterer/spin-rs/pull/188)
- change the weight formula so that it compiles in non-std


Goirad (approved):
LGTM

PR #937, opened 2026-05-29
[RTE-941]: Relax initramfs interface types.

sardok:
Influnced by `std::fs` and after talking with @tvsfx i've decided to add support for Path like object to initramfs interface.

### DIFF
--- a/fortanix-vme/fortanix-vme-initramfs/Cargo.toml
+++ b/fortanix-vme/fortanix-vme-initramfs/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "fortanix-vme-initramfs"
-version = "0.1.0"
+version = "0.2.0"
 edition = "2021"
 authors = ["Fortanix, Inc."]
 license = "MPL-2.0"
--- a/fortanix-vme/fortanix-vme-initramfs/src/fs_tree.rs
+++ b/fortanix-vme/fortanix-vme-initramfs/src/fs_tree.rs
@@ -2,6 +2,7 @@ use crate::{Error, ReadSeek, DEFAULT_GID, DEFAULT_UID};
 use cpio::NewcBuilder;
 use derivative::Derivative;
 use normalize_path::NormalizePath;
+use std::ffi::OsString;
 use std::fmt;
 use std::io::Cursor;
 use std::path::{Path, PathBuf};
@@ -206,7 +207,7 @@ impl FsTreeEntry {
 
     pub fn symlink<P: AsRef<Path>, U: AsRef<Path>>(path: P, mode: u32, target: U) -> Self {
         let path = path.as_ref().to_path_buf();
-        let target = target.as_ref().to_path_buf();
+        let target = target.as_ref().into();
 
         Self {
             path,
@@ -225,7 +226,7 @@ pub enum FsTreeEntryInner {
         content: Box<dyn ReadSeek>,
     },
     Symlink {
-        target: PathBuf,
+        target: OsString,
     },
 }
 
@@ -240,8 +241,7 @@ impl FsTreeEntry {
             FsTreeEntryInner::Directory => Box::new(Cursor::new([] as [u8; 0])),
             FsTreeEntryInner::File { content } => content,
             FsTreeEntryInner::Symlink { target } => {
-                let os_string = target.into_os_string();
-                Box::new(Cursor::new(os_string.into_encoded_bytes()))
+                Box::new(Cursor::new(target.into_encoded_bytes()))
             }
         };
         Ok((builder, content))
--- a/fortanix-vme/fortanix-vme-initramfs/src/fs_tree.rs
+++ b/fortanix-vme/fortanix-vme-initramfs/src/fs_tree.rs
@@ -121,14 +121,14 @@ impl FsTree {
         self
     }
 
-    pub fn add_symlink<P: AsRef<Path>>(self, path: P, target: &str) -> FsTree {
+    pub fn add_symlink<P: AsRef<Path>, U: AsRef<Path>>(self, path: P, target: U) -> FsTree {
         self.add_symlink_with_permissions(path, target, DEFAULT_SYMLINK_PERMS)
     }
 
-    pub fn add_symlink_with_permissions<P: AsRef<Path>>(
+    pub fn add_symlink_with_permissions<P: AsRef<Path>, U: AsRef<Path>>(
         mut self,
         path: P,
-        target: &str,
+        target: U,
         mode: u32,
     ) -> FsTree {
         let path = Self::normalize_path(path);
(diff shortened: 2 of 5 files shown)


fortanix-vme/fortanix-vme-initramfs/src/fs_tree.rs
tvsfx: it's odd that this takes `&PathBuf` - that's a type that's almost never useful

fortanix-vme/fortanix-vme-initramfs/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "fortanix-vme-initramfs"
-version = "0.1.0"
+version = "0.2.0"
```
tvsfx, line 3: This seems to have caused problems in CI; probably cause you didnt update the lock file?

fortanix-vme/fortanix-vme-initramfs/src/fs_tree.rs
```diff
@@ -315,6 +329,34 @@ mod tests {
         assert!(files.eq_metadata(&expected));
     }
 
+    #[test]
+    fn test_structure_with_paths() {
```
tvsfx: you deliberately add `Path` and `PathBuf` here, right? Cause `str` and `String` implement `AsRef<Path>`, so this is technically not necessary

sardok: Yes, wanted to give it a try. Letme remove it as it seems unnecessary.

fortanix-vme/fortanix-vme-initramfs/Cargo.toml
sardok: I updated the lockfile now. I am not very confident about this change. What do you think?

fortanix-vme/fortanix-vme-initramfs/src/fs_tree.rs
tvsfx, line 1: should we also refactor `verify_entry` (note: https://doc.rust-lang.org/std/primitive.str.html#impl-PartialEq%3CPath%3E-for-str), `verify_entry_content` and `read_entry_by_path` in lib.rs?

tvsfx: Mostly the latter since that is public, but then it makes sense to refactor the first 2 I think

fortanix-vme/fortanix-vme-initramfs/Cargo.toml
tvsfx: We're changing the public API, so this is the correct version bump. Or is that not the question?

fortanix-vme/fortanix-vme-initramfs/src/fs_tree.rs
```diff
@@ -112,11 +121,16 @@ impl FsTree {
...
-    pub fn add_symlink(self, path: &str, target: &str) -> FsTree {
+    pub fn add_symlink<P: AsRef<Path>>(self, path: P, target: &str) -> FsTree {
         self.add_symlink_with_permissions(path, target, DEFAULT_SYMLINK_PERMS)
     }
 
-    pub fn add_symlink_with_permissions(mut self, path: &str, target: &str, mode: u32) -> FsTree {
+    pub fn add_symlink_with_permissions<P: AsRef<Path>>(
+        mut self,
+        path: P,
+        target: &str,
```
tvsfx: any reason to not make `target` a path too?

sardok: I tried but i didn't find it useful. On filesystem its actually a string whether exists or not and we're not doing any kind of processing for target that would benefit of being Path.

fortanix-vme/fortanix-vme-initramfs/Cargo.toml
sardok: Yes, i think we should bump too. I was thinking the way we consume those packages, we're sourcing from git but that should be fine as long as we don't update consumer packages lock files.

fortanix-vme/fortanix-vme-initramfs/src/fs_tree.rs
tvsfx: It's just that we might represent targets using `Path`s on the client side at some point, and this way we dont have to do any further manipulation. 

One thing to keep in mind is that this might cause a stark increase of the # of versions of this function being compiled during monomorphisation, but I dont think thats an issue here. (This is sometimes a reason to call `as_ref` as soon as possible)

fortanix-vme/fortanix-vme-initramfs/src/fs_tree.rs
sardok: What do you suggest, should it take `Path`? `&PathBuf` is passed because `FsTreeEntry::path` is `PathBuf` too and we're using this in comparison with existing entries.

fortanix-vme/fortanix-vme-initramfs/Cargo.toml
tvsfx: Yes, that will be fine. The only thing might be that you cannot `cargo update` to a newer commit if your git dep also specifies a `version` key, but I dont think we do that.

fortanix-vme/fortanix-vme-initramfs/src/fs_tree.rs
tvsfx: it can take `&Path`; `&PathBuf` deref-coerces to `&Path`

tvsfx: https://doc.rust-lang.org/beta/std/path/struct.PathBuf.html#impl-PartialEq%3C%26Path%3E-for-PathBuf
equality will still work

sardok: Updated. Due to errors types (invariants expect `String`), i had to add `format!("{}", p.as_ref().display())` but thats fine.

fortanix-vme/fortanix-vme-initramfs/src/fs_tree.rs
sardok: Removed.

fortanix-vme/fortanix-vme-initramfs/src/fs_tree.rs
sardok: Good. Updated.

fortanix-vme/fortanix-vme-initramfs/src/fs_tree.rs
sardok: Updated. So now every symlink related functions take two types. `P: AsRef<Path>` and `U: AsRef<Path>`.

aditijannu (approved):

PR #845, opened 2025-11-23
Add 21 missing stabilized io::ErrorKind variants to sgx_abi::Error (fixes #693)

DejoeNithin:
Fixes #693
* Added only stabilized Rust io::ErrorKind

<table>
<tr><th>Rust ErrorKind</th><th>libc constant</th><th>Decimal</th><th>Hex</th></tr>
<tr><td>ArgumentListTooLong</td><td>E2BIG</td><td>7</td><td>0x07</td></tr>
<tr><td>OutOfMemory</td><td>ENOMEM</td><td>12</td><td>0x0C</td></tr>
<tr><td>ResourceBusy</td><td>EBUSY</td><td>16</td><td>0x10</td></tr>
<tr><td>CrossesDevices</td><td>EXDEV</td><td>18</td><td>0x12</td></tr>
<tr><td>NotADirectory</td><td>ENOTDIR</td><td>20</td><td>0x14</td></tr>
<tr><td>IsADirectory</td><td>EISDIR</td><td>21</td><td>0x15</td></tr>
<tr><td>ExecutableFileBusy</td><td>ETXTBSY</td><td>26</td><td>0x1A</td></tr>
<tr><td>FileTooLarge</td><td>EFBIG</td><td>27</td><td>0x1B</td></tr>
<tr><td>StorageFull</td><td>ENOSPC</td><td>28</td><td>0x1C</td></tr>
<tr><td>NotSeekable</td><td>ESPIPE</td><td>29</td><td>0x1D</td></tr>
<tr><td>ReadOnlyFilesystem</td><td>EROFS</td><td>30</td><td>0x1E</td></tr>
<tr><td>TooManyLinks</td><td>EMLINK</td><td>31</td><td>0x1F</td></tr>
<tr><td>Deadlock</td><td>EDEADLK</td><td>35</td><td>0x23</td></tr>
<tr><td>InvalidFilename</td><td>ENAMETOOLONG</td><td>36</td><td>0x24</td></tr>
<tr><td>Unsupported</td><td>ENOSYS</td><td>38</td><td>0x26</td></tr>
<tr><td>DirectoryNotEmpty</td><td>ENOTEMPTY</td><td>39</td><td>0x27</td></tr>
<tr><td>NetworkDown</td><td>ENETDOWN</td><td>100</td><td>0x64</td></tr>
<tr><td>NetworkUnreachable</td><td>ENETUNREACH</td><td>101</td><td>0x65</td></tr>
[...]

PR #939, opened 2026-06-10
[PROD-11482] Revert adding errorKind to ABI and relax version dependency 

Taowyoo:
In https://github.com/fortanix/rust-sgx/pull/845 we added new ErrorKind to ABI without upstream it.
It causes enclave using this new code to panic while std library cannot recognize new error variant. See

https://github.com/rust-lang/rust/blob/485ec3fbcc12fa14ef6596dabb125ad710499c9e/library/std/src/sys/pal/sgx/abi/usercalls/mod.rs#L284-L310

Before we find out a mature plan about updating both ABI and std lib code, we need to revert related changes.

This PR reverts related changes and relax other crates' dependency on `fortanix-sgx-abi`.


aditijannu (dismissed):

s-arash:
Are we planning on yanking the 0.7.0 version of `fortanix-sgx-abi`?

gilanghamidy:
@Taowyoo Maybe you need to revert this one too for now: https://github.com/fortanix/rust-sgx/pull/873. This workaround was introduced due to this ABI breaking PR to make us can still compile with our own compiler.

Taowyoo:
> Are we planning on yanking the 0.7.0 version of `fortanix-sgx-abi`?

Yes

And also `enclave-runner-sgx` `0.1.1`

intel-sgx/enclave-runner-sgx/Cargo.toml
```diff
@@ -21,7 +21,7 @@ exclude = ["fake-vdso/.gitignore", "fake-vdso/Makefile", "fake-vdso/main.S"]
 [dependencies]
 # Project dependencies
 sgxs = { version = "0.8.1", path = "../sgxs" }
-fortanix-sgx-abi = { version = "0.7.0", path = "../fortanix-sgx-abi" }
+fortanix-sgx-abi = { version = ">=0.6.0, <0.8.0", path = "../fortanix-sgx-abi" }
```
s-arash: If we are yanking fortanix-sgx-abi 0.7, then this dependency can just be `0.6`, right?

s-arash (dismissed):

intel-sgx/fortanix-sgx-abi/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "fortanix-sgx-abi"
-version = "0.7.0"
+version = "0.7.1"
```
s-arash: This is semver incompatible. We cannot just remove variants from an enum and announce semver compatibility.

Taowyoo: since we are going to yank 0.7.0, i think this should be fine

s-arash:
I'm not clear on what the plan is here. Are we removing variants from `enum Error` in `fortanix-sgx-abi` and releasing a semver compatible version? That does not work.
We can either leave `fortanix-sgx-abi` alone, OR remove the enum variants from it and yank the existing version, and revert the version to 0.6.x.

intel-sgx/fortanix-sgx-abi/Cargo.toml
s-arash: No it's not. It tells everyone using (the to be yanked) 0.7 that they should upgrade to 0.7.1, and everything will continue to work. But that's clearly not true.

intel-sgx/enclave-runner-sgx/Cargo.toml
Taowyoo: but we cannot reduce required version here. The published 0.1.1 and 0.1.0 enclave-runner-sgx still depends on 0.7.0 fortanix-sgx-abi

s-arash: I don't think changing dependency requirements (for a private dependency) is sem-ver breaking. It's certainly much better than removing enum variants. If you really don't like it, we can also leave `fortanix-sgx-abi` alone as I said.

intel-sgx/enclave-runner-sgx/Cargo.toml
s-arash: I think we need to bump this version too.

intel-sgx/enclave-runner-sgx/Cargo.toml
```diff
@@ -42,7 +42,3 @@ futures = { version = "0.3", features = ["compat", "io-compat"] } # MIT/Apache-2
 [features]
 default = ["crypto-openssl"]
 crypto-openssl = ["openssl", "sgxs/crypto-openssl"]
-err-compat = ["default"]
```
s-arash, line 45: FYI: removing features also breaks semver compatibility.

Taowyoo:
After discussion, the final approach is:

1. Yank `fortanix-sgx-abi 0.7.0`
2. Publish new `ipc-queue 0.5.2` which will be compatible with fortanix-sgx-abi 0.6.0 
3. Publish new `async-usercalls 0.7.2` which will be compatible with fortanix-sgx-abi 0.6.0
4. Publish new  `enclave-runner-sgx 0.1.3` which will be compatible with fortanix-sgx-abi 0.6.0

s-arash (approved):

gilanghamidy (approved):
I assume no related changes that I made other than #873 and that TDX changes are specifically on the `pcs` and `sgx-isa`, which does not depend on this.

PR #946, opened 2026-08-07
[RTE-1139]: Fix enclave-runner-sgx build on nightly toolchain.

sardok:
Fixes enclave-runner-sgx build error on the latest nightly toolchain.

PR #947, opened 2026-08-07
[RTE-1093][RTE-960] Expose issue dates from `pcs` types and fix TDX module string parsing to use Base 16

gilanghamidy:
We need to expose the dates in the `pcs` data types so the consumer can also use that date. Currently, the date is only used internally for TCB verification.

Also this PR addresses the recent confirmation from Intel that the TDX module numbering format that appears in the TCB info is using Base 16/hexadecimal format instead of Base 10/decimal.

### DIFF
--- a/intel-sgx/pcs/src/tcb_evaluation_data_numbers.rs
+++ b/intel-sgx/pcs/src/tcb_evaluation_data_numbers.rs
@@ -41,6 +41,16 @@ impl<T: PlatformTypeForTcbInfo, V: VerificationType> TcbEvaluationDataNumbers<T,
     }
 }
 
+impl<T: PlatformTypeForTcbInfo> TcbEvaluationDataNumbers<T> {
+    pub fn issue_date(&self) -> &DateTime<Utc> {
+        &self.issue_date
+    }
+
+    pub fn next_update(&self) -> &DateTime<Utc> {
+        &self.next_update
+    }
+}
+
 impl<T: PlatformTypeForTcbInfo> TcbEvaluationDataNumbers<T, Unverified> {
     /// Given a particular TCB level, select the best available TCB eval number.
     /// That is the one that gives the most favorable TCB status, and the higher
--- a/intel-sgx/pcs/src/tcb_info.rs
+++ b/intel-sgx/pcs/src/tcb_info.rs
@@ -324,8 +324,7 @@ impl TdxModuleTcbLevel {
 
 /// Deserializer function for TDX ID as it appears in the TDX Module entry of the TDX TCB Info.
 /// TDX ID is always in the format of "TDX_xx" in which xx is a number of the TDX Module ID.
-/// It is currently unspecified if the number is in base 10 or base 16. Currently it is assumed
-/// it is base 10 with leading zero.
+/// It is confirmed from Intel that xx is in base 16 radix.
 ///
 /// Reference: https://api.portal.trustedservices.intel.com/content/documentation.html#pcs-tcb-info-tdx-v4
 /// step no. 5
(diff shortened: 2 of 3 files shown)


intel-sgx/pcs/src/tcb_info.rs
```diff
@@ -951,4 +959,31 @@ mod tests {
         }
     }
 
+    #[test]
+    fn test_tdx_id_parser() {
+        #[derive(Clone, Debug, Serialize, Deserialize, Eq, PartialEq)]
+        #[serde(rename_all = "camelCase")]
+        pub struct TdxIdTest {
+            #[serde(
+                deserialize_with = "tdx_id_deserializer",
```
Taowyoo, line 968: nit suggestion:

This can be simplified to `#[serde(with = "tdx_id")]`.
For example:
```
#[derive(Clone, Debug, Deserialize, Serialize, Eq, PartialEq)]
#[serde(rename_all = "camelCase")]
pub struct TdxIdTest  {
    #[serde(with = "tdx_id")]
    id: u8,
}

mod tdx_id {
    use serde::{Deserialize, Deserializer, Serializer};

    pub fn serialize<S>(id: &u8, serializer: S) -> Result<S::Ok, S::Error>
    where
        S: Serializer,
    {
        // ...
    }

    pub fn deserialize<'de, D>(deserializer: D) -> Result<u8, D::Error>
    where
        D: Deserializer<'de>,
    {
        // ...
    }
}
```

Taowyoo (dismissed):

Taowyoo:
In latest nightly, OUR_DIR layout changed. For example, stable vs nightly on my laptop:
```
rust-mbedtls/target/x86_64-fortanix-unknown-sgx/debug/build/mbedtls-3202cb041a903437/out
``` 
vs
```
rust-mbedtls/target/x86_64-fortanix-unknown-sgx/debug/build/mbedtls/f8d961f56b5f3f44/out
```
We can use `export __CARGO_TEMPORARY_BUILD_DIR_NEW_LAYOUT_OPT_OUT=1` to opt-out the new layout, but it will be removed soon: https://github.com/rust-lang/cargo/issues/17182

Taowyoo:
Created https://github.com/fortanix/rust-mbedtls/pull/411 to fix CI issue, please review

intel-sgx/pcs/src/tcb_evaluation_data_numbers.rs
```diff
@@ -39,6 +39,14 @@ impl<T: PlatformTypeForTcbInfo, V: VerificationType> TcbEvaluationDataNumbers<T,
     pub fn numbers(&self) -> Iter<'_, TcbEvalNumber> {
         self.tcb_eval_numbers.iter()
     }
+
+    pub fn issue_date(&self) -> &DateTime<Utc> {
```
raoulstrackx, line 45: Do we need to expose this also to unverified TCB Evaluation Data Numbers? I like the idea that you first need to verify that the data comes from Intel before you're able to do anything meaningful. The `issue_data` and `next_update` functions you added to other structs, can only be called after they have been verified.

gilanghamidy: Ah yeah I think I put it in the wrong `impl`. In other struct, the `impl` is defaulted to `Verified`, but in this part, it is explicitly defined to any `V` type.

gilanghamidy: Wait, if I look at it again, the `TcbEvaluationDataNumber` type has many operational functions in `Unverified` type. I think it was like that since beginning. Maybe it was because the use case is different compared to `TcbInfo` type. But I am not entirely sure.

Taowyoo:

intel-sgx/pcs/src/tcb_evaluation_data_numbers.rs
Taowyoo: Is there only `select_best`  below ?

gilanghamidy: Apparently when looking at the consumer codes and the code history, we never really use `TcbEvaluationDataNumber` in the `Verified` form. We are also not enforcing it in the code by putting `numbers` function behind the `Verified` type constraint, unlike other types like `TcbInfo`.

Question to @raoulstrackx, do you want to migrate it to start enforcing this constraint?

intel-sgx/pcs/src/tcb_info.rs
gilanghamidy: Yeah this pattern was the remnants of the decision I made couple months ago when introducing the custom (de)serializer for new PCS fields. I was a bit hesitant with introducing new module for every custom (de)serializer, and some of them was also only one-way (no plan to implement the serializer. I believe `serde` complaints if only one function is implemented). I'll keep it for now and revisit later when we do another refactoring round.

intel-sgx/pcs/src/tcb_evaluation_data_numbers.rs
raoulstrackx: I think that's also used to simply download the list of artifacts based on TCB evaluation numbers. Enforcing it now would be a breaking change.

raoulstrackx (approved):

PR #948, opened 2026-08-10
[RTE-1113]: Add extended prod id to Sigstruct & Report structures - Part 1.

sardok:
This PR extended prod id to `Report` & `Sigstruct` types based on the [Intel® 64 and IA-32 Architectures Software Developer's Manual Volume 3D: System Programming Guide, Part 4](https://cdrdv2.intel.com/v1/dl/getContent/671269).

A new PR on top of this branch is upcoming, which adds Signer support to use the new field.


intel-sgx/sgx-isa/src/lib.rs
gilanghamidy: Should we also include`CONFIGSVN` and `ISVFAMILYID` field that is also missing?

intel-sgx/sgx-isa/src/lib.rs
```diff
@@ -649,11 +640,12 @@ struct_def! {
 pub struct Report {
     pub cpusvn: [u8; 16],
     pub miscselect: Miscselect,
-    #[cfg_attr(feature = "serde", serde(default = "report_reserved1"), serde(skip))]
-    pub _reserved1: [u8; 28],
+    #[cfg_attr(feature = "serde", serde(skip))]
+    pub _reserved1: [u8; 12],
```
gilanghamidy: Also `CET_ATTRIBUTES` here

gilanghamidy (commented):
I would prefer to update the Report structure to reflect the recent specs. So that we have it consistent instead of only implement partially of what we currently need. 

At some point we may want to also update other structures as well.

Taowyoo:

intel-sgx/sgxs/Cargo.toml
```diff
@@ -14,7 +14,7 @@ categories = ["parsing", "encoding"]
 
 [dependencies]
 # Project dependencies
-sgx-isa = { version = ">=0.4.0, <0.6.0", path = "../sgx-isa" }
+sgx-isa = { version = ">=0.4.0, <0.7.0", path = "../sgx-isa" }
```
Taowyoo: how could this crate be compatible with sgx-isa version 0.5 and 0.6?
The `Sigstruct` definitions in two version are different.

intel-sgx/ias/src/api.rs
```diff
@@ -143,7 +143,8 @@ impl EnclaveQuoteBody {
         Report {
             cpusvn,
             miscselect,
-            _reserved1,
+            _reserved1: [0; 12],
```
Taowyoo: should we preserve the original data?
e.g.
```
_reserved1: _reserved1[0..12],
isvextnprodid: _reserved1[12..],
```

intel-sgx/ias/Cargo.toml
```diff
@@ -23,7 +23,7 @@ url = "2.2"
 mbedtls = { version = ">=0.12.0, <0.14.0", features = ["std"], default-features = false, optional = true }
 pkix = ">=0.1.0, <0.3.0"
 
-sgx-isa = { version = ">=0.4.0, <0.6.0", path = "../sgx-isa" }
+sgx-isa = { version = ">=0.4.0, <0.7.0", path = "../sgx-isa" }
```
Taowyoo: Same question here, since related struct's field changed . This crate obviously cannot compile with two versions at same time.

intel-sgx/ias/src/api.rs
sardok: I thought about that, but i don't think there is meaningful data `_reserved1` field of quote body, please let me know if you think there might be. The info i can find about this structure is at `Table 4: ECDSA 256-bit Quote Signature Data Structure` in [Intel_SGX_ECDSA_QuoteLibReference_DCAP_API](https://download.01.org/intel-sgx/latest/dcap-latest/linux/docs/Intel_SGX_ECDSA_QuoteLibReference_DCAP_API.pdf). Letme know if this is incorrect one.

sardok: I talked with @raoulstrackx about this. We will be removing this crate as it is deprecated. Letme push a PR for removal of it.

intel-sgx/sgx-isa/src/lib.rs
sardok: Ok, letme add remaining fields as well.

intel-sgx/sgxs/Cargo.toml
sardok: Yes but i didn't do any change that uses the latest additions made to `Sigstruct` in `sgxs` crate in this PR. So `sgxs` compiles fine in this PR. I planned that work for the Part 2: https://github.com/fortanix/rust-sgx/pull/949 (but apparently i forgot to update in there too, letme fix that quickly).

intel-sgx/ias/Cargo.toml
sardok: Saved it for the Part 2, but i will create soon a PR to remove this crate completely.

intel-sgx/sgx-isa/src/lib.rs
sardok: Added thanks.

intel-sgx/sgx-isa/src/lib.rs
sardok: Added thanks.

intel-sgx/ias/src/api.rs
sardok: `ias` is gone with this PR now: https://github.com/fortanix/rust-sgx/pull/950.

intel-sgx/ias/Cargo.toml
sardok: Removed ias crate.

intel-sgx/sgx-isa/src/lib.rs
```diff
@@ -581,7 +571,8 @@ pub struct Sigstruct {
     pub attributes: Attributes,
     pub attributemask: [u64; 2],
     pub enclavehash: [u8; 32],
-    pub _reserved3: [u8; 32],
+    pub _reserved3: [u8; 16],
+    pub isvextprodid: [u8; 16],
```
jethrogb, line 578: Let's just add all the new fields?

intel-sgx/sgxs/Cargo.toml
Taowyoo: I am referring to code here:

https://github.com/fortanix/rust-sgx/blob/43500b4621db6a4916702fb7fe7c1c924c4275e0/intel-sgx/sgxs/src/sigstruct.rs#L124-L151

It does not compile when it's using 0.5.0 sgx-isa.

I think we need to strict dep requirement in `sgxs`

intel-sgx/sgx-isa/src/lib.rs
sardok: Added all remaining Sigstruct fields. Thanks.

intel-sgx/sgxs/Cargo.toml
sardok: Yes, you're right, i missed that part. Should be fixed now. Thanks.

sardok:
All missing fields in Report & Sigstruct should be added now.

intel-sgx/sgx-isa/src/lib.rs
```diff
@@ -649,19 +643,25 @@ struct_def! {
 pub struct Report {
     pub cpusvn: [u8; 16],
     pub miscselect: Miscselect,
-    #[cfg_attr(feature = "serde", serde(default = "report_reserved1"), serde(skip))]
-    pub _reserved1: [u8; 28],
+    pub cet_attributes: u8,
```
Taowyoo, line 646: Are these `serde(default` attributes removed intentionally?

Taowyoo (commented):
left one comment about removing `serde(default = "report_reservedX")`, other parts LGTM

intel-sgx/sgx-isa/src/lib.rs
sardok: Yes serde can't provide default values for arrays larger than `32` elements. So i removed `default = *_reserved(..)`  for shrinked arrays.

Taowyoo (approved):

PR #949, opened 2026-08-10
[RTE-1113] Add extended prod id support to signer - Part 2.

sardok:
This PR add signer support for the recently added `ISVEXTPRODID` field (see the base branch) of `SIGSTRUCT`. Argument parsed to `Signer` is parsed as `u128` which aligns with the `ISVEXTPRODID` field size (i.e `[u8; 16]`).

### DIFF
--- a/intel-sgx/sgxs-tools/src/bin/sgxs-sign.rs
+++ b/intel-sgx/sgxs-tools/src/bin/sgxs-sign.rs
@@ -159,10 +159,6 @@ fn do_sign<'a>(matches: &clap::ArgMatches<'a>, key: &PKey<pkey::Private>) -> Sig
         signer.miscselect(sel, !mask);
     }
 
-    let isvextprodid = matches
-        .value_of("isvextprodid")
-        .map(parse_num::<u128>)
-        .unwrap_or_default();
 
     let (mut attributes, attributemask) = matches
         .value_of("attributes/attributemask")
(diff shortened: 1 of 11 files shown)


Taowyoo:

Taowyoo (approved):

Taowyoo (dismissed):

intel-sgx/sgxs-tools/src/bin/sgxs-sign.rs
```diff
@@ -174,6 +180,13 @@ fn do_sign<'a>(matches: &clap::ArgMatches<'a>, key: &PKey<pkey::Private>) -> Sig
         attributes |= sgx_isa::AttributesFlags::DEBUG.bits();
         attributemask &= !(sgx_isa::AttributesFlags::DEBUG.bits());
     }
+    // If extended product id is has a value
+    // make KSS essential. Hosts without KSS support
+    // would fail on enclave launch.
+    if isvextprodid > 0 {
```
gilanghamidy: In this case, we cannot use zero value in `isvextprodid`. Is it actually specified in the SGX specification that `isvextprodid` must always be nonzero if it is used?

sardok: We're reserving `0` for SGX that doesn't support KSS also for the default value (where all mitigations are expected case), if i'm not wrong.

raoulstrackx: This may result in a different attributes/attributemask then requested. When specifying an `isvextprodid` without enabling KSS, shouldn't that be an error?

raoulstrackx: In principle this should also pass API review as this is customer facing.

Taowyoo: we can use `matches.is_present("isvextprodid")` to check if cli is given with `--isvextprodid` plus `0` value, or change the unwrap_or_default above. 

If we go with Gilang's proposal of using first bit of `isvextprodid` to indicate if this feature is enabled, then case that `--isvextprodid` plus `0` value should be rejected or be treated as not turning on KSS.

sardok: Fixed the code. Now, if this value is passed, `KSS` is set (regardless of its value). The caller/consumer will be responsible for if `0` makes sense or not.

gilanghamidy (approved):

Taowyoo (approved):

PR #950, opened 2026-08-11
[RTE-1113] Remove ias crate.

sardok:
This PR remove deprecated `ias` crate.


gilanghamidy:
Also remove the release job in `release.yaml` https://github.com/fortanix/rust-sgx/blob/916b43ef8a95062b82334ecc3e1d82e9985a0dca/.github/workflows/release.yml

raoulstrackx (commented):
In general looks good but CI fails...

gilanghamidy:
> In general looks good but CI fails...

YEah the CI for `ias` is also not removed

sardok:
Thanks @gilanghamidy @raoulstrackx , i've removed ias from CI and releases.

gilanghamidy (approved):
Approving this, but we may also want to mark the crates.io entry https://crates.io/crates/ias/0.2.3 to be obsolete or deprecated. I suppose it is @Taowyoo or @Pagten that has access to that

PR #951, opened 2026-08-11
[RTE-1140] Fix AESM v2.30 Compatibility Issue

gilanghamidy:
PSW v2.30 release includes a breaking changes in compatibility to its protobuf definition from `optional` to `required`. The workaround is to always supply the argument with zero value.

Related PR in Intel:
https://github.com/intel/confidential-computing.sgx/pull/1107

Although it is actually against the norm to change protobuf definition after it has been published (https://protobuf.dev/programming-guides/proto3/), we can expect that this behavior may persist moving forward as it has been already published. Specifying zero value should be backward compatible with previous AESM version, or should Intel decides to revert the changes in future version.

### DIFF
--- a/intel-sgx/aesm-client/src/imp/aesm_protobuf/mod.rs
+++ b/intel-sgx/aesm-client/src/imp/aesm_protobuf/mod.rs
@@ -157,12 +157,13 @@ impl AesmClient {
 
         req.set_att_key_id(att_key_id.clone());
         req.set_b_pub_key_id(false);
-        
+        req.set_buf_size(0);
+
         let res = self.transact(req)?;
         let buf_size = res.pub_key_id_size();
 
         let mut req = InitQuoteExRequest::new();
-        
+
         req.set_att_key_id(att_key_id);
         req.set_b_pub_key_id(true);
         req.set_buf_size(buf_size);
--- a/intel-sgx/aesm-client/src/lib.rs
+++ b/intel-sgx/aesm-client/src/lib.rs
@@ -396,4 +396,41 @@ mod tests {
             false
         });
     }
+
+    #[test]
+    fn test_get_quote_full() {
+        use sgxs_loaders::isgx::Device as IsgxDevice;
+
+        const SGX_QL_ALG_ECDSA_P256: u32 = 2;
+
+        let mut device = IsgxDevice::new()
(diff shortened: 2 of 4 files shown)


intel-sgx/aesm-client/src/imp/aesm_protobuf/mod.rs
```diff
@@ -157,12 +157,13 @@ impl AesmClient {
 
         req.set_att_key_id(att_key_id.clone());
         req.set_b_pub_key_id(false);
-        
+        req.set_buf_size(0);
```
raoulstrackx, line 160: Did you test this with a pre 2.30 aesm service?

gilanghamidy: Yes, also to AESM 2.27 in other machine as well.

gilanghamidy: Wait a sec, I think I need to verify some stuff as well

raoulstrackx: Can you also update `aesm_proto.proto` and leave a comment there?

intel-sgx/aesm-client/src/lib.rs
```diff
@@ -396,4 +396,41 @@ mod tests {
             false
         });
     }
+
+    #[test]
+    fn test_get_quote_full() {
```
Taowyoo, line 401: Just curious how you run sgx test for this crate.
I meet issue with `openssl` with
```
cargo +nightly test -p aesm-client --features test-sgx,sgxs --target=x86_64-fortanix-unknown-sgx --no-run
```

gilanghamidy: Not yet working. I am currently investigating it. Because the original test is also not properly working.

raoulstrackx:
The [upstream fix](https://github.com/intel/confidential-computing.sgx/pull/1107) in the aesm service landed.

gilanghamidy:
> The [upstream fix](https://github.com/intel/confidential-computing.sgx/pull/1107) in the aesm service landed.

Should we keep this one, or abandon it? Do you hear anything about their release window and whether they are going to yank v2.30 from their repo?

PR #954, opened 2026-08-20
sgx-isa: fix UB in Sigstruct::signature_data. test with miri in CI.

phlip9:
miri surfaced some UB in `Sigstruct::signature_data` under the default stacked borrows model:

```
$ cargo +nightly miri test -p sgx-isa --lib test_signature_data
test test_signature_data ... error: Undefined Behavior: trying to retag from <458630> for SharedReadOnly permission at alloc138652[0x10], but that tag does not exist in the borrow stack for this location
   --> intel-sgx/sgx-isa/src/lib.rs:600:17
    |
600 |                 slice::from_raw_parts(part1_start, part1_end - (part1_start as usize)),
    |                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ this error occurs as part of retag at alloc138652[0x0..0x80]
    |
```

The issue is the `part1_start` pointer extending past the scope of its original "allocation" (the `self.header` field), which is UB under stacked borrows.

Instead, we can use [`std::mem::offset_of!`](https://doc.rust-lang.org/stable/std/mem/macro.offset_of.html) to just get the actual field offsets instead of deriving them on-the-fly. We can then offset directly from the `self` pointer to get our slices, which stays within the pointer's "allocation" (i.e., all of `Sigstruct`).

---

I've also added a few small unit tests to exercise the `unsafe` blocks in `sgx-isa` (at least, the ones that aren't cfg'd for SGX).

It would be nice if we could get `miri` checking things more systematically, so I've added a `cargo miri test -p sgx-isa` step in CI. This is definitely optional, but I'd like to slowly expand our coverage to more rust-sgx crates if possible.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -51,7 +51,7 @@ jobs:
       - name: Setup Rust toolchain
         run: |
           rustup target add x86_64-fortanix-unknown-sgx x86_64-unknown-linux-musl
-          rustup toolchain add nightly
+          rustup toolchain add nightly --component miri
           rustup target add x86_64-fortanix-unknown-sgx --toolchain nightly
           rustup update
 
@@ -124,3 +124,6 @@ jobs:
 
       - name: snmalloc correctness test
         run: cd ./examples/mem-correctness-test && cargo run
+
+      - name: Nightly Miri test -p sgx-isa
+        run: cargo +nightly miri test --verbose --locked -p sgx-isa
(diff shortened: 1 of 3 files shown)


jethrogb:
Thanks for this and other PRs! The modern approach to implementing a crate like this would probably be with `zerocopy` ~, although I don't know that we can use that as a dependency of `std`.~ (this isn't a dependency of std)

phlip9:
Yeah no problem! I figured we'd start small and fix the immediate issues without more sweeping refactors. Though something like `zerocopy` would be quite nice to express our intentions and invariants more clearly.

PR #965, opened 2026-09-01
[RTE-1179]: Add build script to check stabilized features.

sardok:
Latest nightly compiler (`1.100`) fails to build `async-usercalls` due to the stabilized `never_type` feature. See the failure here: https://github.com/fortanix/rust-sgx/actions/runs/33515166527/job/99880253018. This PR is influced by the serde's approach to the similar problem.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -140,6 +140,7 @@ dependencies = [
  "ipc-queue",
  "lazy_static",
  "rand 0.8.5",
+ "rustc_version",
 ]
 
 [[package]]
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4293,9 +4293,9 @@ dependencies = [
 
 [[package]]
 name = "sdkms"
-version = "0.3.0"
+version = "0.3.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "5b12e3cb05862db268118482cbad26eee384b479de1924fe5404028e3444481a"
+checksum = "6d2e70f9b18786894511f918b4d95cc7b0db138cce7b7ee05a7be6fb4a25e3ee"
 dependencies = [
  "bitflags 1.2.1",
  "chrono",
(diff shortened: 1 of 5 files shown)


Taowyoo:
I will suggest to use rustc_version crate instead of calling rustc.
But I am fine to continue

Taowyoo (dismissed):

sardok:
Added rustc_version usage. Thank you @Taowyoo

Taowyoo (dismissed):

Taowyoo:
Fixed the issue in dependency `sdkms` with new version `0.5.1` https://github.com/fortanix/sdkms-client-rust/pull/14

Taowyoo:
Created this PR to update sdkms version https://github.com/fortanix/rust-sgx/pull/966
Or do we just create a patch version `0.3.1` instead?

Taowyoo:
To avoid too much change here, I release 0.3.1 of `sdkms` with similar patch in `0.5.1`, and create https://github.com/fortanix/rust-sgx/pull/968 to update this

sardok:
Thank you @Taowyoo merged `sdkms 0.3.1` PR.

Taowyoo (approved):

tvsfx:
> I will suggest to use rustc_version crate instead of calling rustc. But I am fine to continue

I thought it might be too much of a left-pad style dependency to make it worth adding, but since the vendored serde function that we were using seemed to have some cargo-culted logic and since it's only a build dep, I guess it's OK

sardok:
>  ... but since the vendored serde function that ...

I thought one reason why serde has that particular function could be about `no_std` feature. Not sure if `rustc_version` is `no_std` compatible but they may be wanting to keep deps as less as possible because of this reason.

tvsfx:
Yes, I'm quite sure that serde just didn't want to add an extra dependency for everyone and hence vendored. They used to use the `rustc_version` crate but reverted that at some point.

